PR #1, opened 2019-10-25
Users

GuillaumeCisco:
This PR implements three ways of authentication :
- BasicAuthentication # for node to node authentication
- SecureJWTAuthentication # for front/sdk/cli authentication
- ExpiryTokenAthentication # for sdk/cli authentication
- CustomSessionAuthentication # for web browsable api

It allows to remove the Basic Authentication in production environment for websites (not internally).  
It also remove the SettingsBackend becoming obsolete.

- SecureJWTAuthentication exposes a `/user/login` view for getting a secure JWT through secure cookies, split in two (based on the work of https://medium.com/@jcbaey/authentication-in-spa-reactjs-and-vuejs-the-right-way-e4a9ac5cd9a3 and https://medium.com/lightrail/getting-token-authentication-right-in-a-stateless-single-page-application-57d0c6474e3).
It uses the code from https://github.com/davesque/django-rest-framework-simplejwt
- ExpiryTokenAthentication exposes a classic `/api-token-auth` view for login. It is basically used for sdk and cli tools.
- CustomSessionAuthentication allows classic authentication thanks to a web form protected with a csrf token.
It exposes a `api-auth/login` view.

For being able to test this new paradigm, please be sure to carefully read the new updated README.

Known issues:
none

Linked with:
hlf-k8s: https://github.com/SubstraFoundation/hlf-k8s/pull/1
substra-backend: https://github.com/SubstraFoundation/substra-backend/pull/1
susbtra-frontend: https://github.com/SubstraFoundation/substra-frontend/pull/1
susbtra: https://github.com/SubstraFoundation/substra/pull/2
susbtra-ci: https://github.com/SubstraFoundation/substra-ci/pull/11
substra-tests: https://github.com/SubstraFoundation/substra-tests/pull/4
___

Tutorial:
# Users by org

## Need
We have the will to manage users by organization.

Node to node authentication i.e worker to backend instances communication is not impacted and use the current BasicAuth implementation.

We want to be able to register them from an administration point of view.

Each user can login himself.

## Authentications

There are three main authentication available running in parallel:
- Secure JWT Authentication
- Expiry Token Authentication
- Session Authentication

Each authentication process owns its login view which run against the same database and one and only user management.

#### Secure JWT Authentication
The first one `Secure JWT Authentication` is mainly used for web single one page application as our susbtrafront project.
We use a Secure JWT Authentication and not a simple JWT Authentication for being sure we are free of XSS and/or CORS attacks.
More reading here: 
- https://medium.com/@jcbaey/authentication-in-spa-reactjs-and-vuejs-the-right-way-e4a9ac5cd9a3
- https://medium.com/lightrail/getting-token-authentication-right-in-a-stateless-single-page-application-57d0c6474e3
It can also be used by SDK and CLI projects such as our substra-sdk and substra-cli.

#### Expiry Token Authentication
The second one `Expiry Token Authentication` is a classic for token authentication for tools like sdk and cli.

#### Session Authentication
The third one `Session Authentication` is a classic for server rendered pages, useful for our web browsable API available via django rest framework api.
It can also be used by SDK and CLI projects but is a little bit more difficult to implement, as two requests need to be done for logging in correctly.

## Testing

Four projects are impacted by this change:
- hlf-k8s #correct-domain
- substra-backend #users
- substra-frontend #users
- substra #users

Each has a branch entitled `users`. Except for hlf-k8s which is `correct-domain`

Launch a `hlf-k8s` network with two organizations (classic) on the `correct-domain` branch, make sure you have updated tour `/etc/hosts` file according to new domain name.

### substra

1. Go to the `users` branch.

That's all :)

#### Test

1. Install new version `pip install -e .`
2. Create configuration:
```bash
$> substra config --profile user -k -v 0.0 -u substra -p 'p@$swr0d44' http://substra-backend.owkin.xyz:8000
```
3. Login
```bash
$> substra login --profile user
```
4. Request
```bash
$> substra list objective --profile user
KEY                                                                 NAME                                    METRICS                 
1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3    Skin Lesion Classification Objective    macro-average recall    
3d70ab46d710dacb0f48cb42db4874fac14e048a0d415e266aad38c09591ee71    Skin Lesion Classification Objective    macro-average recall 
```

### substra-backend

1. Go to the `users` branch.
2. Install new substra with `pip install -e ../substra`
3. Generate nodes username/password:
```bash
$> python ./backend/node/generate_nodes.py
```
4. Build docker images:
```bash
$> sh build-docker-images.sh
```
5. Run images:
```bash
$> cd docker && python start.py -d --no-backup && cd -
```
6. Run populate.py

For each backend it will create by default an user with username `substra` and password `p@$swr0d44`.
The `populate.py` script uses these credentials for logging in.

#### Test:

1. Go to `http://substra-backend.owkin.xyz:8000`, be sure to have the `mod_header` running for not having the version error message.
2. Click on the login link on the top right of the page.
3. Login
4. Enjoy

### susbtra-frontend

1. Go to the `users` branch.
2. Rebuild the images and run them:
```bash
$> docker-compose up -d --force-recreate --build 
```

#### Test:

1. Go to `http://substra-frontend.owkin.xyz:3000`, be sure to have deactivated the `mod_header` version control.
2. Login
4. Enjoy

# Skaffold users

For testing with skaffolding, you need to setup your environment as described here:
https://github.com/owkin/infrastructure/tree/master/env/ci

If you're running on linux, you may encounter some issues..
Here is a little tutorial for making it work correctly.

First of all, make sure you are in folder where you have cloned all needed substra projects (substra-backend, susbtra-frontend, substra-tools, substra-ci, substra (non exhaustive list)).

Select these branches in the following projects
substra-backend #users
substra-frontend #users
substra # users

We will now configure minikube:
```bash
$> minikube delete # make sure to work from an empty environment
$> minikube start --cpus 6 --memory 8192 --disk-size 50g --kubernetes-version='v1.15.4' # make sure to have enough ressources and to use 1.15.4
$> kubectl create serviceaccount --namespace kube-system tiller
$> kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
$> eval $(minikube docker-env)
$> helm init --service-account tiller --upgrade
```
Open three terminals and go inside:
1. hlf-k8s
2. substra-backend
3. substra-frontend

In each terminal run:
```
$> eval $(minikube docker-env)
$> skaffold dev
```

wait...
keep waiting...
wait...
when you're sick of waiting, you can download and run k9s for knowing what is going on: https://github.com/derailed/k9s/releases
wait...
still waiting...
play with k9s...
When everything "look" blue with `running` and `complete` in the list, you are ready to test.

Troubleshooting:
Sometimes the secrets are not correctly removed. Reset them with:
```
$> kubectl delete secrets -n --all peer-1 && kubectl delete secrets -n --all peer-2 && kubectl delete secrets -n --all orderer
```

Do not forget to put in your `/etc/hosts`:
`XXX.XXX.XXX.XXX substra-backend.node-1.com substra-frontend.node-1.com substra-backend.node-2.com substra-frontend.node-2.com
` where `XXX.XXX.XXX.XXX` is the result of `minikube ip` command.

`minikube addons enable ingress`

If you plan to run a populate. make sure you have a substra-tools tagged to the right version:

```bash
$> docker pull eu.gcr.io/substra-208412/substra-tools:0.0.1
```
If you have login difficulties, follow documentation on https://techoverflow.net/2019/04/01/fixing-gcloud-warning-docker-credential-gcloud-not-in-system-path/
and  `$> sudo cp usr/bin/docker-credential-gcr usr/bin/docker-credential-gcloud` and `docker-credential-gcr configure-docker` and `gcloud auth login`

Do not forget to install the last `substra` where you run your populate with `pip install -e ../substra`.  
You can then run a `python populate.py -s -a`, go to `http://substra-backend.node-1.com/` and `http://substra-frontend.node-1.com/`


backend/node/generate_nodes.py
```diff
@@ -0,0 +1,65 @@
...
+    for k, v in files.items():
+        filepath = os.path.join(nodes_path, f'{k}.json')
+        with open(filepath, 'w') as f:
+            f.write(json.dumps(v, indent=4))
+
+
+if __name__ == '__main__':
+    orgs = ['owkinMSP', 'chu-nantesMSP', 'clbMSP']  # TODO should be discovered by discovery service
+
+    generate_for_orgs(orgs)
```
jmorel: I re-read the two "generate" scripts (this one and `scripts/generate_nodes_fixtures.py`) and it really looks like they do the exact same thing. It also looks like the `init_node.py` command is a simpler (because only targeting two models and one use case) version of django's loaddata.

In a previous conversation about these, you said:

> The extra django command is also used for admin modification in the future. When you want to add a new organization, you then will be able to add Incoming/Outgoing Node easily. With fixtures, this is not possible.

I agree that only loading committed fixtures the way we do it now isn't dynamic. But this is because of the "committed" part, not because of the "fixtures". Fixtures can be generated and loaded at any time, with any new content, in the same way you propose here to use the generate_nodes.py / init_nodes.py

Could elaborate on the added benefits of these new scripts? Because if there's no absolute need for them I'd rather we keep the already existing ones.

backend/users/views/user.py
```diff
@@ -0,0 +1,81 @@
...
+from rest_framework.response import Response
+from rest_framework_simplejwt.authentication import AUTH_HEADER_TYPES
+from rest_framework_simplejwt.exceptions import TokenError, InvalidToken, AuthenticationFailed
+
+from users.serializers import CustomTokenObtainPairSerializer
+
+import tldextract
+
+
+class UserViewSet(GenericViewSet):
```
jmorel: Using GenericViewSet here seems to add a log of complexity not only in this views file, but in the whole module. 

We could instead use a basic APIView. This way would wouldn't embark the whole CRUD system below the GenericViewSet.

populate.py
```diff
@@ -28,21 +28,17 @@
...
         # the usernames and passwords are defined in the skaffold.yaml file
-        client.add_profile('owkin', 'http://substra-backend.node-1', '0.0',
-                           user='node-1', password='node-1pwd')
-        client.add_profile('chunantes', 'http://substra-backend.node-2', '0.0',
-                           user='node-2', password='node-2pwd')
-        client.add_profile('clb', 'http://substra-backend.node-3', '0.0',
-                           user='node-3', password='node-3pwd')
+        client.add_profile('owkin', USER, PASSWORD, 'http://substra-backend.node-1.com', '0.0')
+        client.add_profile('chunantes', USER, PASSWORD, 'http://substra-backend.node-2.com', '0.0')
+        client.add_profile('clb', USER, PASSWORD, 'http://substra-backend.node-3.com', '0.0')
```
jmorel: The skaffold conf was using separate user/pwd for each node, this uses the same for all nodes (similar to the docker compose profile). This could lead to issues being not detected because the same credentials could be used to the same nodes. 

I think we should keep things as they were (different user/pwd for skaffold and the just one for the docker-compose setup) and then in a separate PR decide on how to harmonize both (I'd be for different user/pwd in all environments).

populate.py
```diff
@@ -28,21 +28,17 @@
...
+        client.add_profile('clb', USER, PASSWORD, 'http://substra-backend.clb.xyz:8002', '0.0')
     if network == 'skaffold':
         # the usernames and passwords are defined in the skaffold.yaml file
-        client.add_profile('owkin', 'http://substra-backend.node-1', '0.0',
-                           user='node-1', password='node-1pwd')
-        client.add_profile('chunantes', 'http://substra-backend.node-2', '0.0',
-                           user='node-2', password='node-2pwd')
-        client.add_profile('clb', 'http://substra-backend.node-3', '0.0',
-                           user='node-3', password='node-3pwd')
+        client.add_profile('owkin', USER, PASSWORD, 'http://substra-backend.node-1.com', '0.0')
```
jmorel: Could we use .xyz for the URLs instead of .com?

backend/libs/expiryTokenAuthentication.py
```diff
@@ -0,0 +1,48 @@
...
+from rest_framework.exceptions import AuthenticationFailed
+
+from datetime import timedelta
+from django.utils import timezone
+
+
+# this return left time
+def expires_at(token):
+    time_elapsed = timezone.now() - token.created
+    left_time = getattr(settings, 'EXPIRY_TOKEN_LIFETIME') - time_elapsed
```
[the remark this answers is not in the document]
GuillaumeCisco: Yes I totally could. I took the habit to always call settings with `getattr` for having a default value.
But I did not use one there.
Either I add a default value or I can use `settings.EXPIRY_TOKEN_LIFETIME`.

Thoughts?

backend/substrapp/tests/common.py
```diff
@@ -1,27 +1,54 @@
...
 
-        self.credentials(HTTP_AUTHORIZATION=basic_auth_header)
+        # create user
+        username = 'substra'
+        password = 'p@$swr0d44'
+        try:
+            with transaction.atomic():
+                User.objects.create_user(username=username, password=password)
+        except Exception:
+            pass
```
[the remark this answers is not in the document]
GuillaumeCisco: I updated it with last proposal.

backend/substrapp/tests/views/tests_views_algo.py
```diff
@@ -95,8 +95,8 @@ def test_algo_list_filter_dual(self):
         with mock.patch('substrapp.views.algo.query_ledger') as mquery_ledger:
             mquery_ledger.return_value = algo
 
-            search_params = '?search=algo%253Aname%253ALogistic%2520regression'
-            search_params += f'%2Calgo%253Aowner%253A{algo[2]["owner"]}'
+            search_params = f'?search=algo%253Aname%253A{urllib.parse.quote(algo[0]["name"])}'
+            search_params += f'%2Calgo%253Aowner%253A{algo[0]["owner"]}'
```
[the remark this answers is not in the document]
GuillaumeCisco: updated

backend/substrapp/views/utils.py
```diff
@@ -48,75 +45,37 @@ def set_headers(self, filelike):
...
         if permission['public']:
             return True
 
-        node_id = user.username
+        if isinstance(user, NodeUser):
+            node_id = user.username
+        else:
+            # for classic user test on current msp id
+            LEDGER = getattr(settings, 'LEDGER')
+            node_id = LEDGER['client']['msp_id']
```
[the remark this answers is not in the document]
GuillaumeCisco: thanks a lot!!!

populate.py
GuillaumeCisco: Yes I will try to modify that.
It was easier to test right now with the same user, but definitively we should add different users.
It will require extra new code though with the job `add_user` and the `add_user` management command used in the start.py.

backend/libs/expiryTokenAuthentication.py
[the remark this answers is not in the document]
jmorel: Ah thanks, I didn't think of the potential default value! I see this pattern used throughout the codebase, no need to change it!

backend/substrapp/views/utils.py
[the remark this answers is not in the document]
jmorel: 👍

backend/node/generate_nodes.py
jmorel: I can understand the feeling, but these are extra changes that are not strictly necessary. 

Since this is already a pretty hefty PR, coupled with other large ones, I really think we should stick to having as few changes as the overall feature requires.

populate.py
jmorel: 👍

backend/node/generate_nodes.py
GuillaumeCisco: I've just thought to another thing about fixtures.
Before starting this PR the `generate_nodes_fixtures.py` script was not present in the code base, only fixtures files were present.
I don't think commiting fixtures files with secret inside is a good practice.

Currently the main code commited these files:
https://github.com/SubstraFoundation/substra-backend/tree/dev/backend/node/fixtures
And refers to the new `generate_nodes_fixtures.py` script: https://github.com/SubstraFoundation/substra-backend/blob/dev/scripts/generate_nodes_fixtures.py

I'm not feeling comfortable by commiting secret even generated from fixtures.
Using dynamic secret not commited feel more secure and clean.

backend/users/views/user.py
GuillaumeCisco: `GenericAPIView` is needed because `serializer_class` and `queryset` is used here.

```
class GenericViewSet(ViewSetMixin, generics.GenericAPIView):
    """
    The GenericViewSet class does not provide any actions by default,
    but does include the base set of generic view behavior, such as
    the `get_object` and `get_queryset` methods.
    """
    pass
```

`get_serializer` is part of `GenericAPIView`
```
    def get_serializer(self, *args, **kwargs):
        """
        Return the serializer instance that should be used for validating and
        deserializing input, and for serializing output.
        """
        serializer_class = self.get_serializer_class()
        kwargs['context'] = self.get_serializer_context()
        return serializer_class(*args, **kwargs)
```

populate.py
GuillaumeCisco: updated, it works great

backend/node/generate_nodes.py
jmorel: That's a good point, thanks.

backend/users/views/user.py
jmorel: I still think we could have gone without the overhead, but I see your point. Let's keep it this way for now.

I guess this was my last comment about the code, I'll move on to testing and then PR approvals. We're almost there!

PR #5, opened 2019-10-25
(WIP) K8S jobs

Kelvin-M:
Work in progress
The aim of this PR is to be able to switch between docker jobs (through docker api via mounted socket) to k8s jobs (through k8s api via network).

- [ ] Settings parameter to switch between docker / k8s task job
- [ ] Build image through kaniko
- [ ] Launch k8s pods instead of pure docker container
- [ ] Mount k8s volumes for ML materials instead of docker binded volume


Kelvin-M:
Close PR as it will not be merged. We surely need to reflect on this feature !

PR #6, opened 2019-10-25
Update volume binding for read write once

Kelvin-M:
For the mission kubernetize hlf we improve substrabac volume handling to allow only read-write-once on it.
It changes the `get_objective` workflow (we do not register the objective in the local db anymore)

Companion PR https://github.com/SubstraFoundation/substra-tests/pull/82

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -13,7 +13,6 @@ import tarfile
 import docker
 import kubernetes
 from checksumdir import dirhash
-from django.core.exceptions import ObjectDoesNotExist
 from django.conf import settings
 from rest_framework.reverse import reverse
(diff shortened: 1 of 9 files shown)


backend/substrapp/tasks/tasks.py
```diff
@@ -310,6 +295,33 @@ def prepare_testtuple_input_models(directory, tuple_):
 
 
 @timeit
+def _put_model(subtuple_directory, model_content, model_hash, traintuple_hash, filename_prefix=''):
```
samlesu: This function seems not needed anymore.

Kelvin-M: 👍

Kelvin-M:
@samlesu Is this PR is good for you ? I've updated along your review :)

PR #7, opened 2019-10-25
Improve logs and compute flow

Kelvin-M:


Kelvin-M:
May someone review this PR :) ?

Kelvin-M:
@GuillaumeCisco : You made a review, but I don't remember what was issue on this feature. Can you remind me to see if I can improve this PR ?

backend/substrapp/tasks/tasks.py
```diff
@@ -117,10 +117,11 @@ def _put_model(subtuple, subtuple_directory, model_content, model_hash, traintup
 
         if not os.path.exists(model_dst_path):
             os.link(model.file.path, model_dst_path)
-        else:
-            # verify that local subtuple model file is not corrupted
-            if get_hash(model_dst_path, traintuple_key) != model_hash:
-                raise Exception('Model Hash in Subtuple is not the same as in local medias')
+
+    finally:
```
GuillaumeCisco: this should not be updated

backend/substrapp/tasks/tasks.py
```diff
@@ -149,15 +150,13 @@ def put_opener(subtuple, subtuple_directory):
     opener_dst_path = path.join(subtuple_directory, 'opener/opener.py')
     if not os.path.exists(opener_dst_path):
         os.link(datamanager.data_opener.path, opener_dst_path)
-    else:
-        # verify that local subtuple data opener file is not corrupted
-        if get_hash(opener_dst_path) != data_opener_hash:
-            raise Exception('DataOpener Hash in Subtuple is not the same as in local medias')
+
+    if get_hash(opener_dst_path) != data_opener_hash:
+        raise Exception('DataOpener Hash in Subtuple is not the same as in local medias')
```
GuillaumeCisco: I don't think this should be updated

backend/substrapp/tasks/tasks.py
```diff
@@ -318,16 +323,17 @@ def prepare_materials(subtuple, tuple_type):
 
     # create subtuple
     subtuple_directory = build_subtuple_folders(subtuple)
+    put_algo(subtuple_directory, algo_content)
     put_opener(subtuple, subtuple_directory)
-    put_data_sample(subtuple, subtuple_directory)
     put_metric(subtuple_directory, metrics_content)
-    put_algo(subtuple_directory, algo_content)
+    put_data_sample(subtuple, subtuple_directory)
```
GuillaumeCisco: Why do we have to change the order?
Does it matter?

Kelvin-M: Nope, it found it better to order like this :)

backend/substrapp/tasks/tasks.py
Kelvin-M: I found it better to check that the model we put is the wanted one in all case.
Is there any issue doing that ?

backend/substrapp/tasks/tasks.py
Kelvin-M: Same as previous answer !

backend/substrapp/tasks/tasks.py
GuillaumeCisco: I don't know if it implies an issue but was not in the original workflow.
I'm not the right person for reviewing this change.
We need someone else help on that

Kelvin-M: Oh, you're right, the workflow will change by doing that.
@camillemarini is it an issue to do the systematic check or I need to revert to the previous workflow ? (check only on download from external node)

Kelvin-M: Solved by reverting this modification.

backend/substrapp/tasks/tasks.py
Kelvin-M: Solved by reverting this modification.

GuillaumeCisco:
approved

PR #12, opened 2019-10-28
Remove duplicate declaration of property

jmorel:
The `rank` property was declared twice in the traintuple serializer, this PR removes the first declaration that was overwritten by the second.


Kelvin-M:
Nice catch !

PR #13, opened 2019-10-29
Composite|Aggregate (tuple/algo)

inalgnu:
- [x] Creation of a new view to create/retrieve/list (a) compositetraintuple(s)
- [x] Creation of a new task corresponding to a compositetraintuple (two in/out models only, one with restricted permissions to the node):
  - preparation of the task
  - execution of the task
  - reporting of the task 
- [x] Modification of prepare task for testtuple to get two inModels in case of a compositetraintuple
Algo
- [x] Create a new view to create/retrieve/list compositealgo
- [x] Generate assets.py (should be done using generate_assets.py)

#### Associated PRs:
- [cli](https://github.com/SubstraFoundation/substra/pull/5)
- [chaincode](https://github.com/SubstraFoundation/substra-chaincode/pull/5)
- [tools](https://github.com/SubstraFoundation/substra-tools/pull/3)


backend/substrapp/migrations/0001_initial.py
```diff
@@ -24,6 +24,15 @@ class Migration(migrations.Migration):
...
         ),
+        migrations.CreateModel(
+            name='CompositeAlgo',
+            fields=[
+                ('pkhash', models.CharField(blank=True, max_length=64, primary_key=True, serialize=False)),
+                ('file', models.FileField(max_length=500, upload_to=substrapp.models.compositealgo.upload_to)),
+                ('description', models.FileField(max_length=500, upload_to=substrapp.models.compositealgo.upload_to)),
+                ('validated', models.BooleanField(default=False)),
+            ],
+        ),
```
GuillaumeCisco: careful.
Migrations should never be modified, instead create a new one with `manage.py makemigrations`

backend/substrapp/views/filters_utils.py
```diff
@@ -100,7 +103,7 @@ def filter_list(object_type, data, query_params):
 
                 filtering_data = filtering_data if filtering_data else []
 
-                if filter_key == 'algo':
+                if filter_key == 'algo' or filter_key == 'compositealgo':
```
GuillaumeCisco:
```suggestion
                if filter_key in ('algo',  'compositealgo'):
```

backend/substrapp/migrations/0001_initial.py
inalgnu: You are right, I planned to do that before committing, but if forgot it. Thank's.

backend/substrapp/serializers/ledger/compositetuple/serializer.py
```diff
@@ -0,0 +1,61 @@
...
+    def get_args(self, validated_data):
+        algo_key = validated_data.get('algo_key')
+        data_manager_key = validated_data.get('data_manager_key')
+        objective_key = validated_data.get('objective_key')
+        rank = validated_data.get('rank', '')
+        rank = '' if rank is None else str(rank)
+        compute_plan_id = validated_data.get('compute_plan_id', '')
+        train_data_sample_keys = validated_data.get('train_data_sample_keys', [])
+        in_head_model_key = validated_data.get('in_head_model_key')
+        in_trunk_model_key = validated_data.get('in_trunk_model_key')
```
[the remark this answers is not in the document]
inalgnu: good catch, I will add that, thank's

backend/substrapp/urls.py
```diff
@@ -27,6 +27,9 @@
 router.register(r'testtuple', TestTupleViewSet, base_name='testtuple')
 router.register(r'task', TaskViewSet, base_name='task')
 router.register(r'compute_plan', ComputePlanViewSet, base_name='compute_plan')
+router.register(r'compositetraintuple', CompositeTraintupleViewSet, base_name='compositetraintuple')
+router.register(r'compositealgo', CompositeAlgoViewSet, base_name='compositealgo')
+router.register(r'compositealgo', CompositeAlgoPermissionViewSet, base_name='compositealgo')
```
jmorel: Why not use underscores in the routes themselves? We already have them for data_manager, data_sample and compute_plan.

Kelvin-M:
Does `assets.py` was generated ? I don't see any modification on `generate_assets.py` 🐙

jmorel:
I second @Kelvin-M point: we have to re-generate `assets.py`. 

This should include new composite models in the `model` list which will make the filters fail on the `/model` view. The reason for that is that the filter_list for model explicitely looks for model['traintuple']['outModel'] while the composite models have neither `traintuple` nor `outModel`. 

I'll make a PR to update the filter_list function and the assets.

samlesu:
LGTM !
This is introducing a lot of code that could be factorized. This could be done later in a separate PR.

backend/substrapp/views/compositetraintuple.py
```diff
@@ -0,0 +1,118 @@
...
+        data = {
+            'algo_key': request.data.get('algo_key'),
+            'data_manager_key': request.data.get('data_manager_key'),
+            'objective_key': request.data.get('objective_key'),
+            'rank': request.data.get('rank'),
+            'compute_plan_id': request.data.get('compute_plan_id', ''),
+            'in_head_model_key': request.data.get('in_head_model_key', ''),
+            'in_trunk_model_key': request.data.get('in_trunk_model_key', ''),
+            'out_trunk_model_permissions': {
+                'public': False,
```
jmorel: Are we sure we want to force the out trunk model to be private? If so shouldn't we reject input data with public trunk model permissions? Otherwise we'll endup just ignoring user input.

backend/substrapp/tests/generate_assets.py
```diff
@@ -28,6 +28,254 @@ def main(network):
     assets['traintuple'] = json.dumps(client.list_traintuple(), indent=4)
     assets['testtuple'] = json.dumps(client.list_testtuple(), indent=4)
     assets['computeplan'] = json.dumps(client.list_compute_plan(), indent=4)
+    assets['compositetraintuple'] = """[
```
Kelvin-M: Can you explain why the composite traintuple asset is hard-coded ?

backend/substrapp/views/compositetraintuple.py
inalgnu: The trunk should not be public but limited for a set of authorized nodes. so yes, be default public is false and only list should be set.

backend/substrapp/tests/generate_assets.py
inalgnu: The easiest way I found to make the unit tests pass (make it work, make it better). 
We are not using populate to make e2e test but substra-tests on this mission.
We should refactor unit tests, to be less dependent on fixtures, until that, I'm open to other solutions with your help.

backend/substrapp/views/compositetraintuple.py
jmorel: Got it thanks, I'll make a comment in the related `substra` PR to update the doc accordingly.

backend/substrapp/tests/generate_assets.py
Kelvin-M: Solved by @jmorel work !

Issue #14, opened 2019-10-29
Rename Middleware files from camelCase to snake_case


samlesu:
To be consistent and to follow PEP8 rules regarding python files naming, all middleware files (located in *backend/libs* folder) should follow the snake_case convention.

hzk120:
@samlesu I would like to take up this issue

samlesu:
Great! Yes, feel welcome to address this issue.

PR #15, opened 2019-10-29
renamed to snake_case

hzk120:
Changed file names from camelCase to snake_case

PR #17, opened 2019-10-29
#14: Renaming files

darshildave:


Kelvin-M:
@darshildave Thank you for your contribution !
There is another PR (https://github.com/SubstraFoundation/substra-backend/pull/15) which is doing also the renaming but also handle the import !
I will let you move to this PR for contribution to avoid duplicated work !

PR #22, opened 2019-11-04
update gitignore file to ignore tmp charts files

inalgnu:
## Context

In Skaffold watch mode, each time you change a file, Skaffold generate some temporary files in the current workspace. Those files could be added mistakenly to the git stage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,3 +116,8 @@ SECRET
 docker/docker-compose-dynamic.yaml
 **/medias/*
 backend/node/nodes/*
+
+
+# charts tmp files
+charts/substra-backend/tmpcharts/*
+skaffold-overrides.yaml


.gitignore
```diff
@@ -115,3 +115,8 @@ SECRET
 # substra-backend files
 docker/docker-compose-dynamic.yaml
 **/medias/*
+
+
+# charts tmp files
+charts/substra-backend/tmpcharts/*
```
Kelvin-M: @ClementGautier In a working setup tmpcharts should be remove automatically ?
If not, nice catch 👍

.gitignore
```diff
@@ -115,3 +115,8 @@ SECRET
...
+skaffold-overrides.yaml
```
Kelvin-M: Can you provide more information on `skaffold-overrides.yaml`, it's not something related to the description of the PR and the file already available in the repo ?

.gitignore
ClementGautier: Yes tmpcharts are supposed to be cleaned up by skaffold unless you kill it violently. I've no problem adding it to the gitignore tho.

.gitignore
inalgnu: sure, I updated the PR description.

PR #33, opened 2019-11-05
Testtuples: Use reference to traintuple instead of using direct reference to model

AurelienGasser:
See also: https://github.com/SubstraFoundation/substra-chaincode/pull/21

--- 

Instead of accessing `model.hash` and `model.storageAddress` directly, use the new `testtuple.traintupleKey` field. Use that traintuple key to obtain the model hash and storage address by requesting the traintuple metadata. 

Since traintuples and testtuples now have a different structure, update the name of function parameters to make it clear which functions take a `traintuple` or a `testtuple` as an argument.

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -94,8 +94,10 @@ def get_models(subtuple):
     else:
         return []
 
+
 def get_traintuple_metadata(traintupleKey):
     return get_object_from_ledger(traintupleKey, 'queryTraintuple')
+
 
 def _put_model(subtuple, subtuple_directory, model_content, model_hash, traintuple_key):
     if not model_content:
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -66,9 +66,8 @@ def get_algo(subtuple):
     return algo_content
 
 
-def _get_model(model):
-    traintuple_hash = model['traintupleKey']
-    traintuple_metadata = get_object_from_ledger(traintuple_hash, 'queryTraintuple')
+def _get_model(traintuple_hash):
+    traintuple_metadata = get_traintuple_metadata(traintuple_hash)
 
     model_content = get_asset_content(
         traintuple_metadata['outModel']['storageAddress'],
@@ -80,10 +79,10 @@ def _get_model(model):
     return model_content
 
 
-def get_model(subtuple):
-    model = subtuple.get('model')
-    if model:
-        return _get_model(model)
+def get_model(testtuple):
+    traintuple_hash = testtuple.get('traintupleKey')
+    if traintuple_hash:
+        return _get_model(traintuple_hash)
     else:
         return None
 
@@ -91,10 +90,12 @@ def get_model(subtuple):
 def get_models(subtuple):
     input_models = subtuple.get('inModels')
     if input_models:
-        return [_get_model(item) for item in input_models]
+        return [_get_model(item['traintupleKey']) for item in input_models]
     else:
         return []
 
+def get_traintuple_metadata(traintupleKey):
+    return get_object_from_ledger(traintupleKey, 'queryTraintuple')
 
 def _put_model(subtuple, subtuple_directory, model_content, model_hash, traintuple_key):
     if not model_content:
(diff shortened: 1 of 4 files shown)


backend/substrapp/tasks/tasks.py
```diff
@@ -66,9 +66,8 @@ def get_algo(subtuple):
     return algo_content
 
 
-def _get_model(model):
-    traintuple_hash = model['traintupleKey']
-    traintuple_metadata = get_object_from_ledger(traintuple_hash, 'queryTraintuple')
+def _get_model(traintuple_hash):
```
Kelvin-M: Mixing `traintuple_hash` and `traintupleKey` argument for the same thing (call of the function and definition of it) is not consistent ! We should chose one name for both.

backend/substrapp/tasks/tasks.py
```diff
@@ -80,22 +79,26 @@ def _get_model(model):
...
+def get_models(traintuple):
+    input_models = traintuple.get('inModels')
     if input_models:
-        return [_get_model(item) for item in input_models]
+        return [_get_model(item['traintupleKey']) for item in input_models]
     else:
         return []
 
 
+def get_traintuple_metadata(traintupleKey):
```
Kelvin-M: I think we should have `traintuple_hash` here :)

backend/substrapp/tasks/tasks.py
```diff
@@ -323,7 +326,9 @@ def prepare_materials(subtuple, tuple_type):
     put_metric(subtuple_directory, metrics_content)
     put_algo(subtuple_directory, algo_content)
     if tuple_type == 'testtuple':
-        put_model(subtuple, subtuple_directory, model_content)
+        traintuple_metadata = get_traintuple_metadata(subtuple['traintupleKey'])
+        model_hash = traintuple_metadata['outModel']['hash']
+        put_model(subtuple, subtuple_directory, model_content, model_hash)
```
Kelvin-M: Be careful, you change the behavior of the previous flow.
Here you're making a call to the ledger to get a info that it was (before https://github.com/SubstraFoundation/substra-chaincode/pull/21) in the testtuple structure.
Is it something that we want ?

AurelienGasser: @Kelvin-M That is intentional (and actually the whole point of this PR!). The rationale for changing the current behavior is that since composite testtuples have 2 models (and not just one), the current testtuple structure (only one model) doesn't work in the composite case.

The approach I propose solves the problem by making testtuples ignorant of the type of model(s) they work with (regular or composite). This might not be the best approach: I'm happy to hear suggestions :)

The main benefit of the proposed solution is that it prevents the introduction of a new `compositeTesttuple` type in the chaincode, which would be essentially a big copy paste of `testTuple` (with a few slight modifications). That would translate to hundreds if not thousands of new lines of code, we'd have to remember to keep the two structures in sync, test them both, etc. In the long term, that's a significant cost.

Regarding the extra ledger call, I think this could be mitigated in future work with caching on the backend (cache the model hash and storage address associated with a traintuple key). AFAIK, the model hash/address never change for a given traintuple.

cc @inalgnu @thibaultrobert since you're most involved in this

backend/substrapp/tasks/tasks.py
AurelienGasser: Fixed https://github.com/SubstraFoundation/substra-backend/pull/33/commits/d11f3b4803767841bdc46f22aa87492c46845cfc

backend/substrapp/tasks/tasks.py
AurelienGasser: Fixed https://github.com/SubstraFoundation/substra-backend/pull/33/commits/d11f3b4803767841bdc46f22aa87492c46845cfc

backend/substrapp/tasks/tasks.py
Kelvin-M: Okay :) 

> The approach I propose solves the problem by making testtuples ignorant of the type of model(s) they work with (regular or composite). This might not be the best approach: I'm happy to hear suggestions :)

Okay, I think the generic tuple PR will need extra work to make it compatible with it as `put_model` work only with one model ?

I think it will be a refacto in generic tuple because of copy/paste code that already happen as we can see with `CompositeAlgo` :)

 Is there a reason for this fix to not be directly included in generic tuple PR, it seems to be very link to it ?

AurelienGasser: @Kelvin-M I think the required chaincode bits have already been implemented as part of the [chaincode PR](https://github.com/SubstraFoundation/substra-chaincode/pull/5): in particular `getCompositeTraintuple` in addition to `getTraintuple`.

The current plan is to have a `traintupleType` field in the `outputTesttuple` returned by the ledger. Depending on this field, the backend can request the ledger for either a regular model or a composite model.

I think the plan is to split `put_model` into a composite and non composite versions (see [this part of the backend composite PR](https://github.com/SubstraFoundation/substra-backend/pull/13/files#diff-1c7f95fdd9883397a5489522f326ccb7R222).

> Is there a reason for this fix to not be directly included in generic tuple PR, it seems to be very link to it ?

My rationale is that since this PR is preparatory work, and it can be be merged into `dev` independently, doing so would reduce the complexity of the (already huge) composite PR. Merging it early would allow us to see any issues with it earlier rather than later.

Kelvin-M: Okay, I will try to test it in multiple setup to see if it's good :)

PR #34, opened 2019-11-06
List and retrieve compute_plans

jmorel:
Add routes to list and retrieve compute plans.

~~This PR depends on a chaincode update: https://github.com/SubstraFoundation/substra-chaincode/pull/22~~ Merged!

In order to add compute plans to the tests assets fixtures, I wanted to run generate_assets.py against  skaffold instead of docker-compose. I updated the script with a `-s` option that will use our standard skafold dev env.

In order to fully test this PR, you'll need the companion PR in the cli/SDK: https://github.com/SubstraFoundation/substra/pull/24

Then:
1. Run the populate.py
2. Try `substra list compute_plan` and `substra get compute_plan <key>`
3. Use DRF's browsable API and navigate to `/compute_plan/`

NB: the assets.py fixtures will have to be re-generated once the related chaincode PR is finalized.

### DIFF
--- a/backend/substrapp/views/computeplan.py
+++ b/backend/substrapp/views/computeplan.py
@@ -5,7 +5,7 @@ from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
 
 from substrapp.serializers import LedgerComputePlanSerializer
-from substrapp.ledger_utils import query_ledger, LedgerError, get_object_from_ledger
+from substrapp.ledger_utils import query_ledger, LedgerError, get_object_from_ledger, LedgerConflict
 from substrapp.views.utils import get_success_create_code, validate_pk
 from substrapp.views.filters_utils import filter_list
 
@@ -24,6 +24,9 @@ class ComputePlanViewSet(mixins.CreateModelMixin,
         args = serializer.get_args(serializer.validated_data)
         try:
             ledger_response = query_ledger(fcn='createComputePlan', args=args)
+        except LedgerConflict as e:
+            error = {'message': str(e.msg), 'pkhash': e.pkhash}
+            return Response(error, status=e.status)
         except LedgerError as e:
             error = {'message': str(e.msg)}
             return Response(error, status=e.status)
--- a/backend/substrapp/views/computeplan.py
+++ b/backend/substrapp/views/computeplan.py
@@ -1,10 +1,13 @@
-from rest_framework import mixins
+import logging
+
+from rest_framework import mixins, status
 from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
 
 from substrapp.serializers import LedgerComputePlanSerializer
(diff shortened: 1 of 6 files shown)


Kelvin-M:
@jmorel Can you put the reference to the chaincode update (issue or PR) when it comes?

jmorel:
@Kelvin-M done!

backend/substrapp/tests/views/tests_views_tuples.py
```diff
@@ -106,7 +107,8 @@ def test_traintuple_list_filter_tag(self):
         with mock.patch('substrapp.views.traintuple.query_ledger') as mquery_ledger:
             mquery_ledger.return_value = traintuple
 
-            search_params = '?search=traintuple%253Atag%253Asubstra'
+            targetTag = '(should fail) My super tag'
+            search_params = '?search=traintuple%253Atag%253A' + urllib.parse.quote_plus(targetTag)
```
jmorel: These changes were borrowed from @AurelienGasser PR https://github.com/SubstraFoundation/substra-backend/pull/33/files

backend/substrapp/views/computeplan.py
```diff
@@ -21,6 +24,9 @@ def create(self, request, *args, **kwargs):
         args = serializer.get_args(serializer.validated_data)
         try:
             ledger_response = query_ledger(fcn='createComputePlan', args=args)
+        except LedgerConflict as e:
+            error = {'message': str(e.msg), 'pkhash': e.pkhash}
+            return Response(error, status=e.status)
```
samlesu: I'm not sure what's the meaning of a conflict error for the compute plan, if it can happen and how we should handle it.
I don't think it makes sense to return to the API user a 'pkhash' in case of conflict.

jmorel: I can guarantee that it does happen :)

You can try to create the same compute plan multiple times in a row and you will get the conflict error. In that very specific case it makes sense to return the pkhash of the compute plan since you now can retrieve it.

That being said, I agree that the behavior is unclear when for example creating a sub-part of an already existing compute plan. But dealing with this is not part of this PR.

Kelvin-M:
I'm testing it:

image

DRF with mod-header `text/html`

jmorel:
@Kelvin-M I updated the PR's description to explain how to test the PR and fixed DRF's browsable API.

backend/substrapp/views/computeplan.py
samlesu: Thanks for the clarification.

samlesu:
LGTM!

It would be good to also update the `substra-tests` repository to get and list compute plans.
We could also improve an existing test submitting a compute plan to ensure it is working as expected.

jmorel:
@samlesu it's done, you can have a look at it here: https://github.com/SubstraFoundation/substra-tests/pull/20

PR #37, opened 2019-11-07
Handle algo and composite algo as similar in filters

jmorel:
In this PR:

Use `composite_algo` and `composite_traintuple` as asset keywords in filters instead of `compositealgo` and `compositetraintuple` to harmonize with the URLs and the denominations of assets in the cli / frontend.

Apply `composite_algo` filters on `algo` assets and vice versa. This means that requesting `/compositealgo?search=algo:key:foo` will return no results instead of previously returning all results.

This last change allows to query both `/algo` and `/compositealgo` with the same search string and be able to treat the responses as part of a larger query on all types of algos.

### DIFF
--- a/backend/substrapp/tests/views/tests_views_compositetraintuples.py
+++ b/backend/substrapp/tests/views/tests_views_compositetraintuples.py
@@ -106,7 +106,7 @@ class CompositeTraintupleViewTests(APITestCase):
         with mock.patch('substrapp.views.compositetraintuple.query_ledger') as mquery_ledger:
             mquery_ledger.return_value = compositetraintuple
 
-            search_params = '?search=compositetraintuple%253Atag%253Asubstra'
+            search_params = '?search=composite_traintuple%253Atag%253Asubstra'
             response = self.client.get(url + search_params, **self.extra)
             r = response.json()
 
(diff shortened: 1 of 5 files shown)


backend/substrapp/views/filters_utils.py
```diff
@@ -62,6 +62,17 @@ def get_filters(query_params):
     return filters
 
 
+def _same_nature(filter_key, object_type):
+    if filter_key == object_type:
+        return True
+
+    # algo and composite algos are of the same nature
+    if {filter_key, object_type} <= {'algo', 'composite_algo'}:
```
inalgnu: could be simplified
```suggestion
    return {filter_key, object_type} <= {'algo', 'composite_algo'}
```

jmorel: Done, thanks!

backend/substrapp/tests/views/tests_views_compositetraintuples.py
```diff
@@ -117,7 +117,7 @@ def test_compositetraintuple_list_filter_compute_plan_id(self):
         with mock.patch('substrapp.views.compositetraintuple.query_ledger') as mquery_ledger:
             mquery_ledger.return_value = compositetraintuple
             compute_plan_id = get_compute_plan_id(compositetraintuple)
```
maeldebon: Shouldn't we rename 'compositetraintuple' to 'composite_traintuple' here as well?

jmorel: I agree that in the backend knowing when there needs to be an underscore when and there shouldn't be one is pretty hard. In this case it seems to be in line with variable names in similar files (see for example `tests_views_datamanager.py`)

PR #44, opened 2019-11-14
Generic composite: fix model filtering

jmorel:
Until now, models (as returned by `queryModels` and `/model`) could only be filtered by traintuple.outModel.hash: `/model?search=model:foo:bar` would return objects whose `traintuple.outModel.hash`property is `bar`. You could even replace `foo` by anything, it would only check `traintuple.outModel.hash`.

This had the following issues:
* as soon as you had a compositeTraintuple and a search string containing a condition starting with `model:`, the endpoint failed (trying to access the `traintuple` attr of the model)
* if this issue was solved, a compositeTraintuple doesn't have an outModel but rather and outHeadModel and an outTrunkModel

I updated the filter_list method so that model:attr:value would filter on the model's traintuple or compositeTraintuple depending on what's available.

It is however not possible anymore to filter by outModel.hash since it's not a direct attr of the (composite)traintuple.


backend/substrapp/views/filters_utils.py
```diff
@@ -71,6 +71,13 @@ def _same_nature(filter_key, object_type):
     return {filter_key, object_type} <= {'algo', 'composite_algo'}
 
 
+def _model_tuple(model):
+    if 'compositeTraintuple' in model:
+        return model['compositeTraintuple']
+
+    return model['traintuple']
+
+
```
Kelvin-M: can be replace by `model.get('compositeTraintuple', model['traintuple'])` no ?

Kelvin-M: So it will be reduce to `x.get('compositeTraintuple', x['traintuple'])`

jmorel: We'll also get aggregateTuples very soon, so that would become `model.get('aggregateTuple', model.get('compositeTraintuple', model['traintuple'])`. If that's still fine with you I'll update the code with your suggestion, otherwise I'll leave it as it is.

backend/substrapp/views/filters_utils.py
```diff
@@ -71,6 +71,13 @@ def _same_nature(filter_key, object_type):
     return {filter_key, object_type} <= {'algo', 'composite_algo'}
 
 
+def _model_tuple(model):
```
samlesu: Would be good to find a better name for this function: `_get_tuple_from_model()` and maybe add a small comment or a docstring?

```
# XXX This function is required in order to be able to list models using their key from the frontend.
```

It will be more explicit and would improve the code readability. Otherwise, it will be hard in my opinion to remember what's the purpose of this code.

populate.py
```diff
@@ -379,25 +396,9 @@ def do_populate():
...
-            print('')
-            print('-' * 100)
-            print(colored(json.dumps(res, indent=2), 'green'))
-            print(colored(json.dumps(res_t, indent=2), 'yellow'))
-        else:
-            print('.', end='', flush=True)
-
-        time.sleep(3)
+    wait_for_tuple(client, 'traintuple', traintuple_key, 'green')
+    wait_for_tuple(client, 'testtuple', testtuple_key, 'yellow')
```
samlesu: 👍

backend/substrapp/views/filters_utils.py
Kelvin-M: So maybe it better to do 
``` python
def _model_tuple(model):
    if 'compositeTraintuple' in model:
        return model['compositeTraintuple']
    elif 'aggregateTuple' in model
        return model['compositeTraintuple']
    elif 'traintuple' in model:
        return model['traintuple']
    else:
        raise NotImplementedError

```
Or something like that

jmorel: 👍

backend/substrapp/views/filters_utils.py
jmorel: 👍

jmorel:
@Kelvin-M @samlesu I implemented your suggestions, could you re-review this PR?

PR #45, opened 2019-11-14
Fix and improve docker start

Kelvin-M:


Kelvin-M:
@GuillaumeCisco Can you review it ? It fixes docker start and speed up docker stop of substra backend

PR #46, opened 2019-11-15
Do not include tests in containers

jmorel:
The .dockerignore will prevent tests from being included in containers. Which also means that while running `skaffold dev`, we'll be able to edit tests without seeing the images being rebuilt continuously.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+backend/node/tests
+backend/substrapp/tests


Kelvin-M:
Nice catch ! Maybe we can add all the folders that do not need to be in the image too !

jmorel:
Sure thing! Which folders were you thinking of? Currently the only folders copied into the image are:

```
backend/libs
backend/substrapp
backend/events
backend/backend
backend/node
backend/node-register
backend/users
```

This looks already pretty minimal, but if you think of anything that could be removed, I'm all for it!

Kelvin-M:
No, I think it's good like this ! Thank you :)

PR #55, opened 2019-11-27
Refactor CA Cert secret names

AurelienGasser:
See main PR: https://github.com/SubstraFoundation/hlf-k8s/pull/16

### DIFF
--- a/charts/substra-backend/values.yaml
+++ b/charts/substra-backend/values.yaml
@@ -77,7 +77,7 @@ persistence:
 
 # Secrets names
 secrets:
-  ca: hlf-ca
+  caCert: hlf-cacert
   user:
     # Certificate, saved under key 'cert.pem'
     cert: hlf-idcert
--- a/charts/substra-backend/templates/configmap-server.yaml
+++ b/charts/substra-backend/templates/configmap-server.yaml
@@ -26,7 +26,7 @@ data:
           "external": {{ .Values.peer.port }}
         },
         "docker_core_dir": "/var/hyperledger/fabric_cfg",
-        "tlsCACerts": "/var/hyperledger/ca/ca-cert.pem",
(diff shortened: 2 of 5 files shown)


charts/substra-backend/values.yaml
```diff
@@ -86,7 +86,7 @@ secrets:
   # This should include the Orderer TLS 'cacert.pem'
   orderer:
     tls:
-      ca: hlf-client-tlsrootcert
+      caCert: hlf-cacert
```
AurelienGasser: Note that the value we put here doesn't seem to matter.

I set it to:

```
caCert: hlf-idkey
```

and the populate happily succeeded :man_shrugging: 

My understanding is that the backend needs to communicate securely with the orderer in order to commit transactions. But somehow, even with a bad root CA file, the backend-orderer connection succeeds. Maybe something is misconfigured? Or am I missing something?

This is probably out of the scope of this PR, but I thought I'd report it :)

GuillaumeCisco: The backend will use a peer which will need to connect to the orderer for `invoke` transaction.
This connection need a CA cert.
Maybe in the backend code, the peer uses a CA cert referenced from elsewhere

Kelvin-M: I think we get it from the discovery : https://github.com/SubstraFoundation/substra-backend/blob/dev/backend/backend/settings/deps/ledger.py#L111:#L171

We can remove this as we only stuff to connect to our peer !

Kelvin-M: I'm doing a PR for that :)

Kelvin-M:
@AurelienGasser  and @GuillaumeCisco  Do you agree on this version. It's not a refacto but more a renaming !

PR #62, opened 2019-12-05
set local folders for all train like tuples

samlesu:
Local folder is defined for:
- aggregatetuple
- traintuple
- composite_traintuple

All these tuples will share the same local folder (inside a compute plan).

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -578,8 +578,8 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id, 
         model_path: {'bind': model_folder, 'mode': 'rw'}
     }
 
-    # local volume for training subtuple in compute plan
-    if compute_plan_id is not None and tuple_type == TRAINTUPLE_TYPE:
+    # local volume for train like tuples in compute plan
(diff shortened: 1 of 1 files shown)


backend/substrapp/tasks/tasks.py
```diff
@@ -32,6 +32,8 @@
 COMPOSITE_TRAINTUPLE_TYPE = 'compositeTraintuple'
 TESTTUPLE_TYPE = 'testtuple'
 
+TRAIN_LIKE_TUPLE_TYPES = (TRAINTUPLE_TYPE, AGGREGATETUPLE_TYPE, COMPOSITE_TRAINTUPLE_TYPE)
```
Kelvin-M: Is still relevant to have `TYPE`  in those variables
`TRAIN_LIKE_TUPLE_TYPES` could be simplified as something linked with the local volume to make it explicit

samlesu: Yes I think it is relevant to add the  `_TYPE` suffix to these global variables: it defines the type of tuples returned by the ledger.

If that's ok with you, I will remove the `TRAIN_LIKE_TUPLE_TYPES` as it seems confusing.

PR #63, opened 2019-12-05
Fail fast: remove evaluation for train like tuples

samlesu:
Tasks:
- [x] views: remove link between train like tuples and objective
- [x] views: add link between testtuple and objective
- [x] views: update compute plan accordingly (remove global objective_key and define it at the tuple level)
- [x] compute: remove evaluation step for train like tuples
- [x] update populate

Companion PRs:
- [chaincode](https://github.com/SubstraFoundation/substra-chaincode/pull/50)
- [substra](https://github.com/SubstraFoundation/substra/pull/52)
- [tests](https://github.com/SubstraFoundation/substra-tests/pull/25)

### DIFF
--- a/backend/substrapp/ledger_utils.py
+++ b/backend/substrapp/ledger_utils.py
@@ -252,7 +252,6 @@ def log_success_tuple(tuple_type, tuple_key, res):
                 'hash': res["end_model_file_hash"],
                 'storageAddress': res["end_model_file"],
             },
-            'perf': float(res["global_perf"]),
             'log': '',
         }
 
@@ -276,7 +275,6 @@ def log_success_tuple(tuple_type, tuple_key, res):
                 'hash': res["end_trunk_model_file_hash"],
                 'storageAddress': res["end_trunk_model_file"],
             },
-            'perf': float(res["global_perf"]),
             'log': '',
         }
 
@@ -288,7 +286,6 @@ def log_success_tuple(tuple_type, tuple_key, res):
                 'hash': res["end_model_file_hash"],
                 'storageAddress': res["end_model_file"],
             },
-            'perf': float(res["global_perf"]),
             'log': '',
         }
 
--- a/backend/substrapp/serializers/ledger/aggregatetuple/serializer.py
+++ b/backend/substrapp/serializers/ledger/aggregatetuple/serializer.py
@@ -8,7 +8,6 @@ from .tasks import createLedgerAggregateTupleAsync
 
 class LedgerAggregateTupleSerializer(serializers.Serializer):
     algo_key = serializers.CharField(min_length=64, max_length=64)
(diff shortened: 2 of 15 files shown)


samlesu:
> The "create" method for train like tuple views still reference the objective_key

Good catch, thanks!

jmorel:
Since computePlanID and rank have been added to testtuples in the related chaincode PR could add it here as well?

samlesu:
> This function should be refactored too much if/else that make it complicated to understand
> https://github.com/SubstraFoundation/substra-backend/blob/979d73be5b1671de4c5a3b84d72aca69c721c522/backend/substrapp/tasks/tasks.py#L550#L715

Yes, you're completely right, this function and I think the whole module is too complicated to read and to maintain. I think the whole module requires a big refactoring and this refactoring should be done separately. This could be done later, a refactoring session is scheduled in the coming weeks.

Would that be ok for you?

Kelvin-M:
Adding another if seems to add more complexity to this function.
As this function could be refactor without border effect, I think it could be done in this PR.

samlesu:
> Adding another if seems to add more complexity to this function.
> As this function could be refactor without border effect, I think it could be done in this PR.

Correct me if I'm wrong, but no `if` clause has been added to the function pointed in your link (that is to say the `_do_task()` function).
An `if` clause has been added to the `prepare_materials()` function but this one is pretty simple as it has been slightly refactored recently.

Most of the if/else clauses of the function `_do_task()` have been added in an another PR. In my opinion, review should be related to the scope of the reviewed PR only.

Kelvin-M:
@samlesu https://github.com/SubstraFoundation/substra-backend/pull/63/files#diff-1c7f95fdd9883397a5489522f326ccb7R691 (line 691 is between line 550 and line 715)

samlesu:
> @samlesu https://github.com/SubstraFoundation/substra-backend/pull/63/files#diff-1c7f95fdd9883397a5489522f326ccb7R691 (line 691 is between line 550 and line 715)

Yes. This change is not adding an `if` clause, it was already there. It is changing the line

```python
if tuple_type == AGGREGATETUPLE_TYPE:  # skip evaluation
```

to:

```python
if tuple_type != TESTTUPLE_TYPE:  # skip evaluation
```

Kelvin-M:
A modification in git is considered as a removal and a addition :)

samlesu:
> Isn't the compute plan in the populate broken here? It shouldn't have a global algo key as input anymore but each traintuple should have its own algo. It doesn't looks like it is the case.

Indeed, good catch! But that's as far as I know not related to this PR.

camillemarini:
Hello @Kelvin-M @samlesu @inalgnu Let's keep this refactoring for another PR :-) I think we need this PR merged as soon as possible.

PR #66, opened 2019-12-06
Retry a failed task up to 5 times

AurelienGasser:
This PR introduces an auto-retry mechanism for compute tasks: up to 5 retries (i.e. 6 attempts in total), with 2 secs between each attempt.

The retry logic is handled by Celery. Most of the changes in this PR are just moving things around. The actual retry logic is only the `raise self.retry(...)` line (!)

:warning: **Following this PR, compute task results and exceptions are now passed back to Celery**. Prior to this PR, the result (or exception) from `do_task` was passed to `log_success_tuple` (or `log_fail_tuple`) and didn't leave the `compute_task` function. That's not the case anymore. Now the result/exception is returned to Celery (via `return` or `raise`). Celery, in turn, either retries, calls `on_succeed` or calls `on_failure`. I don't think that's a problem because 1. I'm not sure those return/exceptions are stored in DB anyway by celery and 2. Even if they are stored in DB, they are stored on the current node only. It just seemed significant enough of a change for me to mention it.

Also note that I moved around some seemingly unrelated settings and test code: that's because `compute_task(...)` now calls `settings.TASK`, so I had to do move a few settings things around in order to accomodate for that. I also had to remove the `getattr` mock because this mock was always returning `MEDIA_ROOT` (which turned out to not be necessary anyway)

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -15,6 +15,7 @@ from django.conf import settings
 from rest_framework.reverse import reverse
 from celery.result import AsyncResult
 from celery.exceptions import Ignore
+from celery.task import Task
 
 from backend.celery import app
 from substrapp.utils import get_hash, get_owner, create_directory, uncompress_content
@@ -348,7 +349,7 @@ def prepare_data_sample(directory, tuple_):
 
 def build_subtuple_folders(subtuple):
     # create a folder named `subtuple['key']` in /medias/subtuple/ with 5 subfolders opener, data, model, pred, metrics
-    subtuple_directory = path.join(getattr(settings, 'MEDIA_ROOT'), 'subtuple', subtuple['key'])
+    subtuple_directory = get_subtuple_directory(subtuple)
     create_directory(subtuple_directory)
 
     for folder in ['opener', 'data', 'model', 'pred', 'metrics']:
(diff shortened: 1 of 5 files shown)


inalgnu:
Hello @AurelienGasser,
Maybe we could use http://docs.celeryproject.org/en/latest/userguide/tasks.html#automatic-retry-for-known-exceptions that could give us more flexibility (configure retry for each task, define exception wanted, etc). What do you think?

Also, this wasn't mentioned in the specification but retry number should be added as a setting value, so we can be able to configure the backend without modifying its source code.

AurelienGasser:
@inalgnu Thanks, I'll try using the celery approach

AurelienGasser:
@inalgnu this is ready for another review

backend/substrapp/tasks/tasks.py
```diff
@@ -443,9 +446,15 @@ def prepare_tuple(subtuple, tuple_type):
...
 
 
-@app.task(bind=True, ignore_result=False)
+@app.task(bind=True, ignore_result=False,
+          autoretry_for=(Exception,),
+          retry_kwargs={'max_retries': CELERY_TASK_MAX_RETRIES, 'countdown': CELERY_TASK_RETRY_DELAY_SECONDS})
 def compute_task(self, tuple_type, subtuple, compute_plan_id):
 
+    attempt = self.request.retries + 1
+    max_attempts = CELERY_TASK_MAX_RETRIES + 1
```
inalgnu: why do we need to handle this manually? isn't celery autoretry with max_tries sufficient?

backend/substrapp/tasks/tasks.py
```diff
@@ -462,6 +471,11 @@ def compute_task(self, tuple_type, subtuple, compute_plan_id):
         error_code = compute_error_code(e)
         logging.error(error_code, exc_info=True)
 
+        # Trigger the celery auto-retry if this isn't our last attempt
```
inalgnu: I think we should just let the exception raise here? without catching? otherwise celery doc recommand to use self.retry 
https://docs.celeryproject.org/en/latest/userguide/tasks.html#using-a-custom-retry-delay

AurelienGasser: Fixed https://github.com/SubstraFoundation/substra-backend/pull/66/commits/0462f788d66e64ff1ccc02b9d751ac377b434ea2

backend/substrapp/tasks/tasks.py
AurelienGasser: Fixed: https://github.com/SubstraFoundation/substra-backend/pull/66/commits/0462f788d66e64ff1ccc02b9d751ac377b434ea2

PR #69, opened 2019-12-09
Fix local volume for tuples

samlesu:
Local volume was created only for tuple of rank 0.
This is causing a bug as the tuple of rank 0 may be executed on a different worker than the worker for the tuple of rank 1.

Regression introduced by commit b640a5518ff0e5381b10dd4aad40a0a1489abed5

Can be tested with substra-tests:
```
pytest tests/test_execution_compute_plan.py::test_compute_plan_aggregate_composite_traintuples
```

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -537,9 +537,11 @@ def do_task(subtuple, tuple_type):
             remove_subtuple_materials(subtuple_directory)
             if rank == -1:
                 volume_id = f'local-{compute_plan_id}-{org_name}'
-                local_volume = client.volumes.get(volume_id=volume_id)
                 try:
+                    local_volume = client.volumes.get(volume_id=volume_id)
                     local_volume.remove(force=True)
+                except docker.errors.NotFound:
+                    pass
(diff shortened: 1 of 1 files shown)


backend/substrapp/tasks/tasks.py
```diff
@@ -581,10 +583,10 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
     # local volume for train like tuples in compute plan
     if compute_plan_id is not None and tuple_type != TESTTUPLE_TYPE:
         volume_id = f'local-{compute_plan_id}-{org_name}'
-        if rank == 0:
-            client.volumes.create(name=volume_id)
-        else:
+        try:
```
Kelvin-M: The behavior is changed here, local volume should only be created on rank == 0

samlesu: Thanks for asking, I wasn't 100% sure regarding this.

It has been confirmed with @camillemarini, in case of a compute plan, local volume will be created if it does not exist whatever the rank value.

samlesu:
Fix #69

PR #71, opened 2019-12-09
Fix MVCC Read Conflict retry

### LINKED ISSUE #67: MVCC read conflict when reporting the success of a task
Hello,

It looks like a `MVCC READ CONFLICT` occurred when registering the success of a training task. Attached are the logs. 
[logs.txt](https://github.com/SubstraFoundation/substra-backend/files/3937155/logs.txt)
For the context, I had submitted two `composite_traintuples` and an `aggregatetuple` with `inmodels` the two `composite_traintuples`. The error occured for one of the two `composite_traintuple`.

I am not sure to understand why we get it here. Is it because when logging the success of this task, when updating the status of its child (the `aggregatetuple`), it reads an older version of the other parents (the other `composite_traintuple`) which status has been updated at the same time ?

Couldn't we add a retry mechanism in order to fix this ?

samlesu:
Fix #67

- fix parsing of fabric-sdk-py exceptions to raise internal MVCC Read Conflict errors
- enable retry for `log_*` in case of Ledger errors by default (was disabled in dev)

### DIFF
--- a/backend/backend/settings/dev.py
+++ b/backend/backend/settings/dev.py
@@ -13,7 +13,7 @@ TASK = {
     'CACHE_DOCKER_IMAGES': to_bool(os.environ.get('TASK_CACHE_DOCKER_IMAGES', False)),
 }
 
-LEDGER_CALL_RETRY = False  # Overwrite the ledger setting value
+# LEDGER_CALL_RETRY = False  # uncomment to overwrite the ledger setting value
 
 # Database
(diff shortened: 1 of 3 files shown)


samlesu:
Do  not merge, may cause issue with unit tests.

inalgnu:
@samlesu I have restarted the build, it was stopped without errors.

samlesu:
> Seems that tests are whiling without output, maybe because the retry is enabled one tests . I think this should be disabled in tests settings.

Thanks. I will do that!

PR #75, opened 2019-12-10
Composite separate methods

samlesu:
Should be merged after other PR fixing the populate.


samlesu:
I need to regenerate the assets to ensure the unit tests are passing.

Issue #77, opened 2019-12-11
Error "Rendezvous of RPC that terminated - failed to connect to all addresse"


jmorel:
While trying to add a bunch of assets, I regularly end up with the following message (as returned by the SDK):

```
ERROR    substra.sdk.rest_client:rest_client.py:116 Requests error status 400: {"message":"<_Rendezvous of RPC that terminated with:\n\tstatus = StatusCode.UNAVAILABLE\n\tdetails = \"failed to connect to all addresses\"\n\tdebug_error_string = \"{\"created\":\"@1576051644.524371597\",\"description\":\"Failed to pick subchannel\",\"file\":\"src/core/ext/filters/client_channel/client_channel.cc\",\"file_line\":3934,\"referenced_errors\":[{\"created\":\"@1576051644.524366407\",\"description\":\"failed to connect to all addresses\",\"file\":\"src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc\",\"file_line\":393,\"grpc_status\":14}]}\"\n>"}
```

samlesu:
Thanks @jmorel do you have the associated traceback in the backend?

Kelvin-M:
@jmorel Do you manage to reproduce it ? It would be nice to have the backend traceback in order to be able to find a fix for this issue.
I suspect that we will need to change the way we initialize the fabric client !

jmorel:
I used to have the issue while creating assets on the demo setup, I tried reproducing it locally but can't. I'll run my test case against the demo as soon as it's available again and fetch the traceback.

jmorel:
I got the error again while running a very big (about 1400 tuples) compute plan on the demo env, here is the stacktrace from org4-worker:

```
[2020-01-08 08:04:52,559: ERROR/ForkPoolWorker-1] <_Rendezvous of RPC that terminated with:
    status = StatusCode.UNAVAILABLE
    details = "failed to connect to all addresses"
    debug_error_string = "{"created":"@1578470692.558659550","description":"Failed to pick subchannel","file":"src/core/ext/filters
/client_channel/client_channel.cc","file_line":3934,"referenced_errors":[{"created":"@1578470692.558654186","description":"failed t
o connect to all addresses","file":"src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc","file_line":393,"grpc_s
tatus":14}]}"
>
Traceback (most recent call last):
[...]

PR #79, opened 2019-12-11
(wip) try to fix grpc issue #77

inalgnu:


charts/substra-backend/templates/deployment-server.yaml
```diff
@@ -36,7 +36,7 @@ spec:
         {{- if eq .Values.backend.settings "prod" }}
         args: ["-c", "uwsgi --http :{{ .Values.backend.service.port }} --module backend.wsgi --static-map /static=/usr/src/app/backend/statics --master --processes 4 --threads 2 --need-app --env DJANGO_SETTINGS_MODULE=backend.settings.server.prod "]
         {{- else }}
-        args: ["-c", "python manage.py runserver --noreload 0.0.0.0:{{ .Values.backend.service.port }} --settings=backend.settings.server.{{ .Values.backend.settings }}"]
+        args: ["-c", "unset http_proxy; unset https_proxy; python manage.py runserver --noreload 0.0.0.0:{{ .Values.backend.service.port }} --settings=backend.settings.server.{{ .Values.backend.settings }}"]
```
Kelvin-M: I think this cannot work as `http_proxy` and `https_proxy` are not defined in our local setup but also on servers.
What is the reason of this `unset` of those env. variables ?

inalgnu: Regarding this issue seems, that was solved by unsetting http(*)_proxy variables, this branch was simply trying to test this behavior https://github.com/grpc/grpc/issues/9987#issuecomment-370156099

Kelvin-M: Okay, I as read the issue, it did not seems related to our problem (even if the name of the error code is the same). It seems that the issue is caused by the setup of their proxy !

Kelvin-M:
This PR could be closed as the fix suggested has no relation with our issue, no ?

PR #80, opened 2019-12-11
fix discovery

inalgnu:


charts/substra-backend/templates/deployment-server.yaml
```diff
@@ -36,7 +36,7 @@ spec:
         {{- if eq .Values.backend.settings "prod" }}
         args: ["-c", "uwsgi --http :{{ .Values.backend.service.port }} --module backend.wsgi --static-map /static=/usr/src/app/backend/statics --master --processes 4 --threads 2 --need-app --env DJANGO_SETTINGS_MODULE=backend.settings.server.prod "]
         {{- else }}
-        args: ["-c", "python manage.py runserver --noreload 0.0.0.0:{{ .Values.backend.service.port }} --settings=backend.settings.server.{{ .Values.backend.settings }}"]
+        args: ["-c", "unset http_proxy; unset https_proxy; python manage.py runserver --noreload 0.0.0.0:{{ .Values.backend.service.port }} --settings=backend.settings.server.{{ .Values.backend.settings }}"]
```
Kelvin-M: This seems unrelated to the fix of the discovery when `discovery['members']` can be empty.
I saw it in this PR https://github.com/SubstraFoundation/substra-backend/pull/79 but I didn't understand why we are trying to unset a env. variable which is never set, can you provide more info on it ?

inalgnu: @Kelvin-M yes, you are right this is unrelated to discovery (bad base branch) for this PR.

Kelvin-M:
@inalgnu Can you base this branch on `master` instead of `fix-grpc-issue` as this fix should be merged :) ?

inalgnu:
@Kelvin-M done

PR #90, opened 2019-12-16
remove local folder following a chaincode event

AurelienGasser:
Companion PR: https://github.com/SubstraFoundation/substra-chaincode/pull/64

### DIFF
--- a/backend/events/apps.py
+++ b/backend/events/apps.py
@@ -37,12 +37,6 @@ def get_event_loop():
         loop.close()
 
 
-def tuple_get_worker(tuple_type, _tuple):
-    if tuple_type == 'aggregatetuple':
-        return _tuple['worker']
-    return _tuple['dataset']['worker']
-
-
 def on_event(cc_event, block_number, tx_id, tx_status):
     payload = json.loads(cc_event['payload'])
     owner = get_owner()
@@ -97,6 +91,12 @@ def handle_tuple(owner, worker_queue, tuple_type, tx_status, _tuple):
         task_id=key,
         queue=worker_queue
     )
+
+
+def tuple_get_worker(tuple_type, _tuple):
+    if tuple_type == 'aggregatetuple':
+        return _tuple['worker']
+    return _tuple['dataset']['worker']
 
 
 def handle_compute_plan(owner, worker_queue, tx_status, cp):
--- a/backend/events/apps.py
+++ b/backend/events/apps.py
@@ -48,59 +48,66 @@ def on_tuples(cc_event, block_number, tx_id, tx_status):
     owner = get_owner()
     worker_queue = f"{LEDGER['name']}.worker"
 
-    for tuple_type, _tuples in payload.items():
+    for key, items in payload.items():
 
-        if tuple_type not in ['traintuple', 'compositeTraintuple', 'aggregatetuple']:
+        if not items:
             continue
 
-        if not _tuples:
-            continue
+        if key in ['traintuple', 'compositeTraintuple', 'aggregatetuple']:
+            tuple_type = key
+            for tuple in items:
+                handle_tuple(owner, worker_queue, tx_status, tuple_type, tuple)
 
-        for _tuple in _tuples:
-            key = _tuple['key']
-            status = _tuple['status']
+        elif key == 'computePlan':
+            for cp in items:
+                handle_compute_plan(owner, worker_queue, tx_status, cp)
 
-            if tx_status != 'VALID':
-                logger.error(
-                    f'Failed transaction on task {key}: type={tuple_type}'
-                    f' status={status} with tx status: {tx_status}')
-                continue
-            logger.info(f'Processing task {key}: type={tuple_type} status={status}')
-
-            if status != 'todo':
-                continue
-
-            if tuple_type is None:
-                continue
-
-            tuple_owner = tuple_get_worker(tuple_type, _tuple)
-            if tuple_owner != owner:
-                logger.debug(f'Skipping task {key}: owner does not match'
-                             f' ({tuple_owner} vs {owner})')
-                continue
-
-            if AsyncResult(key).state != 'PENDING':
-                logger.info(f'Skipping task {key}: already exists')
-                continue
-
-            prepare_tuple.apply_async(
-                (_tuple, tuple_type),
-                task_id=key,
-                queue=worker_queue
-            )
-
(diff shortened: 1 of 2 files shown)


backend/events/apps.py
```diff
@@ -85,6 +85,24 @@ def on_tuples(cc_event, block_number, tx_id, tx_status):
...
+            handle_compute_plan_events(payload['ComputePlans'])
+
+
+def handle_compute_plan_events(cps):
+    for cp in cps:
+        handle_compute_plan_event(cp)
+
+
+def handle_compute_plan_event(cp):
+    id = cp['computePlanID']
```
inalgnu: be aware that id is a builtin callable in Python and here you override it.
```python
print(id(1))
id=1
print(id(1))
```

AurelienGasser: wow, I had no idea, thanks!

AurelienGasser: Fixed: https://github.com/SubstraFoundation/substra-backend/pull/90/commits/14804e7fd2ff97ad2ac2599bed5bcb4ce6a01baa

backend/events/apps.py
```diff
@@ -85,6 +85,24 @@ def on_tuples(cc_event, block_number, tx_id, tx_status):
...
+            handle_compute_plan_events(payload['ComputePlans'])
+
+
+def handle_compute_plan_events(cps):
+    for cp in cps:
+        handle_compute_plan_event(cp)
+
+
+def handle_compute_plan_event(cp):
+    cpID = cp['computePlanID']
```
samlesu: Python variable should be defined in lower case with underscore (https://www.python.org/dev/peps/pep-0008/#function-and-variable-names).

For reserved keywords like id, you could add a suffix `_` (https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments).

backend/events/apps.py
```diff
@@ -85,6 +85,24 @@ def on_tuples(cc_event, block_number, tx_id, tx_status):
                 queue=worker_queue
             )
 
+        if 'ComputePlans' in payload:
```
thibaultrobert: I think you have to handle the tuple/compute plan differentiation before that. At [L51](https://github.com/SubstraFoundation/substra-backend/blob/14804e7fd2ff97ad2ac2599bed5bcb4ce6a01baa/backend/events/apps.py#L51) we have no means yet to differentiate between tuple type lists and the compute plan. So the compute plan will be treated as a tuple and I think it will failed somewhere after that.

backend/events/apps.py
AurelienGasser: Good point!

backend/events/apps.py
AurelienGasser: Good catch!

backend/events/apps.py
AurelienGasser: Fixed: https://github.com/SubstraFoundation/substra-backend/pull/90/commits/bb2b5ba2928b13420a71cf7a94268225934ff8f0

backend/events/apps.py
AurelienGasser: Fixed: https://github.com/SubstraFoundation/substra-backend/pull/90/commits/c29f41d927b2b1db1a802818f349a47d976714cf

backend/events/apps.py
```diff
@@ -49,6 +49,10 @@ def on_tuples(cc_event, block_number, tx_id, tx_status):
     worker_queue = f"{LEDGER['name']}.worker"
 
     for tuple_type, _tuples in payload.items():
+
+        if tuple_type not in ['traintuple', 'compositeTraintuple', 'aggregatetuple']:
+            continue
```
samlesu: What's the reason for adding these 2 lines? It looks like testtuple won't be executed anymore.

EDIT:
I get it, it's because we want to filter `computePlan` (it has been added in the companion PR). In this case I would add a comment to explain the goal of this filter. Or maybe do the opposite filter: `if tuple_type == 'computePlan'` ?

backend/events/apps.py
```diff
@@ -85,6 +89,24 @@ def on_tuples(cc_event, block_number, tx_id, tx_status):
                 queue=worker_queue
             )
 
+        if 'computePlan' in payload:
+            handle_compute_plan_events(payload['computePlan'])
```
samlesu: Should be outside of the for loop (on payload) and this should be ignored if `tx_status` is not valid.

Could be worth to handle the `tx_status` at the beginning of the function. It was done inside the for loop to easily log all skipped tuples.

backend/events/apps.py
```diff
@@ -85,6 +89,24 @@ def on_tuples(cc_event, block_number, tx_id, tx_status):
...
+
+
+def handle_compute_plan_events(cps):
+    for cp in cps:
+        handle_compute_plan_event(cp)
+
+
+def handle_compute_plan_event(cp):
+    id_ = cp['computePlanID']
+    status = cp['status']
```
thibaultrobert: Since we will send an event for every change of status (`Done`, `Failed` and `Cancelled`), we should enforce here that we remode the local folder only in the `Done` case. For now, it's important for the project that we keep this local folder.

thibaultrobert: Nevermind it seems I can't read more than two line... Sorry  about that

backend/events/apps.py
AurelienGasser: I'm going to convert this code to a switch/case, it'll make things much clearer!

backend/events/apps.py
AurelienGasser: Good points, thank you!

AurelienGasser:
@thibaultrobert @samlesu Would you guys mind having a look at the latest commit (https://github.com/SubstraFoundation/substra-backend/pull/90/commits/b6d98db09e19098fbfec0afab4e52d044ee8d67a) and let me know what you think?

Kelvin-M:
@AurelienGasser : Is this work still relevant ?

AurelienGasser:
I think this work has been de-prioritized. @thibaultrobert might know more.

thibaultrobert:
It's indeed not a priority anymore. The [PR in the chaincode](https://github.com/SubstraFoundation/substra-chaincode/pull/64) is also on hold. And since there have been massive changes to the compute plan mechanism in the chaincode it would be quite a bit of work to restart it.
From my perspective this can be closed for now.

PR #91, opened 2019-12-17
Fix 500 on wrong pk for leaderbord

### LINKED ISSUE #47: CLI: Invalid key in the leaderboard command
When I try to use the `leaderboard` command on a key which doesn't exists, I get the following error:
`Error: Request failed: InternalServerError: 500 Server Error: Internal Server Error for url: http://substra-backend.node-1.com/objective/foo/leaderboard/?sort=desc`

Maybe we could have a more specific error explaining that the key doesn't exists?

Kelvin-M:
Fix #47

### DIFF
--- a/backend/substrapp/views/objective.py
+++ b/backend/substrapp/views/objective.py
@@ -231,9 +231,13 @@ class ObjectiveViewSet(mixins.CreateModelMixin,
 
     @action(detail=True, methods=['GET'])
     def leaderboard(self, request, pk):
-        validate_pk(pk)
-
         sort = request.query_params.get('sort', 'desc')
+
+        try:
+            validate_pk(pk)
+        except Exception as e:
+            return Response({'message': str(e)}, status=status.HTTP_400_BAD_REQUEST)
+
         try:
             validate_sort(sort)
         except Exception as e:


backend/substrapp/views/objective.py
```diff
@@ -231,10 +231,9 @@ def data(self, request, *args, **kwargs):
 
     @action(detail=True, methods=['GET'])
     def leaderboard(self, request, pk):
-        validate_pk(pk)
-
         sort = request.query_params.get('sort', 'desc')
         try:
+            validate_pk(pk)
```
GuillaumeCisco: `validate_pk(pk)` should be in its own try/except.

A try clause  should only try on one call.
Maybe creating a dedicated method:
```python
def method(self, request, pk):
    validate_pk(pk)
    sort = request.query_params.get('sort', 'desc')
    validate_sort(sort)
```

is better.
Then try/except on this method for handling all the possible cases.

Kelvin-M: Thank you, I will update my PR :)

Kelvin-M: I don't know if we need to create a dedicated method for this.
I'd rather prefer to use a try/except

backend/substrapp/views/objective.py
```diff
@@ -246,6 +250,9 @@ def leaderboard(self, request, pk):
             })
         except LedgerError as e:
             return Response({'message': str(e.msg)}, status=e.status)
+        except Exception as e:
```
GuillaumeCisco: if using one dedicated try/except, do you still need this?
If yes, using a dedicated method will be better for avoiding code repetition and keeping DRY.

Kelvin-M: Updated !

PR #92, opened 2019-12-18
Remove intermediate model

thibaultrobert:
- [ ] Handle both composite and aggregate case
- [ ] Work only in a compute plan situation
- [ ] Create a setting to enable feature flipping
- [ ] Check how a node respond to a request on a deleted model (We might want to throw a ` 410  `Gone` HTTP status)

Link with SubstraFoundation/substra-tests#44


thibaultrobert:
This feature has been post-pone for now so I will close this PR.

PR #93, opened 2019-12-20
K8s fetch secrets

inalgnu:


backend/substrapp/tasks/tasks.py
```diff
@@ -625,6 +626,33 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+        try:
+            secret_namespace = os.getenv('K8S_SECRET_NAMESPACE', 'default')
+            secrets = k8s_client.list_namespaced_secret(secret_namespace, label_selector=label_selector)
+        except ApiException as e:
+            logging.error(f'failed to fetch namespaced secrets {secret_namespace} with selector {label_selector}')
+            raise e
+
+        chainkeys_directory = get_chainkeys_directory()
+        if not os.path.exists(chainkeys_directory):
+            os.mkdir(chainkeys_directory)
```
samlesu: As far as I know, `mkdir` does not allow to create more than 1 directory. Maybe use `os.makedirs()` instead? https://docs.python.org/3/library/os.html#os.makedirs

backend/substrapp/tasks/tasks.py
```diff
@@ -625,6 +626,33 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+            with open(path.join(chainkeys_directory, 'chainkeys.json'), 'w') as file:
+                file.write(str(secrets))
```
samlesu: To create a json file, you should the json library:
```python
json.dump(fh, secrets)
```

Otherwise you will create a text file

backend/substrapp/tasks/tasks.py
```diff
@@ -625,6 +626,33 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+
+        volumes[chainkeys_directory] = {'bind': '/sandbox/chainkeys', 'mode': 'rw'}
+
+    # Environment current node index
+    node_index = os.getenv('NODE_INDEX', None)
```
samlesu:
```python
node_index = os.getenv('NODE_INDEX')
```

Returning None is already the default behaviour of the getenv() method.

backend/substrapp/tasks/tasks.py
```diff
@@ -625,6 +626,33 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+    if node_index:
+        environment = {
```
samlesu: `environment` is declared as a `list` object line 601. It looks confusing.

What about creating an empty dict line 601 and updating it here?
```python
environment["NODE_INDEX"] = node_index
```

backend/substrapp/tasks/utils.py
```diff
@@ -236,11 +237,12 @@ def compute_docker(client, resources_manager, dockerfile_path, image_name, conta
...
         'network_mode': 'none',
         'privileged': False,
-        'cap_drop': ['ALL']
+        'cap_drop': ['ALL'],
+        'environment': environment
     }
 
     if gpu_set is not None:
-        task_args['environment'] = {'NVIDIA_VISIBLE_DEVICES': gpu_set}
+        task_args['environment'].append({'NVIDIA_VISIBLE_DEVICES': gpu_set})
```
samlesu: if environment is a dict, you should use `.update()` instead of `.append()`

backend/substrapp/tests/tests_tasks.py
```diff
@@ -337,7 +337,7 @@ def test_compute_docker(self):
         hash_docker = uuid.uuid4().hex
         compute_docker(client, self.ResourcesManager,
                        self.subtuple_path, 'test_compute_docker_' + hash_docker,
-                       'test_compute_docker_name_' + hash_docker, None, None)
+                       'test_compute_docker_name_' + hash_docker, None, None, environment=[])
```
samlesu: environment should be a dict object?

inalgnu: done

backend/substrapp/tasks/tasks.py
```diff
@@ -625,6 +626,31 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
             client.volumes.create(name=volume_id)
         model_volume[volume_id] = {'bind': '/sandbox/local', 'mode': 'rw'}
 
+    if settings.TASK['CHAINKEYS_ENABLED']:
+        label_selector = f"compute_plan={subtuple['tag']}"
+        secrets = []
```
samlesu: declaring this variable is useless

backend/substrapp/tasks/tasks.py
```diff
@@ -625,6 +626,31 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+
+        k8s_client = get_k8s_client()
+        try:
+            secret_namespace = os.getenv('K8S_SECRET_NAMESPACE', 'default')
+            secrets = k8s_client.list_namespaced_secret(secret_namespace, label_selector=label_selector)
+        except ApiException as e:
+            logging.error(f'failed to fetch namespaced secrets {secret_namespace} with selector {label_selector}')
+            raise e
```
samlesu: This should be done inside the if condition line 642 as the secrets need to be fetched at the initialisation step only.

backend/substrapp/tasks/tasks.py
```diff
@@ -625,6 +626,31 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+            label_selector = f"compute_plan={subtuple.get('tag')}"
+
+            k8s_client = get_k8s_client()
+            try:
+                secrets = k8s_client.list_namespaced_secret(secret_namespace, label_selector=label_selector)
+            except ApiException as e:
+                logging.error(f'failed to fetch namespaced secrets {secret_namespace} with selector {label_selector}')
+                raise e
+
+            os.mkdirs(chainkeys_directory)
```
GuillaumeCisco:
```python
>>> import os
>>> os.mkdirs
Traceback (most recent call last):
  File "<input>", line 1, in <module>
AttributeError: module 'os' has no attribute 'mkdirs'
```

backend/substrapp/tasks/tasks.py
```diff
@@ -625,6 +626,31 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+            with open(path.join(chainkeys_directory, 'chainkeys.json'), 'w') as f:
+                json.dump(f, secrets)
```
GuillaumeCisco: `secrets` will be a `V1SecretList` of objects of this kind:
```python
[
{           'api_version': None,
            'data': {'key': 'q1mjlIXJIEt6ADKrrHglXITCavH6AWq3sv83+XknsMs='},
            'kind': None,
            'metadata': {'annotations': None,
                         'cluster_name': None,
                         'creation_timestamp': datetime.datetime(2020, 1, 13, 10, 23, 4, tzinfo=tzlocal()),
                         'deletion_grace_period_seconds': None,
                         'deletion_timestamp': None,
                         'finalizers': None,
                         'generate_name': None,
                         'generation': None,
                         'initializers': None,
                         'labels': {'compute_plan': 'compute-plan-1'},
                         'managed_fields': None,
                         'name': 'compute-plan-1-a-next-ck',
                         'namespace': 'participant',
                         'owner_references': None,
                         'resource_version': '3622243',
                         'self_link': '/api/v1/namespaces/participant/secrets/compute-plan-1-a-next-ck',
                         'uid': '87bf9600-5529-4a3e-92b2-3a1a68cd256d'},
            'string_data': None,
            'type': 'Opaque'},
]
```
Dumping it to json will fail:
```
TypeError: Object of type 'V1SecretList' is not JSON serializable
```
`V1SecretList` are not iterable.
You need to first transform it:
```python
secrets = secrets.to_dict()['items']
```

Furthermore, I don't think this is really what we want here...
Secrets are base64 encoded, we need to base64 decode with `b64decode`

Updated code would be:
```python
secrets  = [{s['metadata']['name']: b64decode(s['data']['key'])} for s in secrets.to_dict()['items']]
```
You will then get a list of secrets with their name and their bytes values.
Each secret value is the representation of a very big integer to bytes.
You can load this value thanks to `int.from_bytes` method, this way:
```python
int.from_bytes(secret, 'big')
```

Remember json cannot dump bytes values, so you need to bring them back to int values:
Final code is:
```python
secrets = [{s['metadata']['name']: int.from_bytes(b64decode(s['data']['key']), 'big')} for s in secrets.to_dict()['items']]
json.dump(f, secrets)
```

Last thing to keep in mind, these secrets values do not incorporate their sign for the masking part, you need to compute them in the substra-backend or algo (or secure aggregation library) part.

backend/substrapp/tasks/tasks.py
```diff
@@ -625,6 +627,41 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+
+            k8s_client = get_k8s_client()
+            try:
+                secrets = k8s_client.list_namespaced_secret(secret_namespace, label_selector=label_selector)
+            except ApiException as e:
+                logging.error(f'failed to fetch namespaced secrets {secret_namespace} with selector {label_selector}')
+                raise e
+
+            secrets = [{s['metadata']['name']: int.from_bytes(b64decode(s['data']['key']), 'big')}
+                       for s in secrets.to_dict()['items']]
```
GuillaumeCisco: After some thinking,

```python
secrets = {s['metadata']['name']: int.from_bytes(b64decode(s['data']['key']), 'big') for s in secrets.to_dict()['items']}
```
seems better as it will create a dict instead of a list of dict.
Will be easier then for accessing secrets.
Example:
```python
secrets['name_of_secret']
```

backend/substrapp/tasks/tasks.py
```diff
@@ -625,6 +627,41 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+
+            os.makedirs(chainkeys_directory)
+            with open(path.join(chainkeys_directory, 'chainkeys.json'), 'w') as f:
+                json.dump(f, secrets)
+
+            for secret in secrets:
+                try:
+                    k8s_client.delete_namespaced_secret(secret['name'], secret_namespace)
```
GuillaumeCisco: Does `name` key still exist in this dictionary?

inalgnu: fixed 5da9db0

backend/substrapp/tasks/tasks.py
inalgnu: updated 531f9b2

backend/substrapp/tasks/tasks.py
```diff
@@ -625,6 +627,41 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+            secrets = {s['metadata']['name']: int.from_bytes(b64decode(s['data']['key']), 'big')
+                       for s in secrets.to_dict()['items']}
+
+            os.makedirs(chainkeys_directory)
+            with open(path.join(chainkeys_directory, 'chainkeys.json'), 'w') as f:
+                json.dump(f, secrets)
+
+            for secret in secrets:
+                try:
+                    k8s_client.delete_namespaced_secret(secret['metadata']['name'], secret_namespace)
```
GuillaumeCisco: secrets will be a dict of pairs key/value with a depth of 1.
It is defined line 644.
Its keys will be the results of `s['metadata']['name']`.

Therefore, `secret['metadata']['name']` cannot exist.

inalgnu: updated 2c1ba92

PR #94, opened 2020-01-03
Retry ledger call on phantom read conflict

jmorel:
@inalgnu and I ran into a new error from the ledger: "PHANTOM READ CONFLICT".

This PR adds a retry in case this error is returned by the ledger. The status code attached to the error is really just a placeholder, we didn't take the time to properly investigate what the code should be. If you have an opinion about it please feel free to share!

### DIFF
--- a/backend/substrapp/ledger_utils.py
+++ b/backend/substrapp/ledger_utils.py
@@ -74,6 +74,10 @@ class LedgerUnavailable(LedgerError):
     status = status.HTTP_503_SERVICE_UNAVAILABLE
 
 
+class LedgerPhantomReadConflictError(LedgerError):
+    status = status.HTTP_412_PRECONDITION_FAILED
+
+
 class LedgerBadResponse(LedgerResponseError):
     pass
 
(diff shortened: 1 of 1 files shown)


GuillaumeCisco:
Looks like an issue in substra design.
Something with a range query seems to have been updated between chaincode execution and commit.

More information: https://stackoverflow.com/questions/51950726/hyperledger-phantom-read-conflict-vs-mvcc-read-conflict

Kelvin-M:
I agree with Guillaume that it may come from a substra design. @jmorel Do we have an issue in the chaincode to expose this problem ?
Nevertheless, we should handle this issue :) Thank you @jmorel for this update =)

jmorel:
I'm sorry, I don't have a specific scenario to generate these errors, they just "appears" while trying to add A LOT of tuples to the platform.

samlesu:
I do agree with @Kelvin-M and @GuillaumeCisco. It will be interesting to investigate more and it could be caused by a wrong design.

As a quick term solution, a retry should fix it.

PR #96, opened 2020-01-06
Add statics shared volume

### LINKED ISSUE #64: Statics files are generated in the init container
With the new version of server deployment, the static files are generated in the init container in production but are not present in the server container at the end. We should remove this init container or we should copy those static in the server container once generated

Kelvin-M:
Fix #64


Kelvin-M:
@GuillaumeCisco It fixes the issue created when adding the init container for the django static files.

PR #100, opened 2020-01-09
add chainkeys env variables

inalgnu:


charts/substra-backend/values.yaml
```diff
@@ -197,3 +197,13 @@ celeryworker:
...
+
+extraEnv:
+  - name: CHAINKEYS_FOR_TRAINING_TASKS_ENABLED
+    value: False
+  - name: NODE_INDEX
+    value: 1
+  - name: K8S_SECRET_NAMESPACE
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.namespace
```
ClementGautier: the default values should be empty in the opensourced chart. Put those specific values in the infrastructure repo

charts/substra-backend/values.yaml
```diff
@@ -197,3 +197,7 @@ celeryworker:
   tolerations: []
 
   affinity: {}
+
+extraEnv:
```
ClementGautier:
```suggestion
extraEnv: []
```

Issue #103, opened 2020-01-16
Error GRPC "StatusCode.CANCELLED"


jmorel:
While adding a bunch of tuples individually while a whole other bunch of tuples is being executed, I got the following error in the worker's log. The traceback is incomplete (keyboard mishaps and the logs were gone), but I hope this is enough to investigate

```
[2020-01-16 10:09:16,738: ERROR/ForkPoolWorker-1] exception calling callback for <Future at 0x7f79d456f630 state=finished raised _MultiThreadedRendezvous>
Traceback (most recent call last):
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/local/lib/python3.6/dist-packages/aiogrpc/utils.py", line 126, in _next
    return next(self._iterator)
  File "/usr/local/lib/python3.6/dist-packages/grpc/_channel.py", line 416, in __next__
    return self._next()
  File "/usr/local/lib/python3.6/dist-packages/grpc/_channel.py", line 703, in _next
    raise self
grpc._channel._MultiThreadedRendezvous: <_MultiThreadedRendezvous of RPC that terminated with:
    status = StatusCode.CANCELLED
    details = "Locally cancelled by application!"
    debug_error_string = "None"
>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.6/concurrent/futures/_base.py", line 324, in _invoke_callbacks
    callback(self)
  File "/usr/lib/python3.6/asyncio/futures.py", line 417, in _call_set_state
    dest_loop.call_soon_threadsafe(_set_state, destination, source)
  File "/usr/lib/python3.6/asyncio/base_events.py", line 637, in call_soon_threadsafe
    self._check_closed()
  File "/usr/lib/python3.6/asyncio/base_events.py", line 377, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
[2020-01-16 10:09:16,743: ERROR/ForkPoolWorker-1] exception calling callback for <Future at 0x7f79d458abe0 state=finished raised _MultiThreadedRendezvous>
Traceback (most recent call last):
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/local/lib/python3.6/dist-packages/aiogrpc/utils.py", line 126, in _next
[...]

PR #105, opened 2020-01-17
Remove status check before retries

inalgnu:
fix https://github.com/SubstraFoundation/substra-backend/issues/103

### DIFF
--- a/backend/substrapp/ledger_utils.py
+++ b/backend/substrapp/ledger_utils.py
@@ -244,7 +244,7 @@ def invoke_ledger(*args, **kwargs):
     return _invoke_ledger(*args, **kwargs)
 
 
-@retry_on_error()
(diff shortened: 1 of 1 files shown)


backend/substrapp/ledger_utils.py
```diff
@@ -267,37 +267,6 @@ def get_object_from_ledger(pk, query):
...
-        status = tuple_['status']
-        if status == expected_status:
-            return
-
-        logger.error(
-            f'{tuple_type} {tuple_key} wrong status {status}: expecting {expected_status} (trie {trie})'
-        )
-        trie += 1
-
-    raise LedgerTimeoutNotHandled(
```
samlesu: I think this exception is now unused and could be removed.

inalgnu: good catch, removed here 93c430b

PR #106, opened 2020-01-17
Stop catching raw Exceptions

samlesu:


samlesu:
@jmorel yes, 100% agree with you that's a draft to ensure exceptions are properly displayed.

It's ready to use, not ready to review

samlesu:
@inalgnu the PR has been rebased on top of master

backend/substrapp/views/compositetraintuple.py
```diff
@@ -110,8 +104,6 @@ def retrieve(self, request, *args, **kwargs):
...
             data = self._retrieve(pk)
         except LedgerError as e:
-            return Response({'message': str(e.msg)}, status=e.status)
-        except Exception as e:
-            return Response({'message': str(e)}, status.HTTP_400_BAD_REQUEST)
-        else:
-            return Response(data, status=status.HTTP_200_OK)
+            raise exceptions.from_ledger_error(e)
+
+        return Response(data, status=status.HTTP_200_OK)
```
jmorel: This is the only use of `from_ledger_error `, everywhere else you kept the `return Response({'message': str(e.msg)}, status=e.status)`.
I think we should either use from_ledger_error everywhere or nowhere. In this specific case, I'd be for using it nowhere and then in a separate PR converting all manual returns when catching an exception with raising excepptions.from_ledger_error.

backend/substrapp/exceptions.py
```diff
@@ -0,0 +1,60 @@
...
+        super().__init__(message)
+
+    def response(self):
+        """Get HTTP Response from error instance."""
+        return drf.response.Response(self.error, status=self.status)
+
+
+class BadRequestError(_ApiError):
+    status = drf.status.HTTP_400_BAD_REQUEST
+    message = "Bad request."""
```
jmorel: Typo: `"""` instead of `"`

backend/substrapp/views/compositetraintuple.py
samlesu: 100% agree, my last commit was not intended to be part of this PR (it's named wip...).
I will move it to a distinct PR.

samlesu: Done

backend/substrapp/exceptions.py
samlesu: Done

backend/substrapp/views/filters_utils.py
```diff
@@ -96,8 +98,12 @@ def _get_model_tuple(model):
 
 
 def filter_list(object_type, data, query_params):
-
-    filters = get_filters(query_params)
+    try:
+        filters = get_filters(query_params)
+    except Exception as e:
+        # TODO add better filters parsing to avoid this catch all
+        logging.exception(e)
```
inalgnu: I think it would be useful here to log query_params that led to this exception, what do you think?

thibaultrobert: Either in the log or in the BadRequestError raised under.

samlesu: Yes good idea, we should add the query_params to the BadRequestError.
In an ideal world we would have a better parsing of the query args with better validation error.

samlesu: Done, it has been added to the BadRequest response and to the exception log message.

PR #107, opened 2020-01-21
fix ledger error inheritance

samlesu:
- status code was incorrect due the from_response class method returning the base class
- can be tested with: https://github.com/SubstraFoundation/substra-tests/pull/58 (end to end test is not working without this fix)

EDIT:
This fix enables us to discover a new bug: we should not retry ledger request on BadResponseError. We could not see this bug before as the LedgerError was not properly created.

### DIFF
--- a/backend/substrapp/ledger_utils.py
+++ b/backend/substrapp/ledger_utils.py
@@ -14,86 +14,108 @@ logger = logging.getLogger(__name__)
 
 
 class LedgerError(Exception):
-    status = status.HTTP_400_BAD_REQUEST
+    """Base error from ledger."""
+    # FIXME the base error status code should be 500, the chaincode is currently
+    #       responding with 500 status code for some 400 errors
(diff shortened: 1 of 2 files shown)


samlesu:
As pointed by @jmorel this fix is highlighting a new bug:

We retry requests on LedgerBadResponse. This should not be the case as we should not retry invalid requests.

jmorel:
To be more precise: retrying a LedgerBadResponse means that a call that is never meant to pass will be retried again and again, never returning to the user. This behaviour broke the tests `test_permission_invalid_node_id` and `test_compute_plan_circular_dependency_failure` from `substra-tests`.

samlesu:
After re-reading the code with @inalgnu, it is normal to retry on a BadResponse (it is not the same as BadRequest!).

But this is showing an another bug, BadRequest is mapped to BadResponse.

PR #110, opened 2020-01-22
Add ledger wait for event timeout in settings

samlesu:
~~Also decrease its value from 45 to 5.~~
This setting can also be set using an env var.


samlesu:
You're right this could be done in a separate PR after more investigation.

samlesu:
Following your comments, I'm keeping the original value.

PR #112, opened 2020-01-22
Fix issue with celerybeat periodic tasks

Kelvin-M:
When celerybeat sends periodic tasks, it try to push the celery result to django-db. It will fail as celerybeat is not connected to the posgresql db.

As we do not need this database, we could avoid, for the celerybeat to send the result of celery task.

Moreover, I added a setting to setup the frequency of the periodic task (in case of event hub failure we can increase frequency).

### DIFF
--- a/backend/backend/settings/dev.py
+++ b/backend/backend/settings/dev.py
@@ -27,6 +27,8 @@ MEDIA_ROOT = os.environ.get('MEDIA_ROOT', os.path.join(PROJECT_ROOT, f'medias/{O
 SITE_HOST = f'substra-backend.{ORG_NAME}.xyz'
 SITE_PORT = DEFAULT_PORT
 DEFAULT_DOMAIN = os.environ.get('DEFAULT_DOMAIN', f'http://{SITE_HOST}:{SITE_PORT}')
+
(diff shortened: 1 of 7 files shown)


Kelvin-M:
Can someone review it ? @GuillaumeCisco  @samlesu  @inalgnu

backend/backend/settings/dev.py
```diff
@@ -27,6 +27,8 @@
 SITE_HOST = f'substra-backend.{ORG_NAME}.xyz'
 SITE_PORT = DEFAULT_PORT
 DEFAULT_DOMAIN = os.environ.get('DEFAULT_DOMAIN', f'http://{SITE_HOST}:{SITE_PORT}')
+
+CELERY_RESULT_BACKEND = 'django-db'
```
GuillaumeCisco: What's the purpose of putting this assignment in dev and prod settings instead of common?

Kelvin-M: Because we use `common.py` for the celery beat.
Consequently, if so, as celerybeat is not connected to the postgresql db. when launching the periodic task, it will try to send the result to the django-db and fails !

PR #117, opened 2020-01-23
Fix model download permissions

jmorel:
This PR depends on https://github.com/SubstraFoundation/substra-chaincode/pull/79

Companion PR in tests: https://github.com/SubstraFoundation/substra-tests/pull/70

- Adds a new check on permission preventing model downloads from nodes who don't have permissions
- Relies on a new chaincode smartContract to retrieve the permissions associated with a model.

To reproduce the original issue:
- run `test_permissions_denied_model_process` from substra-tests
- get the model url from the traintuple whose algo is `xxxxxxx_test_permissions_denied_model_process - Algo 1` This model should only be processable by MyOrg2MSP.
- curl the URL using the node1 to node2 credentials: `MyOrg1MSP:nodeSecret2w1`

You should get a 200 without the fix and a 403 with the fix.

### DIFF
--- a/backend/substrapp/views/model.py
+++ b/backend/substrapp/views/model.py
@@ -1,4 +1,3 @@
-import os
 import tempfile
 import logging
 from django.http import Http404
@@ -11,7 +10,7 @@ from node.authentication import NodeUser
 from substrapp.models import Model
 from substrapp.serializers import ModelSerializer
 from substrapp.ledger_utils import query_ledger, get_object_from_ledger, LedgerError
-from substrapp.views.utils import CustomFileResponse, validate_pk, get_remote_asset, PermissionMixin
+from substrapp.views.utils import validate_pk, get_remote_asset, PermissionMixin
 from substrapp.views.filters_utils import filter_list
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


backend/substrapp/views/model.py
```diff
@@ -120,15 +119,22 @@ class ModelPermissionViewSet(PermissionMixin,
 
     queryset = Model.objects.all()
     serializer_class = ModelSerializer
-    ledger_query_call = 'queryModelDetails'
+    ledger_query_call = 'queryModelPermissions'
 
-    @action(detail=True)
-    def file(self, request, *args, **kwargs):
+    def _has_access(self, user, asset):
```
samlesu: It seems wrong to modify a private method in a child class located in a different module.

backend/substrapp/views/utils.py
```diff
@@ -89,34 +89,62 @@ def download_file(self, request, django_field, ledger_field=None):
...
-            for header in r.headers:
-                # We don't use hop_by_hop headers since they are incompatible
-                # with WSGI
-                if not is_hop_by_hop(header):
-                    response[header] = r.headers.get(header)
+            response = self._download_remote_file(ledger_field, asset)
+
+        return response
+
+    def download_local_file(self, request, django_field, ledger_field=None):
```
samlesu: Not related to this PR but `download` is misleading (same for all `download_*` methods). What about `stream`, `forward`, `send`?

backend/substrapp/views/model.py
jmorel: I made it into a public method, much cleaner. Thanks!

backend/substrapp/views/utils.py
jmorel: I agree we could find a better name, for a later refactoring maybe!

jmorel:
@samlesu it should be fixed now, my rename was a bit hasty.

Kelvin-M:
From the test with the old version, it seems we can download a model from the drf api with an user account

Kelvin-M:
You should change all `isinstance` to `type` checking as isinstance will work also for subclass

Kelvin-M:
@jmorel  I've done some modification here f15384e767f7ed49a9aea52ee9033e14e8c059a1

jmorel:
@Kelvin-M Good catch, thanks a lot!

jmorel:
@Kelvin-M Tested and approved: I can only retrieve the model with the owner's node credentials, and never with user credentials.

PR #121, opened 2020-01-24
Fix model filtering

### LINKED ISSUE #120: Error 500 when filtering models on objective
Here is an example URL that will trigger the error 500:

`http://substra-backend.node-1.com/model/?search=objective%253Aname%253ASkin%252520Lesion%252520Classification%252520Objective`

Bug found by @GuillaumeCisco

jmorel:
Fix #120 

All tuples previously had an objective attached to then when created so that they could have a training score. Now that this has disappeared, only testtuples have objective attached.

As a result, only models that have testtuples can be filtered by objective.

### DIFF
--- a/backend/substrapp/tests/views/tests_views_compositealgo.py
+++ b/backend/substrapp/tests/views/tests_views_compositealgo.py
@@ -19,7 +19,7 @@ from substrapp.ledger_utils import LedgerError
 from substrapp.utils import get_hash
 
 from ..common import get_sample_composite_algo, AuthenticatedClient
-from ..assets import objective, datamanager, compositealgo, algo
+from ..assets import objective, datamanager, compositealgo, algo, model
 
 MEDIA_ROOT = "/tmp/unittests_views/"
 
@@ -137,6 +137,24 @@ class CompositeAlgoViewTests(APITestCase):
             r = response.json()
 
             self.assertIn('Malformed search filters', r['message'])
+
+    def test_composite_algo_list_filter_model(self):
+        url = reverse('substrapp:composite_algo-list')
+        done_model = [
+            m for m in model if 'compositeTraintuple' in m and m['compositeTraintuple']['status'] == 'done'
+        ][0]
+
+        with mock.patch('substrapp.views.compositealgo.query_ledger') as mquery_ledger, \
+                mock.patch('substrapp.views.filters_utils.query_ledger') as mquery_ledger2:
+            mquery_ledger.return_value = compositealgo
+            mquery_ledger2.return_value = model
+
+            pkhash = done_model['compositeTraintuple']['outTrunkModel']['outModel']['hash']
+            search_params = f'?search=model%253Ahash%253A{pkhash}'
+            response = self.client.get(url + search_params, **self.extra)
+            r = response.json()
+
+            self.assertEqual(len(r[0]), 1)
 
     def test_composite_algo_retrieve(self):
         dir_path = os.path.dirname(os.path.realpath(__file__))
--- a/backend/substrapp/tests/views/tests_views_datamanager.py
+++ b/backend/substrapp/tests/views/tests_views_datamanager.py
@@ -16,7 +16,7 @@ from substrapp.utils import get_hash
 
 
 from ..common import get_sample_datamanager, AuthenticatedClient
-from ..assets import objective, datamanager, traintuple, model
+from ..assets import objective, datamanager, model
 
 MEDIA_ROOT = "/tmp/unittests_views/"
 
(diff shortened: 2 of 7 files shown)


backend/substrapp/views/filters_utils.py
```diff
@@ -193,8 +193,11 @@ def filter_list(object_type, data, query_params):
                         hashes = [x['key'] for x in filtering_data]
 
                         if object_type == 'model':
-                            filtered_list = [x for x in filtered_list
-                                             if _get_model_tuple(x)['objective']['hash'] in hashes]
+                            filtered_list = [
+                                x for x in filtered_list
+                                if (x['testtuple'] and x['testtuple']['objective'] and
+                                    x['testtuple']['objective']['hash'] in hashes)
```
Kelvin-M: Nice lazy evaluation but is it possible that a `testtuple` has no objective ?

jmorel: Objective can now only be attached to testtuple and no other tuple, so there's no implicit way of inferring the objective. It has to be there.

jmorel:
@Kelvin-M tests updated!

Kelvin-M:
> Fix #120
> 
> All tuples previously had an objective attached to then when created so that they could have a training score. Now that this has disappeared, only testtuples have objective attached.
> 
> As a result, only models that have testtuples can be filtered by objective.

Is https://github.com/SubstraFoundation/substra-backend/blob/master/backend/substrapp/tests/assets.py up to date if we considerer your remark ?

jmorel:
It's not. I'll update it...

jmorel:
@Kelvin-M it's done, could you review it? I had to change quite a few things unfortunately, we had a few filter test broken.

Kelvin-M:
Yes, I will do it this afternoon :)

backend/substrapp/tests/views/tests_views_compositealgo.py
```diff
@@ -138,6 +138,24 @@ def test_composite_algo_list_filter_objective_fail(self):
 
             self.assertIn('Malformed search filters', r['message'])
 
+    def test_composite_algo_list_filter_model(self):
+        url = reverse('substrapp:composite_algo-list')
+        done_model = [
```
GuillaumeCisco: shouldn't we use use a `filter` here?
```
done_models = filter(lambda m: 'compositeTraintuple' in m and m['compositeTraintuple']['status'] == 'done' > 18, model)
done_model = done_models[0]
```

backend/substrapp/tests/views/tests_views_datamanager.py
```diff
@@ -103,12 +103,15 @@ def test_datamanager_list_filter_objective(self):
 
     def test_datamanager_list_filter_model(self):
         url = reverse('substrapp:data_manager-list')
-        done_model = [m for m in model if 'traintuple' in m and m['traintuple']['status'] == 'done'][0]
+        done_model = [
+            m for m in model
+            if 'traintuple' in m and m['traintuple']['status'] == 'done' and m['testtuple']['traintupleKey']
+        ][0]
```
GuillaumeCisco: same here

backend/substrapp/tests/views/tests_views_objective.py
```diff
@@ -133,11 +133,14 @@ def test_objective_list_filter_datamanager(self):
 
     def test_objective_list_filter_model(self):
         url = reverse('substrapp:objective-list')
-        done_model = [m for m in model if 'traintuple' in m and m['traintuple']['status'] == 'done'][0]
+        done_model = [
+            m for m in model
+            if 'traintuple' in m and m['traintuple']['status'] == 'done' and m['testtuple']['objective']
+        ][0]
```
GuillaumeCisco: same here

backend/substrapp/tests/views/tests_views_compositealgo.py
jmorel: We never use filters in our codebase but rather list comprehensions such as this one. I stuck with the current pattern to be consistent (this new test is adapted from existing ones). And from what I'm reading, the functions map/filter/reduce are at the center of quite a hot topic in the python community :)

GuillaumeCisco: `filter` is better in a pythonic way : explicit before implicit.
Not using it in our current codebase is not a sign of good health.

jmorel: And yet Guido Van Rossum wanted it out of the language. All I'm saying is that consistency trumps other arguments here and that this small PR isn't the best place to have this discussion.

samlesu: Both syntaxes are valid and could be used. 
Personally,  I'm using more often the list comprehension as I find it more readable and easier to maintain.

Kelvin-M:
Finally we can merge this PR ?

Kelvin-M:
@GuillaumeCisco Good for you ? Consistency should be used for PR that fix code issue.
We can update the list comprehension style to function style in a refacto PR if we think it's a better behavior

GuillaumeCisco:
Yes let's do another PR for fixing this.

PR #123, opened 2020-01-24
Update dependencies

GuillaumeCisco:


Kelvin-M:
Nice work, I need to test it !

backend/requirements.txt
```diff
@@ -1,34 +1,31 @@
...
+pycryptodomex>=3.9.4
+pyOpenSSL==19.1.0
+raven==6.10.0
+requests==2.22.0
+sentry-sdk==0.14.1
+six>=1.14.0
+tldextract==2.2.2
+uwsgi==2.0.18
 zxcvbn==4.4.28
 git+git://github.com/hyperledger/fabric-sdk-py.git@36cc15021f74c11c7ae3196e380a5275c220145f # fabric-sdk-py==0.8.1
```
inalgnu: Since fabric-sdk-py 0.8.1 was released, we could use a release tag here?

```
fabric-sdk-py==0.8.1

Kelvin-M: I remembered that there was an issue on the 0.8.1, we should test it :)

Kelvin-M: I've checked, the commit '36cc15...' is later than 0.8.1
I'm currently testing the last commit of fabric-sdk-py

Kelvin-M:
@GuillaumeCisco  I've updated the fabric sdk py (tested on docker and skaffold)

Kelvin-M:
@GuillaumeCisco Good for me, feel free to merge if you do not have more dependencies to update. See my update on fabric version too :)

GuillaumeCisco:
@inalgnu is this update ok for you?
If yes, do not hesitate to merge it.

inalgnu:
I'm still wondering why we are pointing a commit instead of using a tag for fabric-sdk-py (If I remember well we did that because the feature was not released yet). It seems not to be the case anymore? do you confirm @GuillaumeCisco?

Kelvin-M:
> I'm still wondering why we are pointing a commit instead of using a tag for fabric-sdk-py (If I remember well we did that because the feature was not released yet). It seems not to be the case anymore? do you confirm @GuillaumeCisco?

Because, fabric-sdk-py 0.8.1 is broken, so we need a commit later than the tag because @GuillaumeCisco  have fixed an issue introduced in this release. As there is no new release since we have to use a commit number for the version !

GuillaumeCisco:
Baohua still did not give his agreement for releasing a 8.2 version.
That's why we use last commit on master.

inalgnu:
That makes sense. Thank's

PR #125, opened 2020-01-27
add query/invoke strategy settings to give flexibilty

inalgnu:


backend/substrapp/ledger_utils.py
```diff
@@ -154,6 +154,23 @@ def get_hfc():
...
 
+def get_invoke_signing_peers(current_peer, all_peers):
+    return _get_signing_peers(strategy=settings.LEDGER_INVOKE_STRATEGY, current_peer=current_peer, all_peers=all_peers)
+
+
+def get_query_singing_peers(current_peer, all_peers):
+    return _get_signing_peers(strategy=settings.LEDGER_QUERY_STRATEGY, current_peer=current_peer, all_peers=all_peers)
+
+
+def _get_signing_peers(strategy, current_peer, all_peers):
```
samlesu: Minor, it looks like we are defining low to high level methods from top to bottom.
So, here `_get_signing_peers()` could be defined before `get_invoke_signing_peers` / `get_query_singing_peers`

inalgnu: done 👍

Kelvin-M: I think it's better to talk about `endorsing` peers than `signing` peers :

> Endorsing peer is the peer that will "endorse"/provide the seal of approval to a transaction when it is proposed. After the transaction is endorsed, the transaction(plus the endorsement) will be submitted to blockchain.

backend/substrapp/ledger_utils.py
```diff
@@ -154,6 +154,23 @@ def get_hfc():
         loop.close()
 
 
+def _get_signing_peers(strategy, current_peer, all_peers):
+    if strategy == 'SELF':
+        return [current_peer]
+    if strategy == 'MAJORITY':
```
Kelvin-M: This looks more like `ALL `  than `MAJORITY`.
All Majority combinations will be constructed by   
```python
from itertools import combinations 
all_comb = combinations(all_peers, len(all_peers)//2 + 1) 

```

backend/substrapp/ledger_utils.py
inalgnu: done here 9e65ccc

backend/substrapp/ledger_utils.py
inalgnu: Thank's @Kelvin-M, fixed here 3a9d353

PR #126, opened 2020-01-27
Only mount pred volume for testtuples

jmorel:
Since there are no more predictions for traintuples and composite_traintuples, we should only mount the pred volume in case of testtuple. 

I tested this change against https://github.com/SubstraFoundation/substra-tools/pull/24

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -615,9 +615,11 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id, 
 
     volumes = {
         data_path: {'bind': '/sandbox/data', 'mode': 'ro'},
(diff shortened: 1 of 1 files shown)


jmorel:
Quick afterthought: isn't this going to break the usage of substra-tools images older than 0.5.0 ? The algo will try to save predictions to a volume that doesn't exists.

samlesu:
I'm not 100% sure, substra-tools might create the pred folder if it doesn't exist. We should test it.

Even if we don't have to be backward compatible, dropping support for substra-tools 0.4.0 image could be painful.

jmorel:
I ran the substra-tests master against this branch of the backend and all test passes, you were right!

Kelvin-M:
Feel free to merge if there is no more modification to make @jmorel

PR #131, opened 2020-01-28
[WIP] Ledger: get rid of only_pkhash option

samlesu:
It seems that this code is not needed anymore. 
It looks like old legacy code left after a refactoring.

### DIFF
--- a/backend/substrapp/ledger.py
+++ b/backend/substrapp/ledger.py
@@ -113,7 +113,7 @@ def create_compositetraintuple(args):
 
 
 def create_computeplan(args):
-    return _create_asset('createComputePlan', args, only_pkhash=False)
+    return _create_asset('createComputePlan', args)
 
 
 def create_algo(args, key):
--- a/backend/substrapp/ledger_utils.py
+++ b/backend/substrapp/ledger_utils.py
@@ -233,7 +233,7 @@ def call_ledger(call_type, fcn, args=None, kwargs=None):
         return response
 
 
-def _invoke_ledger(fcn, args=None, cc_pattern=None, sync=False, only_pkhash=True):
+def _invoke_ledger(fcn, args=None, cc_pattern=None, sync=False):
     params = {
         'wait_for_event': sync,
     }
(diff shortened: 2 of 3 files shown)


samlesu:
Do not merge, it looks like it's breaking the e2e tests response parsing (missing validated field for data sample).

Kelvin-M:
@samlesu Can you provide more information (like results of the invoke/query response of the chaincode) between the previous and the current behavior. It may result of a change in the chaincode if this option is not necessary anymore.

samlesu:
> @samlesu Can you provide more information (like results of the invoke/query response of the chaincode) between the previous and the current behavior. It may result of a change in the chaincode if this option is not necessary anymore.

Thanks for your comment. To be honest, I don't understand the goal of this argument `only_pkhash`: change the chaincode response to return a dict with a single `pkhash` field. At first sight it seems uneeded but I may be wrong.

Do you have an idea of the purpose of this option and its role?

Kelvin-M:
I think it was at the past, when the sdk/cli needed to fetch the key of the tuple which was not expose directly in the chaincode response of an invoke. But it may have been harmonized now, dunno !
Maybe fixing the e2e test will show you where it is used :)

samlesu:
> I think it was at the past, when the sdk/cli needed to fetch the key of the tuple which was not expose directly in the chaincode response of an invoke. But it may have been harmonized now, dunno !
Maybe fixing the e2e test will show you where it is used :)

Thanks for the history. As you said, I need to investigate more and fix the tests.

Kelvin-M:
@samlesu You can ask @thibaultrobert because this introduction seems to be introduced here https://github.com/SubstraFoundation/substra-backend/commit/a07a6b17ca122dc6887f18a7fe01f6598617946c
in this context : https://github.com/SubstraFoundation/substra-backend/commits/master?before=fd2dce346ba4b0ceb5aa453b2cff46c0e31041c0+595

samlesu:
The tests were failing because the backend is adding a field "validated" to the create data sample response. This is done only if the chaincode responds with a "pkhash" field (that's not the case anymore with this PR). The e2e tests were failing because the field "validated" is missing.

It looks like to do a clean we should also remove the field validated as it seems deprecated with the removal of the dry-run option.

As it requires more time to do this properly, I'm closing this PR.

PR #132, opened 2020-01-28
Fix connection leaks

AurelienGasser:
This fixes connection leaks from the backend to the peers and the orderer.

# Before / after

If we look at the number of open TCP connections on `org-1-peer-1-backend-server` after running `substra-tests/make test`. Without the fix:

```
     90 network-orderer.or:7050
    720 network-org-1-peer:7051
     91 network-org-2-peer:7051
```

(Note that these connections don't close themselves after a while - I waited up to 30 min)

With the fix:

```
      1 network-org-1-peer:7051
```

# Impact on memory leaks

It looks like this fix solves _some_ of the memory leaks (but not all). Memory increase in `minikube` after running `make test`:

- Without fix: 400 MB
- With fix: 80 MB

(it's kind of hard to measure, so those numbers are just indicative)

# Possible next steps

## Close in parallel

We could close the channels in instead of sequentially, it should be a tad faster.

## PR to the python HLF SDK

We could consider adding the function `close_grpc_channels` to the `Client` class of the python HLF SDK, in a PR. What do you think? cc @GuillaumeCisco

## Reuse connections

(Note: this suggestion is out of the scope of this PR, I put it here for the sake of starting a discussion)

Currently, a new ledger client - along with the associated gRPC channels - is created every time we do an "invoke". We could consider re-using clients accross calls (and hence re-using gRPC channels). 

For instance, we could have only one client for each thread, that would be re-instanciated :

- If an error occurs during an "invoke" or a "query"
- Every N seconds

Since instantiating a client triggers adiscovery, this would ensure that we run discovery if we hit an invoke error, or every N seconds, whichever comes first.

### DIFF
--- a/backend/substrapp/ledger_utils.py
+++ b/backend/substrapp/ledger_utils.py
@@ -144,12 +144,22 @@ def retry_on_error(delay=1, nbtries=5, backoff=2, exceptions=None):
     return _retry
 
 
+async def close_grpc_channels(client):
+    for name in client.peers:
+        await client.peers[name]._channel.close()
+    for name in client.orderers:
+        await client.orderers[name]._channel.close()
+
+
 @contextlib.contextmanager
 def get_hfc():
     loop, client = LEDGER['hfc']()
     try:
         yield (loop, client)
     finally:
+        loop.run_until_complete(
+            close_grpc_channels(client)
+        )
         del client
         loop.close()
 


GuillaumeCisco:
Great job!
The fabric-sdk-py should be totally rewritten from scratch to be updated to the last specifications.

`fabric-sdk-py` could integrate a `close` method for closing the grpc channels, but it is up to the backend to handle these connections and to call the close method.

PR #133, opened 2020-01-29
Fix: dataset can be linked with only one objective

samlesu:
The backend was in the middle of nowhere to set the objective key of a dataset:
- accepting  'objective_keys' from the request and sending this field to the ledger serializer
- ledger serializer looking for an 'objective_key' field

As a dataset can be linked with only one objective, the correct field is 'objective_key'.

Companion PRs:
- [substra](https://github.com/SubstraFoundation/substra/pull/94)
- [tests](https://github.com/SubstraFoundation/substra-tests/pull/66)

### DIFF
--- a/backend/substrapp/management/commands/createdataset.py
+++ b/backend/substrapp/management/commands/createdataset.py
@@ -21,11 +21,11 @@ def path_leaf(path):
 class Command(BaseCommand):
     help = '''  # noqa
     create dataset
-    python ./manage.py createdataset '{"data_manager": {"name": "foo", "data_opener": "./opener.py", "description": "./description.md", "type": "foo", "objective_key": "foo", "permissions": {"public": True, "authorized_ids": []}}, "data_samples": {"paths": ["./data.zip", "./train/data"], "test_only": false}}'
+    python ./manage.py createdataset '{"data_manager": {"name": "foo", "data_opener": "./opener.py", "description": "./description.md", "type": "foo", "objective_key": "", "permissions": {"public": True, "authorized_ids": []}}, "data_samples": {"paths": ["./data.zip", "./train/data"], "test_only": false}}'
     python ./manage.py createdataset dataset.json
     # datamanager.json:
     # objective_key is optional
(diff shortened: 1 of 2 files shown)


samlesu:
> The createdataset command still uses keys instead of key. Could you check it out?

I don't think we should maintain these commands.

Kelvin-M:
@samlesu As the command still exists, we should maintain it until its removal, if one is programmed :)

samlesu:
> @samlesu As the command still exists, we should maintain it until its removal, if one is programmed :)

It has been updated to avoid a debate.

backend/substrapp/management/commands/createdataset.py
```diff
@@ -21,11 +21,11 @@ def path_leaf(path):
...
+    python ./manage.py createdataset '{"data_manager": {"name": "foo", "data_opener": "./opener.py", "description": "./description.md", "type": "foo", "objective_key": "foo", "permissions": {"public": True, "authorized_ids": []}}, "data_samples": {"paths": ["./data.zip", "./train/data"], "test_only": false}}'
     python ./manage.py createdataset dataset.json
     # datamanager.json:
-    # objective_keys are optional
-    # {"data_manager": {"name": "foo", "data_opener": "./opener.py", "description": "./description.md", "type": "foo", "objective_keys": [], "permissions": {"public": True, "authorized_ids": []}}, "data_samples": {"paths": ["./data.zip", "./train/data"], "test_only": false}}
+    # objective_key is optional
+    # {"data_manager": {"name": "foo", "data_opener": "./opener.py", "description": "./description.md", "type": "foo", "objective_key": "foo", "permissions": {"public": True, "authorized_ids": []}}, "data_samples": {"paths": ["./data.zip", "./train/data"], "test_only": false}}
```
Kelvin-M: `"foo"` seems to not be compatible with the smart contract. An empty or 64 characters string is better.
To keep the consistency, we can just put `""`

samlesu: Thanks.

PR #134, opened 2020-01-29
Fix description route

jmorel:
So it appears `description` may be a semi-reserved keyword in DRF.

Trying to access any "description" route made the server break without even entering the view. Everything in the stacktrace was internal to django and DRF.


jmorel:
All e2e tests are passing with this fix.

PR #138, opened 2020-02-03
Implement transfert bucket

thibaultrobert:


backend/substrapp/tasks/tasks.py
```diff
@@ -36,6 +39,8 @@
 COMPOSITE_TRAINTUPLE_TYPE = 'compositeTraintuple'
 TESTTUPLE_TYPE = 'testtuple'
 
+TAG_VALUE_FOR_TRANSFERT_BUCKET = "sendToTransfertBucket"
```
Kelvin-M: This tag value is maybe to long and error prone with this typo, no ?
I fell like something simpler could be enough like `bucket` :)

backend/substrapp/tasks/tasks.py
```diff
@@ -797,9 +802,38 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+        except ClientError as e:
+            logging.error(e)
+
     return result
 
 
+def transfer_to_bucket(tuple_key, pred_path, model_path):
+    access_key = os.getenv('BUCKET_TRANSFERT_ID')
+    secret_key = os.getenv('BUCKET_TRANSFERT_SECRET')
+    bucket_name = os.getenv('BUCKET_TRANSFERT_NAME')
```
Kelvin-M: We can set those values as global variable, in settings or at the beginning of the file to avoid reading them multiple times :)

backend/substrapp/tasks/tasks.py
```diff
@@ -797,9 +802,38 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+    if not access_key or not secret_key or not bucket_name:
+        return
```
Kelvin-M: We can add a log for this to understand what happened :)

backend/substrapp/tasks/tasks.py
```diff
@@ -797,9 +802,38 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+
+    with tarfile.open(tar_path, 'w:gz') as tar:
+        tar.add(pred_path)
+        tar.add(model_path)
+    s3 = boto3.client(
+        's3',
+        aws_access_key_id=access_key,
+        aws_secret_access_key=secret_key)
+    s3.upload_file(tar_path, bucket_name, tar_file)
+    os.remove(tar_path)
```
Kelvin-M: You should do it in a try/except/finally to be sur to remove the file if tar construction or boto raise an exception

backend/substrapp/tasks/tasks.py
```diff
@@ -797,9 +802,38 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+    if tag and TAG_VALUE_FOR_TRANSFERT_BUCKET in tag:
+        try:
+            transfer_to_bucket(subtuple["key"], pred_path, model_path)
+        except ClientError as e:
+            logging.error(e)
+
     return result
 
 
+def transfer_to_bucket(tuple_key, pred_path, model_path):
```
ClementGautier: This file is almost 1k LoC. We should maybe extract this function in a kind of `exporter` class. I would also have a list of file as argument so it can be used to export predictions without the associated model for example. The signature might be `def export(name: str, paths: List[]) -> None:` and the class a S3Exporter in an exporters/ package or something similar.

backend/substrapp/tasks/tasks.py
thibaultrobert: Done

backend/substrapp/tasks/tasks.py
thibaultrobert: Done

backend/substrapp/tasks/tasks.py
thibaultrobert: Indeed, this implementation was a bit optimistic. Fixed

backend/substrapp/tasks/tasks.py
thibaultrobert: This is indeed a long file and the backend could use a bit of abstraction regarding inputs and outputs. But I believe this exceed this PR scope.
I will take the paths' list advise though. Thank you for the feedback.

backend/substrapp/tasks/tasks.py
thibaultrobert: half-done -> `transfertBucket`. Is it ok for you?

thibaultrobert:
Tested with a real upload to S3. Works like charm. Thank you @ClementGautier for the test bucket

backend/substrapp/tasks/tasks.py
```diff
@@ -797,9 +805,38 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+    if not ACCESS_KEY or not SECRET_KEY or not BUCKET_NAME:
+        logger.info(f'unset global env for bucket transter: {ACCESS_KEY} {SECRET_KEY} {BUCKET_NAME}')
+        return
+
+    tar_file = f'{tuple_key}.tar.gz'
+    tar_path = path.join('/tmp', tar_file)
+
+    try:
+        with tarfile.open(tar_path, 'x:gz') as tar:
+            for file_path in paths:
```
Kelvin-M: `dir_path` is more appropriate for this loop

backend/substrapp/tasks/tasks.py
```diff
@@ -797,9 +805,38 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+                tar.add(file_path)
+        s3 = boto3.client(
+            's3',
+            aws_access_key_id=ACCESS_KEY,
+            aws_secret_access_key=SECRET_KEY)
+        s3.upload_file(tar_path, BUCKET_NAME, tar_file)
```
Kelvin-M: I think it' better to separate the tar creation and the upload. It may be useful in the futur to handle errors with retry =)

backend/substrapp/tasks/tasks.py
```diff
@@ -797,9 +805,38 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
     with open(path.join(pred_path, 'perf.json'), 'r') as perf_file:
         perf = json.load(perf_file)
     result['global_perf'] = perf['all']
+
+    tag = subtuple.get("tag")
+    if tag and TAG_VALUE_FOR_TRANSFERT_BUCKET in tag:
```
samlesu: as tag is a string, it should not be `TAG_VALUE_FOR_TRANSFERT_BUCKET == tag` or the `in` is done on purpose?

backend/substrapp/tasks/tasks.py
```diff
@@ -797,9 +805,38 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+        transfer_to_bucket(subtuple["key"], [pred_path, model_path])
+
     return result
 
 
+def transfer_to_bucket(tuple_key, paths):
+    if not ACCESS_KEY or not SECRET_KEY or not BUCKET_NAME:
+        logger.info(f'unset global env for bucket transter: {ACCESS_KEY} {SECRET_KEY} {BUCKET_NAME}')
```
samlesu: minor: could be a `logger.warning` or `logger.error` instead ?

backend/substrapp/tasks/tasks.py
```diff
@@ -797,9 +805,38 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+        return
+
+    tar_file = f'{tuple_key}.tar.gz'
+    tar_path = path.join('/tmp', tar_file)
```
samlesu: Instead of creating and handling a temporary file manually you could rely on the tempfile library.
https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

For instance (and to be tested):

```python
with tempfile.TemporaryDirectory(prefix='/tmp/') as tmpdir:
    tar_path = path.join(tmpdir, f'{tuple_key}.tar.gz')
    try:
        ...
    except:
        ...
```

This way you don't have to deal with the deletion of the temporary folder in case of success / failure.

backend/substrapp/tasks/tasks.py
thibaultrobert: Done

backend/substrapp/tasks/tasks.py
thibaultrobert: I've just separate it in two different try/except for now.

backend/substrapp/tasks/tasks.py
thibaultrobert: Yes, it's on purpose. We want to match a potential sub-string here.

backend/substrapp/tasks/tasks.py
thibaultrobert: Yes and done

backend/substrapp/tasks/tasks.py
thibaultrobert: Very good point

thibaultrobert:
> If I'm not mistaken, "transfer" in english doesn't end with the letter "t".

Yes you are right. Fixed.

I've update the PR and re-test an upload to a real bucket S3 with success.

backend/substrapp/tasks/tasks.py
```diff
@@ -797,9 +805,40 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+
+    with tempfile.TemporaryDirectory(prefix='/tmp/') as tmpdir:
+        tar_path = path.join(tmpdir, f'{tuple_key}.tar.gz')
+        try:
+            with tarfile.open(tar_path, 'x:gz') as tar:
+                for dir_path in paths:
+                    tar.add(dir_path)
+        except tarfile.TarError as e:
+            logger.warning(e)
```
samlesu: you should return here as in case of tar errors, you don't want to continue the transfer.

backend/substrapp/tasks/tasks.py
samlesu: 👍

backend/substrapp/tasks/tasks.py
```diff
@@ -797,9 +805,43 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
     return result
 
 
+def safe(fn):
+    @functools.wraps(fn)
+    def wrapper(*args, **kwargs):
+        try:
+            return fn(*args, **kwargs)
+        except Exception as e:
+            logging.exception(e)
```
Kelvin-M: You should use the instantiated `logger` instead of logging :)

backend/substrapp/tasks/tasks.py
```diff
@@ -797,9 +805,43 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
     result['global_perf'] = perf['all']
+
+    tag = subtuple.get("tag")
+    if tag and TAG_VALUE_FOR_TRANSFER_BUCKET in tag:
+        transfer_to_bucket(subtuple["key"], [pred_path, model_path])
+
     return result
 
 
+def safe(fn):
```
Kelvin-M: `safe` seems to be a tooling function, it should not be in tasks.py directly

PR #141, opened 2020-02-04
Secrets plumbing fixes

samlesu:
- use compute plan tag to link secrets and compute plan (tuple tag was used)
- raise error if no secret match tag
- fix typo in label selector

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -663,7 +663,7 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id, 
             cp_metadata = get_object_from_ledger(compute_plan_id, 'queryComputePlan')
             cp_tag = cp_metadata['tag']
             secret_namespace = os.getenv('K8S_SECRET_NAMESPACE', 'default')
-            label_selector = f"compute_plan=f{cp_tag}"
+            label_selector = f"compute_plan={cp_tag}"
 
             k8s_client = get_k8s_client()
             try:
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -591,10 +591,13 @@ def do_task(subtuple, tuple_type):
     # compute plan / federated learning variables
     compute_plan_id = None
     rank = None
+    compute_plan_tag = None
 
     if 'computePlanID' in subtuple and subtuple['computePlanID']:
         compute_plan_id = subtuple['computePlanID']
         rank = int(subtuple['rank'])
+        compute_plan = get_object_from_ledger(compute_plan_id, 'queryComputePlan')
+        compute_plan_tag = compute_plan['tag']
 
     client = docker.from_env()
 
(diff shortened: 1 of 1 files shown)


samlesu:
@Kelvin-M could you review the fixes please?

backend/substrapp/tasks/tasks.py
```diff
@@ -640,8 +640,10 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
         chainkeys_directory = get_chainkeys_directory(compute_plan_id)
 
         if not os.path.exists(chainkeys_directory):
+            cp_metadata = get_object_from_ledger(compute_plan_id, 'queryComputePlan')
+            cp_tag = cp_metadata['tag']
             secret_namespace = os.getenv('K8S_SECRET_NAMESPACE', 'default')
-            label_selector = f"compute_plan={subtuple.get('tag')}"
+            label_selector = f"compute_plan={cp_tag}"
```
Kelvin-M: The tag is not associated to the tuple ?
If it's on the compute this part should be moved outside this function which should not do a ledger call

samlesu:
> How can I test it :) ?

Will be answered on slack.

backend/substrapp/tasks/tasks.py
samlesu: The tag is associated with the compute plan (cf PR description).

Why this function should not do a ledger call? 

I can move the ledger call to the upper function but I'm afraid it will make the code harder to read: this call needs to be done only on specific cases: 
- secrets mode enabled
- and, tuple with compute plan
- and, secrets directory not existing

Kelvin-M: The assumption is that all metadata needed in the `do_task` are already provided as it's a compute task, it may have no connection to any services at all.
It's better to fetch all metadata before, where we could potential do ledger call.

You can just pass the `cp_tag` no matter what and let the strategy in the `do_task`, only the ledger part should be done outside

backend/substrapp/tasks/tasks.py
```diff
@@ -661,7 +665,7 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
 
         if not os.path.exists(chainkeys_directory):
             secret_namespace = os.getenv('K8S_SECRET_NAMESPACE', 'default')
-            label_selector = f"compute_plan={subtuple.get('tag')}"
+            label_selector = f"compute_plan={compute_plan_tag}"
```
Kelvin-M:
```python
label_selector = f'compute_plan={compute_plan_tag}'
```

samlesu:
```suggestion
            label_selector = f'compute_plan={compute_plan_tag}'
```

backend/substrapp/tasks/tasks.py
```diff
@@ -670,12 +674,17 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
                 logger.error(f'failed to fetch namespaced secrets {secret_namespace} with selector {label_selector}')
                 raise e
 
-            secrets = {s['metadata']['name']: int.from_bytes(b64decode(s['data']['key']), 'big')
-                       for s in secrets.to_dict()['items']}
+            secrets = {
+                s['metadata']['labels']['index']: list(b64decode(s['data']['key']))
+                for s in secrets.to_dict()['items']
+            }
```
Kelvin-M: Can you explain the type change here  (from `int` to `list`) ?

samlesu: Answered on Slack.

PR #149, opened 2020-02-06
(WIP) Reuse HLF client

AurelienGasser:
I would like to try to add unit tests and do some cleanup, but I would be interested in early feedback ! :)

---

## TL;DR

- Have 1 HLF client per thread
- Renew the client - and redo discovery - every N seconds (I put 30 seconds arbitrarily)
- Renew the client every time an invoke/query fails

Note that currently (before the PR), the HLF client is essentially renewed for every invoke/query.

## Performance impact

On my local it speeds up `test_aggregate_composite_traintuples` by about 8%.

10 attempts on each branch, sorted by duration:

image

### DIFF
--- a/backend/substrapp/ledger_client.py
+++ b/backend/substrapp/ledger_client.py
@@ -1,0 +1,79 @@
+import datetime
+import logging
+import threading
+
+from django.conf import settings
+
+LEDGER = getattr(settings, 'LEDGER', None)
+logger = logging.getLogger(__name__)
+
+LEDGER_CLIENT_TTL_SECONDS = 30
+
+_client_ttl_delta = datetime.timedelta(0, LEDGER_CLIENT_TTL_SECONDS)
+_threadLocal = threading.local()
+
+
+class ClientWrapper:
+    """
+    Validate whether the password is of a maximum length.
+    """
+    _loop = None
+    _client = None
+    _expires_on = None
+    _destroyed = False
+
+    def __init__(self, now):
+        (loop, client) = LEDGER['hfc']()
+        self._loop = loop
+        self._client = client
+        self._expires_on = now + _client_ttl_delta
+        logger.info("XXXXX Initializing")
+
+    def has_expired(self, now):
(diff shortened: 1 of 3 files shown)


AurelienGasser:
@Kelvin-M @samlesu @GuillaumeCisco @inalgnu I'd be interested in early feedback, including coding style.

backend/substrapp/ledger_client.py
```diff
@@ -0,0 +1,79 @@
...
+
+LEDGER_CLIENT_TTL_SECONDS = 30
+
+_client_ttl_delta = datetime.timedelta(0, LEDGER_CLIENT_TTL_SECONDS)
+_threadLocal = threading.local()
+
+
+class ClientWrapper:
+    """
+    Validate whether the password is of a maximum length.
```
AurelienGasser: copy pasting is bad

Kelvin-M:
Nice work !
As you notice, the idea that we can centralise hyperledger calls into a share object (or a service) can be nice. I'd rather prefer to have on service that deals with hyperledger communications instead of putting a share object into the local thread which may be problematic as in event app we launch the call in a process, and we heavily use asyncio in fabric-sdk-py. I'm not aware if there is any risk to do that.
And I'm not talking about debugging that ^^ ! The fact that we use one instance per call is maybe not optimised but we don't deal with potential concurrency/share object issue in a replicated setup (as we are with uwsgi).

AurelienGasser:
Following discussion with @Kelvin-M: sharing a client instance within a thread could bring potential asyncio concurrency issues. There's no evidence of such issues, but a lot of testing would be needed to validate this PR!

A better approach would be to implement a HLF client "pool". The pool would be a single service (i.e. one service per process) which holds a fixed number of HLF clients. A caller can request a HLF client, and that client is _locked_ (unusable by other callers) until the caller releases it. This way, there's no concurrency issue: each HLF client is never used by more than one caller at a time. Work is needed to implement this solution as part of another PR.

PR #154, opened 2020-02-10
Test pod

AurelienGasser:
Run load tests from the cluster directly, as opposed to running them from a developer's computer.

Advantages:
- Share logs easily (anyone can SSH into the pod)
- Load test runs survive developer machine crashes
- substra CLI is already configured

## Run CLI commands and tests

```bash
# SSH into the pod
$ kubectl exec -it -n org-1 backend-org-1-substra-backend-test-pod-56fc9c44d9-8bsnd bash
Welcome! 😬
Type "substra" or "cd substra-tests" to begin.

# CLI is already configured
(.venv) /pod$ substra list traintuple
KEY                                                                 ALGO NAME                                           STATUS      RANK    TAG     COMPUTE PLAN ID     
c44babc1955d2ac0c3fab9c12c6d012feaff4b903013e83e5d49882da5662404   

# substra-test is available
(.venv) /pod$ cd substra-tests
(.venv) /pod/substra-tests$ make test
... hopefully the tests pass
```

## Use a private `substra-tests` implementation

in `skaffold.yaml`:

```yaml
overrides:
    testPod:
        enabled: true
        substraTests:
            directory: substra-tests
            tarball: https://github.com/<user>/<repo>/archive/<branch>.tar.gz
            githubToken: 75dceadc88107d907de6c24b0cfa66346xxxxxxxx
```

Instructions to create a private access token: [here](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line)

### DIFF
--- a/charts/substra-backend/templates/deployment-test-pod.yaml
+++ b/charts/substra-backend/templates/deployment-test-pod.yaml
@@ -81,7 +81,7 @@ spec:
             b64="H4sIAAAAAAAAA1NQoBaISY22NMu10gHSBrlcVDZWX0FBD2rwodVxSkqPGveXg2QfTZ0CVZcO5cYBAVC2DiKKohgokwWSBokvRFWHsAZkyFSIVVkIVzyaOgvKKkLhHm4FkUgKwZLzYKwpcyCMWHQ7kGxRSDi0qe5RzwQdHBbVAcGjqRP8IPIoitEt0wGBujiEXbE6UPCoswM5XhKAtoM1z4Bq9kW1+1HHSrh3oGqxeUxXh0upDmT8oVVgYaVyIFv30RSIqYdW1YHd3jMB7CIUlVhDSZULAEt2AdOOAgAA"
             banner="$(echo $b64 | base64 -d | gunzip)"
             echo -e "$banner"
-            echo "Welcome! 😬" # TODO: add giant substra ASCII-art banner :p
+            echo "Welcome! 😬"
             echo "Type \"substra\" or \"cd substra-tests\" to begin."
             EOF
 
--- a/charts/substra-backend/templates/deployment-test-pod.yaml
+++ b/charts/substra-backend/templates/deployment-test-pod.yaml
@@ -1,0 +1,133 @@
+{{- if .Values.testPod.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "substra.fullname" . }}-test-pod
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    app.kubernetes.io/name: {{ template "substra.name" . }}-test-pod
+    app.kubernetes.io/part-of: {{ template "substra.name" . }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+        app.kubernetes.io/name: {{ template "substra.name" . }}-test-pod
+        app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: {{ template "substra.name" . }}-test-pod
+        app.kubernetes.io/instance: {{ .Release.Name }}
+    spec:
+      {{- with $.Values.testPod.image.pullSecrets }}
+      imagePullSecrets:
+      {{- range . }}
+        - name: {{ . }}
+      {{- end }}
+      {{- end }}
+      initContainers:
+      - name: init
+        image: "{{ .Values.testPod.image.repository }}:{{ .Values.testPod.image.tag }}"
+        workingDir: /pod
+        command: ["/bin/bash", "-c"]
+        args:
+          - |
+            # Echo commands, abort on first error
+            set -xe
+
+            # Use venv because every folder except /pod will be wiped after the init container dies
+            pip install virtualenv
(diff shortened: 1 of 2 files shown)


charts/substra-backend/templates/deployment-test-pod.yaml
```diff
@@ -0,0 +1,98 @@
...
+    matchLabels:
+        app.kubernetes.io/name: {{ template "substra.name" . }}-test-pod
+        app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: {{ template "substra.name" . }}-test-pod
+        app.kubernetes.io/instance: {{ .Release.Name }}
+    spec:
+      {{- with $.Values.backend.image.pullSecrets }}
```
ClementGautier:
```suggestion
      {{- with $.Values.testPod.image.pullSecrets }}
```

charts/substra-backend/templates/deployment-test-pod.yaml
```diff
@@ -0,0 +1,98 @@
...
+      imagePullSecrets:
+      {{- range . }}
+        - name: {{ . }}
+      {{- end }}
+      {{- end }}
+      containers:
+      - name: test-pod
+        image: "{{ .Values.testPod.image.repository }}:{{ .Values.testPod.image.tag }}"
+        {{- if .Values.backend.image.pullPolicy }}
```
ClementGautier:
```suggestion
        {{- if .Values.testPod.image.pullPolicy }}
```

charts/substra-backend/templates/deployment-test-pod.yaml
```diff
@@ -0,0 +1,98 @@
...
+        imagePullPolicy: "{{ .Values.backend.image.pullPolicy }}"
```
ClementGautier:
```suggestion
        imagePullPolicy: "{{ .Values.testPod.image.pullPolicy }}"
```

charts/substra-backend/templates/deployment-test-pod.yaml
```diff
@@ -0,0 +1,98 @@
...
+        args:
+          - |
+            {{- with $.Values.testPod.extraRepos }}
+            {{- range . }}
+                git clone {{ .url }}
+                cd {{ .name }}
+                [ -f requirements.txt ] && pip install -r requirements.txt
+            {{- end }}
+            {{- end }}
+      {{- end }}
```
ClementGautier: You will loose those files if you don't use a volume shared between all the pods. You can use a simple emptyDir.

charts/substra-backend/templates/deployment-test-pod.yaml
```diff
@@ -0,0 +1,133 @@
...
+            add_user() {
+                substra config $3 --profile default --username $1 --password $2
+            }
+            source .venv/bin/activate
+            add_user $user_password $BACKEND_URL
+            substra login
+
+            # Zzz
+            while [ true ]; do
+                sleep 1000;
```
ClementGautier: why not sleep infinity here and skip the loop?

AurelienGasser:
@ClementGautier 

-  +1 Agreed! The CLI is non-essential for this PR, so [this whole block](https://github.com/SubstraFoundation/substra-backend/blob/ea9e60c8657f2c6d11d086ef2bd82de98fd30962/charts/substra-backend/templates/deployment-test-pod.yaml#L88-L95) can be safely removed. I only added the CLI as a "nice-to-have", but I now realize it was distracting from the main point of this yaml file (which is to provide a shared environment to run long-lasting tests from, see below). Note that installing the `substra` package is still required because `substra-tests` depends on it.
- My choice of words wasn't great: instead of "load tests", I really meant "long lasting tests". We're not trying to overload the sytem as much as we're trying to measure its behavior over long runs.
  - In this context, is it an issue to run the test pod in the same cluster as the system under test? My understanding is that k8s will not restart/shutdown the test-pod even if other pods in the same cluster misbehavae or use too many resources. Also, the test pod itself uses virtually no ressouces.
  - The metrics we want to measure are independent from the environment we run the tests from (developer machine v.s. test pod). For instance, we're measuring the time to download models from the backend. Do your comments still apply in that context?

To be clear (which isn't very clear from reading the description of this PR, my bad!), the point of this PR is to have a **shared environment to run long-lasting tests from**. Its value proposition:

- not rely on a developer machine staying on overnight to run long tests
- shared visibility on 
  - are the tests being run (`ps`)
  - which tests are being run (`ps`)
  - since when have the tests been started (`ps`)
  - anyone can read the logs (multiple devs can run `kubectl logs -F` concurrently) 

A dedicated machine or VM, independent from the system under test, would be ideal. In the absence of such machine/VM, wouldn't the test pod  proposed in this PR be a step forward?

> I don't want to be the man that question everything [...]

Thank you very much for your review! By all means, please keep on proving me wrong :)

ClementGautier:
I still think this shouldn't be part of the application package. As I understand the goal here is to ease the testing from a central place. As I told you benchmarking by adding workload on the same system than your application is not a good idea so you should be able to deploy the test runner on another machine. 

To solve that I suggest you package the substra-test scripts (at least Dockerfile + instructions to deploy on k8s or add a Chart if you think it ease the deployment). This way you could deployer the tester pod somewhere and use it on a CI tool or something automated on the long run.

The Dockerfile on substra-tests could start from the latest image substrafoundation/substra (assuming you create an image for substra cli) so the test pod would benefit from having a configured cli as well.

AurelienGasser:
Closing in favor of a Dockerfile in `substra-tests`

PR #155, opened 2020-02-12
Timing logs and optimizations

jmorel:
This PR adds timing logs to lots of worker methods in order to better see what is impacting the overall execution time.

It also adds the following optimizations:
- parallel download of inmodels
- only build docker image if it doesn't already exists

### DIFF
--- a/backend/substrapp/tasks/utils.py
+++ b/backend/substrapp/tasks/utils.py
@@ -225,12 +225,15 @@ def compute_docker(client, resources_manager, dockerfile_path, image_name, conta
 
     # Check if image already exist
     try:
+        ts = time.time()
         client.images.get(image_name)
     except docker.errors.ImageNotFound:
         logger.info(f'ImageNotFound: {image_name}. Building it')
     else:
         logger.info(f'ImageFound: {image_name}. Use it')
         build_image = False
+    finally:
+        logger.info(f'[TIME] client.images.get - {(time.time() - ts)} s')
 
     if build_image:
         try:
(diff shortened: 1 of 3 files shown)


backend/substrapp/tasks/utils.py
```diff
@@ -207,17 +225,36 @@ def compute_docker(client, resources_manager, dockerfile_path, image_name, conta
...
+        except docker.errors.BuildError as e:
+            # catch build errors and print them for easier debugging of failed build
+            lines = [line['stream'].strip() for line in e.build_log if 'stream' in line]
+            lines = [l for l in lines if l]
+            error = '\n'.join(lines)
+            logger.error(f'BuildError: {error}')
+            raise
+        finally:
+            logger.info(f'BuildSuccess - {image_name} - keep cache : {remove_image}')
+            logger.info(f'[TIME] client.images.build - {(time.time() - ts)} s')
```
samlesu: finally is called even in case of errors so having a log with `BuildSuccess` seems wrong. Could be moved to an else clause?

backend/substrapp/utils.py
```diff
@@ -186,3 +193,16 @@ def get_remote_file_content(url, auth, content_hash, salt=None):
     if computed_hash != content_hash:
         raise NodeError(f"url {url}: hash doesn't match {content_hash} vs {computed_hash}")
     return response.content
+
+
+def get_and_put_remote_file_content(url, auth, content_hash, salt=None, content_dst_path=None):
```
samlesu: The signature seems incorrect, when using stream=True, content_dst_path should be set. The following signature seems better: 
```python
def get_and_put_remote_file_content(url, auth, content_hash, content_dst_path, salt=None):
```

Same comment for the function `get_and_put_asset_content`.

PR #161, opened 2020-02-14
Update server values

Kelvin-M:
It may be better to have this in the values.yaml instead of default ! What do you think

### DIFF
--- a/charts/substra-backend/templates/deployment-server.yaml
+++ b/charts/substra-backend/templates/deployment-server.yaml
@@ -34,7 +34,7 @@ spec:
         {{- end }}
         command: ["/bin/bash"]
         {{- if eq .Values.backend.settings "prod" }}
-        args: ["-c", "uwsgi --http :{{ .Values.backend.service.port }} --module backend.wsgi --static-map /static=/usr/src/app/backend/statics --master --processes 4 --threads 2 --need-app --env DJANGO_SETTINGS_MODULE=backend.settings.server.prod "]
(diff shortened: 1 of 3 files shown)


charts/substra-backend/templates/deployment-server.yaml
```diff
@@ -34,7 +34,7 @@ spec:
         {{- end }}
         command: ["/bin/bash"]
         {{- if eq .Values.backend.settings "prod" }}
-        args: ["-c", "uwsgi --http :{{ .Values.backend.service.port }} --module backend.wsgi --static-map /static=/usr/src/app/backend/statics --master --processes 4 --threads 2 --need-app --env DJANGO_SETTINGS_MODULE=backend.settings.server.prod "]
+        args: ["-c", "uwsgi --http :{{ .Values.backend.service.port }} --module backend.wsgi --static-map /static=/usr/src/app/backend/statics --master --processes 8 --threads 2 --need-app --env DJANGO_SETTINGS_MODULE=backend.settings.server.prod "]
```
ClementGautier: Could be nice to make that a template value so we can tune it on each use case.

charts/substra-backend/templates/deployment-server.yaml
```diff
@@ -98,7 +98,7 @@ spec:
               - name: Accept
                 value: "text/html;version=0.0, */*;version=0.0"
           initialDelaySeconds: 60
-          timeoutSeconds: 5
+          timeoutSeconds: 10
```
ClementGautier: to be reverted?

Kelvin-M: Yes :)

Kelvin-M:
@samlesu @inalgnu  can you review :) ?

PR #162, opened 2020-02-14
No limit on grpc client

Kelvin-M:
Companion PR https://github.com/SubstraFoundation/hlf-k8s/pull/37

### DIFF
--- a/backend/backend/settings/deps/ledger.py
+++ b/backend/backend/settings/deps/ledger.py
@@ -141,8 +141,9 @@ def update_client_with_discovery(client, discovery_results):
                 peer.init_with_bundle({
                     'url': url,
                     'grpcOptions': {
-                        'grpc-max-send-message-length': 15,
-                        'grpc.ssl_target_name_override': peer_info['endpoint'].split(':')[0]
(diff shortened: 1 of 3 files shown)


backend/backend/settings/deps/ledger.py
```diff
@@ -142,7 +142,9 @@ def update_client_with_discovery(client, discovery_results):
                     'url': url,
                     'grpcOptions': {
                         'grpc-max-send-message-length': 15,
-                        'grpc.ssl_target_name_override': peer_info['endpoint'].split(':')[0]
+                        'grpc.ssl_target_name_override': peer_info['endpoint'].split(':')[0],
+                        'grpc.max_send_message_length': -1,
+                        'grpc.max_receive_message_length': -1
```
samlesu: Is it intended to have the keys `grpc-max-send-message-length` and  `grpc.max_send_message_length` in `grpcOptions`? 

It looks a bit confusing: it's the same name with different naming styles and different values (15 vs -1).

Kelvin-M: Yes nice catch, I would like to have a setup to test it but I need @thibaultrobert  to do that easily I think

PR #163, opened 2020-02-17
Add celery task option for more resiliency in the compute task function

Kelvin-M:
The celery worker process may be sigkill by the system.
This PR provides a way to retry the task because of that ! 

You can try it by launching a substra network and killing (in the worker) a celery process with `ps -awwwx | grep [c]elery | tail -n 1 | awk '{print $1}' | xargs kill -9` and see the logs to see that the task will comes back

http://docs.celeryproject.org/en/latest/userguide/configuration.html#task-reject-on-worker-lost
https://github.com/celery/celery/issues/5106


samlesu:
Tested

PR #165, opened 2020-02-19
fix delete secrets

samlesu:
To be tested

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -682,7 +682,7 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id, 
             k8s_client = get_k8s_client()
             try:
                 secrets = k8s_client.list_namespaced_secret(secret_namespace, label_selector=label_selector)
-            except kubernetes.client.ApiException as e:
+            except kubernetes.client.rest.ApiException as e:
                 logger.error(f'failed to fetch namespaced secrets {secret_namespace} with selector {label_selector}')
                 raise e
             secrets = secrets.to_dict()['items']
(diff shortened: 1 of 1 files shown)


backend/substrapp/tasks/tasks.py
```diff
@@ -678,31 +678,46 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
+                        body=kubernetes.client.V1Secret(
+                            data={},
+                            metadata=kubernetes.client.V1ObjectMeta(
+                                name=secret['metadata']['name'],
+                                labels=secret['metadata']['labels'],
+                            ),
+                        ),
+                    )
+                except kubernetes.client.ApiException as e:
                     logger.error(f'failed to delete secrets from namespace {secret_namespace}')
```
inalgnu: maybe we could change the error log message, failed to replace the secret?

backend/substrapp/tasks/tasks.py
```diff
@@ -678,31 +678,46 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
             secret_namespace = os.getenv('K8S_SECRET_NAMESPACE', 'default')
             label_selector = f'compute_plan={compute_plan_tag}'
 
+            # fetch secrets and write them to disk
             k8s_client = get_k8s_client()
             try:
                 secrets = k8s_client.list_namespaced_secret(secret_namespace, label_selector=label_selector)
-            except ApiException as e:
+            except kubernetes.client.ApiException as e:
```
Kelvin-M: `from kubernetes.client.rest import ApiException` to `kubernetes.client.ApiException` ?

samlesu: because the `kubernetes.client` package is also used to create the Secret structure (introduced in this PR).

backend/substrapp/tasks/tasks.py
samlesu: good catch

backend/substrapp/tasks/tasks.py
samlesu: Good catch! Will be updated.

PR #166, opened 2020-02-20
Implement compute plan events

Kelvin-M:
Handle new events structure. Add removal of local volume and algo ima…ges at the end of compute plan.

Companion PR https://github.com/SubstraFoundation/substra-chaincode/pull/98

### DIFF
--- a/backend/events/apps.py
+++ b/backend/events/apps.py
@@ -17,7 +17,7 @@ from hfc.fabric.peer import Peer
 from hfc.fabric.user import create_user
 from hfc.util.keyvaluestore import FileKeyValueStore
 
-from substrapp.tasks.tasks import prepare_tuple
+from substrapp.tasks.tasks import prepare_tuple, on_finished_compute_plan
 from substrapp.utils import get_owner
 from substrapp.ledger_utils import get_hfc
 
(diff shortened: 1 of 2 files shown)


backend/events/apps.py
```diff
@@ -40,53 +40,82 @@ def get_event_loop():
...
+def on_compute_plan(tx_status, _tuple):
+
+    worker_queue = f"{LEDGER['name']}.worker"
+
+    key = _tuple['computePlanID']
+    status = _tuple['status']
+
+    logger.info(f'Processing cleaning task {key}: type=computePlan status={status}')
+
+    clean_compute_plan.apply_async(
```
AurelienGasser: Do we want clean the compute plan on _any_ compute plan event? I would expect a condition on the `status` here.

backend/substrapp/tasks/tasks.py
```diff
@@ -905,3 +887,25 @@ def get_subtuple_directory(subtuple):
 def get_chainkeys_directory(compute_plan_id):
     return path.join(getattr(settings, 'MEDIA_ROOT'), getattr(settings, 'ORG_NAME'), 'computeplan',
                      compute_plan_id, 'chainkeys')
+
+
+def remove_algo_images(algo_hashes):
+    client = docker.from_env()
+    for algo_hash in algo_hashes:
+        algo_docker = f'substra/algo_{algo_hash[0:8]}'.lower()
```
AurelienGasser: What about extracting this line into a function that could also be called from `_do_task`?

backend/events/apps.py
```diff
@@ -40,53 +40,82 @@ def get_event_loop():
...
+
+def on_event(cc_event, block_number, tx_id, tx_status):
+    payload = json.loads(cc_event['payload'])
+
+    for event_type, _tuples in payload.items():
+
         if not _tuples:
             continue
 
         for _tuple in _tuples:
```
thibaultrobert: I think I would prefer not calling it `_tuples` and `_ tuple`, here but `_asset(s)` or something like that.

backend/events/apps.py
Kelvin-M: Currently, we only received the event on `done, `failed` and `canceled` and we apply the same behavior for those three status. Maybe in the futur we will do conditionnal strategy based on the status :)

backend/substrapp/tasks/tasks.py
Kelvin-M: You're talkin about line 895, to make it as a function because it's also used in `_do_task` ?

backend/events/apps.py
Kelvin-M: Yes

backend/substrapp/tasks/tasks.py

backend/events/apps.py
Kelvin-M: Done

backend/substrapp/tasks/tasks.py
```diff
@@ -633,12 +615,12 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
...
     opener_file = path.join(subtuple_directory, 'opener/opener.py')
     algo_path = path.join(subtuple_directory)
-    algo_docker = f'substra/algo_{algo_hash[0:8]}'.lower()  # tag must be lowercase for docker
+    algo_docker = get_algo_image_name(algo_hash)  # tag must be lowercase for docker
     algo_docker_name = f'{tuple_type}_{subtuple["key"][0:8]}'
     output_head_model_filename = 'head_model'
     output_trunk_model_filename = 'trunk_model'
 
-    remove_image = not((compute_plan_id is not None and rank != -1) or settings.TASK['CACHE_DOCKER_IMAGES'])
+    remove_image = not(compute_plan_id is not None or settings.TASK['CACHE_DOCKER_IMAGES'])
```
thibaultrobert: Not directly link with this PR but this `remove_image` value is now only use one time so we might want to set it either directly when calling `compute_docker` or just above it.

backend/events/apps.py
AurelienGasser: Ah, that makes sense ! What would you think of copying your comment above ^ into a comment in the code ? That would make it clearer IMO.

Kelvin-M: Yes

backend/events/apps.py
```diff
@@ -40,53 +40,82 @@ def get_event_loop():
         loop.close()
 
 
-def tuple_get_worker(tuple_type, _tuple):
-    if tuple_type == 'aggregatetuple':
-        return _tuple['worker']
-    return _tuple['dataset']['worker']
+def tuple_get_worker(event_type, _asset):
```
samlesu: minor: why adding a `_` prefix to asset (same everywhere)?

As far ass I know the `_` prefix on tuple is recommended because tuple is a python reserved keyword.

backend/events/apps.py
samlesu: Maybe we can rename the function to `on_compute_plan_completed` (or something like that) to avoid confusion? Or we could add a comment: "this event is triggered when ..."?

backend/events/apps.py
```diff
@@ -40,53 +40,82 @@ def get_event_loop():
...
-    for tuple_type, _tuples in payload.items():
-        if not _tuples:
+    key = _asset['key']
+    status = _asset['status']
+
+    if tx_status != 'VALID':
+        logger.error(
+            f'Failed transaction on task {key}: type={event_type}'
+            f' status={status} with tx status: {tx_status}')
+        return
```
samlesu: This should not be done for all events?

backend/events/apps.py
Kelvin-M: Done

backend/substrapp/tasks/tasks.py

backend/events/apps.py
samlesu: You have been too fast, my comment is deprecated :|

backend/events/apps.py
Kelvin-M: Yes, only do a replace shortcut :)

backend/events/apps.py
Kelvin-M: What do you mean ? This is the transaction status, if the transaction is not valid, it's not commited  in the ledger !

backend/substrapp/tasks/tasks.py
```diff
@@ -633,12 +615,11 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
     pred_path = path.join(subtuple_directory, 'pred')
     opener_file = path.join(subtuple_directory, 'opener/opener.py')
     algo_path = path.join(subtuple_directory)
-    algo_docker = f'substra/algo_{algo_hash[0:8]}'.lower()  # tag must be lowercase for docker
+    algo_docker = get_algo_image_name(algo_hash)  # tag must be lowercase for docker
```
AurelienGasser:
```suggestion
    algo_docker = get_algo_image_name(algo_hash)
```

backend/events/apps.py
```diff
@@ -40,53 +40,86 @@ def get_event_loop():
...
+
+def on_compute_plan(tx_status, _asset):
+
+    worker_queue = f"{LEDGER['name']}.worker"
+
+    key = _asset['computePlanID']
+
+    # Currently, we received this event on done, failed and canceled status
+    # We apply the same behavior for those three status.
+    # In the futur, we can do conditionnal strategy based on the status
```
AurelienGasser:
```suggestion
    # In the future, we can apply a conditional strategy based on the status.
```

backend/substrapp/tasks/tasks.py
```diff
@@ -898,10 +879,36 @@ def get_volume_id(compute_plan_id, prefix='local'):
     return f'{prefix}-{compute_plan_id}-{org_name}'
 
 
+def get_algo_image_name(algo_hash):
+    return f'substra/algo_{algo_hash[0:8]}'.lower()
```
AurelienGasser:
```suggestion
    return f'substra/algo_{algo_hash[0:8]}'.lower()  # tag must be lowercase for docker
```

backend/events/apps.py
Kelvin-M: Done

backend/substrapp/tasks/tasks.py

backend/substrapp/tasks/tasks.py

backend/events/apps.py
Kelvin-M: Done

backend/events/apps.py
samlesu: The transaction status is checked only when processing the tuples event. 

I'm wondering if it should be checked for the compute_plan event as well?

samlesu: It has already be done in the meantime.

PR #167, opened 2020-02-20
Computeplan remove chainkeys volume

inalgnu:
This is following local folder removal after receiving a chaincode event #166 extending it to the chain keys volume.

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -402,6 +402,16 @@ def remove_local_folder(compute_plan_id):
     except Exception:
         logger.error(f'Cannot remove local volume {volume_id}', exc_info=True)
 
+    if settings.TASK['CHAINKEYS_ENABLED']:
+        chainkeys_volume_id = get_chainkeys_directory(compute_plan_id)
+        try:
+            chainkeys_volume = client.volumes.get(volume_id=chainkeys_volume_id)
+            chainkeys_volume.remove(force=True)
+        except docker.errors.NotFound:
+            pass
+        except Exception:
+            logger.error(f'Cannot remove chainkeys volume {chainkeys_volume_id}', exc_info=True)
+
 
 # Instatiate Ressource Manager in BaseManager to share it between celery concurrent tasks
 BaseManager.register('ResourcesManager', ResourcesManager)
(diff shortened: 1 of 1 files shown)


Kelvin-M:
How can I test it ? :)

backend/substrapp/tasks/tasks.py
```diff
@@ -704,9 +708,8 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id,
                 except kubernetes.client.rest.ApiException as e:
                     logger.error(f'failed to remove secrets from namespace {secret_namespace}')
                     raise e
-            logger.info(f'{len(secrets)} secrets have been removed')
-
-        volumes[chainkeys_directory] = {'bind': '/sandbox/chainkeys', 'mode': 'rw'}
+            else:
+                logger.info(f'{len(secrets)} secrets have been removed')
```
inalgnu: the idea here was to replace the secrets before making them available to the task in a volume, instructions order is important here. Maybe we should add a comment.

Kelvin-M: I don't understand you comment. Secrets are retrieve from k8s secret and then dump into file in the volume which will be binded.
The instruction here, only erase secret value. I add the log in the else (for .. else loop) to make it explicit that logger is trigger if and only if we manage to do this because we raise if we don't. It does the same stuff than before, it just more explicit.

Kelvin-M: I don't understand you comment. Secrets are retrieve from k8s secret and then dump into file in the volume which will be binded.
The instruction here, only erase secret value. I add the log in the else (for .. else loop) to make it explicit that logger is trigger if and only if we manage to do this because we raise if we don't. It does the same stuff than before, it just more explicit.

backend/substrapp/tasks/tasks.py
```diff
@@ -391,16 +391,23 @@ def remove_subtuple_materials(subtuple_directory):
...
     try:
         local_volume = client.volumes.get(volume_id=volume_id)
         local_volume.remove(force=True)
     except docker.errors.NotFound:
         pass
     except Exception:
-        logger.error(f'Cannot remove local volume {volume_id}', exc_info=True)
+        logger.error(f'Cannot remove volume {volume_id}', exc_info=True)
+
+    chainkeys_directory = get_chainkeys_directory(compute_plan_id)
```
inalgnu: this should be done only if settings.TASK['CHAINKEYS_ENABLED'] is enabled

Kelvin-M:
Tested, it should work.

PR #168, opened 2020-02-23
gzip models during download

AurelienGasser:
1. Only gzip models, to be conservative. We could decide to gzip more routes (or all the routes) in the future.
2. The middleware takes care of not gzipping small payloads (<200b), not gzipping the payload if the the client doesn't support it, not gzipping if the response is already compressed, etc
3. The client (`requests.get`) already adds `Accept-Encoding: gzip,deflate`, and decompresses accordingly if the response has `Content-Encoding: gzip`. So we don't need to do anything client-side.

You can check that the response is actually encoded by doing a HTTP request manually with curl (and adding `Accept-Encoding: gzip`), or by adding this to `utils.py`

```python
enc = response.headers.get('Content-Encoding', 'unknown')
logger.info(f'Response encoding: {enc}')
```

### DIFF
--- a/backend/substrapp/views/model.py
+++ b/backend/substrapp/views/model.py
@@ -1,6 +1,8 @@
 import tempfile
 import logging
 from django.http import Http404
+from functools import wraps
+from django.middleware.gzip import GZipMiddleware
 from rest_framework import status, mixins
 from rest_framework.decorators import action
 from rest_framework.response import Response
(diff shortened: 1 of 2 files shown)


AurelienGasser:
tested with `substra-tests`: :heavy_check_mark:

AurelienGasser:
need to fix unit tests :roll_eyes:

backend/substrapp/views/model.py
```diff
@@ -116,6 +118,19 @@ def details(self, request, *args, **kwargs):
         return Response(data, status=status.HTTP_200_OK)
 
 
+def gzip_action(func):
+    gz = GZipMiddleware()
+
+    @wraps(func)
+    def wrapper(self, request, *args, **kwargs):
+        resp = func(self, request, *args, **kwargs)
+        if request.method == 'GET':
```
Kelvin-M: Only in the case of response with status code `200` will return a model content, in other case if will fail.
Maybe you can add something like this : `if request.method == 'GET' and resp.status_code == 200:`

AurelienGasser: I don't think it's necessary to add a condition because even if the request doesn't respond with `200`, compressing the response body doesn't hurt. The middleware won't compress short payloads anyway (<200 bytes), so "Page Not Found" kind-of-responses won't be compressed

We could even remove the condition `if request.method == 'GET'`, actually!

PR #172, opened 2020-02-26
Remove listed models from Compute plan event

thibaultrobert:
Companion PR : https://github.com/SubstraFoundation/substra-tests/pull/86 & https://github.com/SubstraFoundation/substra-chaincode/pull/106 & https://github.com/SubstraFoundation/substra/pull/135

### DIFF
--- a/backend/substrapp/serializers/ledger/computeplan/serializer.py
+++ b/backend/substrapp/serializers/ledger/computeplan/serializer.py
@@ -61,7 +61,7 @@ class LedgerComputePlanSerializer(serializers.Serializer):
     composite_traintuples = ComputePlanCompositeTrainTupleSerializer(many=True, required=False)
     aggregatetuples = ComputePlanAggregatetupleSerializer(many=True, required=False)
     tag = serializers.CharField(min_length=0, max_length=64, allow_blank=True, required=False)
-    cleanModels = serializers.BooleanField(required=False)
+    clean_models = serializers.BooleanField(required=False)
 
     def get_args(self, data):
         # convert snake case fields to camel case fields to match chaincode expected inputs
@@ -138,7 +138,7 @@ class LedgerComputePlanSerializer(serializers.Serializer):
             'compositeTraintuples': composite_traintuples,
             'aggregatetuples': aggregatetuples,
             'tag': data.get('tag'),
-            'cleanModels': data.get('cleanModels', False),
+            'cleanModels': data.get('clean_models', False),
         }
 
     def create(self, validated_data):
--- a/backend/substrapp/serializers/ledger/computeplan/serializer.py
+++ b/backend/substrapp/serializers/ledger/computeplan/serializer.py
@@ -61,6 +61,7 @@ class LedgerComputePlanSerializer(serializers.Serializer):
     composite_traintuples = ComputePlanCompositeTrainTupleSerializer(many=True, required=False)
     aggregatetuples = ComputePlanAggregatetupleSerializer(many=True, required=False)
     tag = serializers.CharField(min_length=0, max_length=64, allow_blank=True, required=False)
+    cleanModels = serializers.BooleanField(required=False)
 
     def get_args(self, data):
         # convert snake case fields to camel case fields to match chaincode expected inputs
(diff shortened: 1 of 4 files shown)


backend/substrapp/tasks/tasks.py
```diff
@@ -897,19 +897,48 @@ def remove_algo_images(algo_hashes):
...
     logger.info(f'Remove local volume of compute plan {compute_plan_id}')
-    remove_local_folder(compute_plan_id)
+
+    if status in ['done', 'failed', 'canceled']:
+        remove_local_folder(compute_plan_id)
 
     # Remove algorithm images
-    remove_algo_images(algo_hashes)
+    if algo_hashes:
+        remove_algo_images(algo_hashes)
```
thibaultrobert: We removed the algo images at the end of the compute plan I think.

backend/substrapp/serializers/ledger/computeplan/serializer.py
```diff
@@ -137,6 +138,7 @@ def get_args(self, data):
             'compositeTraintuples': composite_traintuples,
             'aggregatetuples': aggregatetuples,
             'tag': data.get('tag'),
+            'cleanModels': data.get('cleanModels', False),
```
samlesu:
```suggestion
            'cleanModels': data.get('clean_models', False),
```

To be consistent with the other endpoints, substra backend should accept snake case fields.

Kelvin-M:
@samlesu The docker/start.py is relatrd to the current PR. In the docker setup, the database used is the same, to the model asset removal remove everything from the db, without checking that the model is, in fact, owned by the node.
We need to split the service for both node to isolate the database.

samlesu:
Got it, thanks for explanation.
If I understand well, when using the docker-compose setup, a single Postgres database instance was used by the 2 nodes?

Kelvin-M:
@samlesu  Yes, so when receiving the event compute plan with the full list of model to delete (for every nodes), each node tries to delete the full list. It will successfully delete the db instance (as the db is shared) but the delete signal inherited will be able to delete model files that the node owned. So we will have a clean db, but not a clean file storage.

backend/substrapp/serializers/ledger/computeplan/serializer.py
Kelvin-M: Done

backend/substrapp/tasks/tasks.py
```diff
@@ -916,19 +919,43 @@ def remove_algo_images(algo_hashes):
     client = docker.from_env()
     for algo_hash in algo_hashes:
         algo_docker = get_algo_image_name(algo_hash)
-        logger.info(f'Remove docker image {algo_docker}')
-        client.images.remove(algo_docker, force=True)
+
+        try:
+            if client.images.get(algo_docker):
+                client.images.remove(algo_docker, force=True)
+                logger.info(f'Remove docker image {algo_docker}')
```
AurelienGasser: swap with the previous line?

backend/substrapp/tasks/tasks.py
```diff
@@ -916,19 +919,43 @@ def remove_algo_images(algo_hashes):
...
+            logger.exception(e)
+
+
+def remove_intermediary_models(model_hashes):
+    from substrapp.models import Model
+
+    models = Model.objects.filter(pkhash__in=model_hashes, validated=True)
+    filtered_model_hashes = [model.pk for model in models]
+
+    models.delete()
```
AurelienGasser: aren't the models also stored on disk?

Kelvin-M: Yes. Models are stored on the disk :-)
But there is a model signal that delete it when the instance is deleted !

backend/substrapp/tasks/tasks.py
Kelvin-M: Yes ! Nice catch

backend/substrapp/tasks/tasks.py
AurelienGasser: Would you mind pointing me to the piece of code that does that?

Also: have you checked the files actually get deleted on disk?

Kelvin-M: You can find the code here : https://github.com/SubstraFoundation/substra-backend/blob/master/backend/substrapp/signals/model/post_delete.py
Linked here : https://github.com/SubstraFoundation/substra-backend/blob/master/backend/substrapp/apps.py#L26
Yes, I've checked that models are removed from db and deleted on disk

AurelienGasser: Thnks for the links!

PR #175, opened 2020-02-26
add setting for gzip

AurelienGasser:


charts/substra-backend/values.yaml
```diff
@@ -18,6 +18,7 @@ backend:
   defaultDomain: localhost
   uwsgiProcesses: 8
   uwsgiThreads: 2
+  gzipModels: true
```
Kelvin-M: Should be false by default :)

AurelienGasser: oops, I forgot to change it to false

AurelienGasser:
@Kelvin-M tested: :heavy_check_mark:

PR #179, opened 2020-03-03
Improve worker

Kelvin-M:
- Remove resources manager
- Improve fetch memory limit
- Enhance logs

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -37,6 +37,13 @@ AGGREGATETUPLE_TYPE = 'aggregatetuple'
 COMPOSITE_TRAINTUPLE_TYPE = 'compositeTraintuple'
 TESTTUPLE_TYPE = 'testtuple'
 
+TUPLE_COMMANDS = {
+    TRAINTUPLE_TYPE: 'train',
+    TESTTUPLE_TYPE: 'predict',
+    COMPOSITE_TRAINTUPLE_TYPE: 'train',
+    AGGREGATETUPLE_TYPE: 'aggregate',
+}
+
 MODEL_FOLDER = '/sandbox/model'
 OUTPUT_HEAD_MODEL_FILENAME = 'head_model'
 OUTPUT_TRUNK_MODEL_FILENAME = 'trunk_model'
@@ -635,7 +642,8 @@ def _do_task(client, subtuple_directory, tuple_type, subtuple, compute_plan_id, 
     else:
         environment = {}
 
-    container_name, command = generate_container_name_and_cmd(tuple_type, subtuple, rank)
+    container_name = f'{tuple_type}_{subtuple["key"][0:8]}_{TUPLE_COMMANDS[tuple_type]}'
+    command = generate_command(tuple_type, subtuple, rank)
 
     compute_docker(
         client=client,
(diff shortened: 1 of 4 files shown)


backend/substrapp/tasks/tasks.py
```diff
@@ -626,27 +630,93 @@ def do_task(subtuple, tuple_type):
...
+        client, subtuple_directory, tuple_type, compute_plan_id, compute_plan_tag)
+
+    # Add node index to environment variable for the compute
+    node_index = os.getenv('NODE_INDEX')
+    if node_index:
+        environment = {'NODE_INDEX': node_index}
+    else:
+        environment = {}
+
+    container_name, command = generate_container_name_and_cmd(tuple_type, subtuple, rank)
```
AurelienGasser: Nit, and maybe out of scope: what about simplifying the container name (there's only one operation - train, predict, etc - we can do on each type, no need to specify it?) 

```suggestion
    command = get_command(tuple_type, subtuple, rank)
    container_name = f'{tuple_type}_{subtuple["key"][0:8]}'
```

backend/substrapp/tasks/utils.py
```diff
@@ -54,6 +54,52 @@ def get_and_put_asset_content(url, node_id, content_hash, content_dst_path, salt
...
+
+    try:
+        memory_limit_bytes = docker_client.containers.run(**task_args)
+    except (docker.errors.ContainerError, docker.errors.ImageNotFound, docker.errors.APIError):
+        logger.info('[Warning] Cannot get memory limit from remote')
+    else:
+        if memory_limit_bytes:
+            memory_limit = int(memory_limit_bytes)
+
+    return memory_limit
```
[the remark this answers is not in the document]
Kelvin-M: Done

backend/substrapp/tasks/tasks.py
```diff
@@ -626,27 +630,93 @@ def do_task(subtuple, tuple_type):
...
+        result['end_model_file_hash'] = models['end_model']['hash']
+
+    elif tuple_type == COMPOSITE_TRAINTUPLE_TYPE:
+        # Head model does not expose storage address
+        result['end_head_model_file_hash'] = models['end_head_model']['hash']
+
+        result['end_trunk_model_file'] = models['end_trunk_model']['file']
+        result['end_trunk_model_file_hash'] = models['end_trunk_model']['hash']
+
+    # Evaluation
```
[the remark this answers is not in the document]

backend/substrapp/tasks/tasks.py

backend/substrapp/tasks/utils.py
```diff
@@ -88,15 +86,23 @@ def get_memory_limit():
...
-        logger.info('[Warning] Cannot get memory limit from remote')
-    else:
-        if memory_limit_bytes:
-            memory_limit = int(memory_limit_bytes)
+        memory_limit = docker_memory_limit()
+    except (docker.errors.ContainerError, docker.errors.ImageNotFound,
+            docker.errors.APIError, ValueError):
+
+        logger.info('[Warning] Cannot get memory limit from remote, get it from local.')
+        memory_limit = local_memory_limit()
```
AurelienGasser: we're calling local_memory_limit twice (also on line 97). refactor?

Kelvin-M: Yes !

AurelienGasser: @Kelvin-M what about other exceptions that aren't covered on line 101-102?

Kelvin-M: It should not raise, if so we let the worker raise the exception :)

AurelienGasser: I don't understand what's the advantage of only catching specific exception types. 

If the wrapped method raises a new exception type in the future (e.g. following code changes in the wrapped method, or changes in a library), then the wrapper method will return nothing 😱 

I'll leave that up to you though

Kelvin-M: Here, we just want to catch exceptions that could happen and we are aware of.
In other case, I prefer to let the function raise an exception.
I don't understand the "wrapped method" part. For me it's just a classical function (with a decorator)

AurelienGasser: That makes sense, I withdraw my comment! (I had mistakenly assumed uncaught exceptions will not be propagated)

Re: wrapper/wrapped. The wrapped functions are the two functions that actually get the memory limit (docker/memory). The wrapper function is `get_memory_limit`. I used "wrapping" here to refer to nesting (one function calls another one, Russian-dolls like)

AurelienGasser:
e2e tests: :+1:

AurelienGasser:
@inalgnu would you have some time to review this ?

PR #180, opened 2020-03-04
Improve fetch models error handling and content stream

Kelvin-M:
During a test, we see that exceptions in thread are not handled correctly.
Moreover there is a proper way to handle streaming content from requests : https://stackoverflow.com/a/16696317

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -7,7 +7,6 @@ import tempfile
 from os import path
 import json
 from multiprocessing.managers import BaseManager
-from threading import Thread
 import logging
 import tarfile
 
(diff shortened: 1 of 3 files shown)


backend/substrapp/tasks/tasks.py
```diff
@@ -209,14 +208,24 @@ def prepare_traintuple_input_models(directory, tuple_):
...
     for proc in models:
         proc.join()
 
+    exceptions = []
+
+    for proc in models:
+        if hasattr(proc, "_exception"):
+            exceptions.append(proc._exception)
+            logger.exception(proc._exception)
+    else:
```
jmorel: This made me discover (again) the `for...else` syntax. Since we don't have a break statement in the loop isn't the else clause unecessary here? 
Unless this is an existing code convention, in which case nevermind.

Kelvin-M: No particular reason, I just do that because it concerns the same idea (looping, then raising)

backend/substrapp/tasks/tasks.py
```diff
@@ -228,14 +237,24 @@ def prepare_aggregatetuple_input_models(directory, tuple_):
...
+        if exceptions:
+            raise Exception(exceptions)
+
```
jmorel: This suggestion goes a bit beyond the purpose of the PR, but what do you think about putting this logic (running threads and catching exceptions) in a separate function and adding a test making sure that we catch all exceptions no matter in which thread they occur ?

Kelvin-M: Yes, see the last commit :)

jmorel: That was fast :)

PR #182, opened 2020-03-05
Dev Hot Reload

Kelvin-M:
Complete and solve : https://github.com/SubstraFoundation/substra-backend/pull/169
We may need to wait for @AurelienGasser or @ClementGautier to test it under `minikube` setup to see if everything work :)

Feel free to comment ! 

Note : 
- We use fixed prod settings in the env variable for server because we use static files in both cases :)

### DIFF
--- a/charts/substra-backend/templates/deployment-server.yaml
+++ b/charts/substra-backend/templates/deployment-server.yaml
@@ -34,9 +34,9 @@ spec:
         {{- end }}
         command: ["/bin/bash"]
         {{- if eq .Values.backend.settings "prod" }}
-        args: ["-c", "update-ca-certificates && uwsgi --http :{{ .Values.backend.service.port }} --module backend.wsgi --static-map /static=/usr/src/app/backend/statics --master --processes {{ .Values.backend.uwsgiProcesses }} --threads {{ .Values.backend.uwsgiThreads }} --need-app --env DJANGO_SETTINGS_MODULE=backend.settings.server.prod "]
+        args: ["-c", "update-ca-certificates && uwsgi --http :{{ .Values.backend.service.port }} --module backend.wsgi --static-map /static=/usr/src/app/backend/statics --master --processes {{ .Values.backend.uwsgiProcesses }} --threads {{ .Values.backend.uwsgiThreads }} --need-app --env DJANGO_SETTINGS_MODULE=backend.settings.server.{{ .Values.backend.settings }} "]
         {{- else }}
-        args: ["-c", "update-ca-certificates && python manage.py runserver --noreload 0.0.0.0:{{ .Values.backend.service.port }} --settings=backend.settings.server.{{ .Values.backend.settings }}"]
+        args: ["-c", "update-ca-certificates && uwsgi --http :{{ .Values.backend.service.port }} --module backend.wsgi --static-map /static=/usr/src/app/backend/statics --master --processes {{ .Values.backend.uwsgiProcesses }} --threads {{ .Values.backend.uwsgiThreads }} --need-app --env DJANGO_SETTINGS_MODULE=backend.settings.server.{{ .Values.backend.settings }} --py-autoreload 3"]
         {{- end }}
         env:
           - name: DJANGO_SETTINGS_MODULE
@@ -150,19 +150,17 @@ spec:
           value: {{ .Values.postgresql.postgresqlPassword }}
         - name: DATABASE_HOST
           value: {{ .Release.Name }}-postgresql
-      {{- if eq .Values.backend.settings "prod" }}
       - name: init-collectstatic
         image: "{{ .Values.backend.image.repository }}:{{ .Values.backend.image.tag }}"
         command: ['python', 'manage.py', 'collectstatic', '--noinput']
         env:
           - name: DJANGO_SETTINGS_MODULE
-            value: backend.settings.{{ .Values.backend.settings }}
+            value: backend.settings.prod
           - name: BACKEND_ORG
             value: {{ .Values.organization.name }}
         volumeMounts:
           - name: statics
             mountPath: /usr/src/app/backend/statics
-      {{- end }}
       volumes:
       {{- range .Values.persistence.volumes }}
       - name: data-{{ .name }}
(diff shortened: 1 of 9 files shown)


charts/substra-backend/templates/deployment-server.yaml
```diff
@@ -135,19 +135,17 @@ spec:
           value: {{ .Values.postgresql.postgresqlPassword }}
         - name: DATABASE_HOST
           value: {{ .Release.Name }}-postgresql
-      {{- if eq .Values.backend.settings "prod" }}
```
AurelienGasser: Could you please add context regarding this change?

AurelienGasser:
I just had a quick look without properly testing: it looks great! 

The only thing that I find missing (unless I missed something) is the auto reload of _libraries_. I find it very useful, in my workflow, to also be able to add debug code in libraries and quickly reload. For some types of problems it's very hard to do without it (or rather, very _slow_ ;) ) However, this could be addressed separately, outside of the scope of this PR.

I'll test this properly in minikube when I get a chance.

Kelvin-M:
@AurelienGasser 
What do you mean by "auto reload of libraries" ?
You want to hot reload patched python libraries ? It is possible but you will need to do lot of extra work because there are not watched and are installed via pip !

AurelienGasser:
@Kelvin-M yes, that's what I mean. That's the one thing that is missing from #169.

charts/substra-backend/templates/deployment-server.yaml
Kelvin-M: Because, the new way we serve the server needs the statics files :)

charts/substra-backend/templates/deployment-server.yaml
```diff
@@ -140,19 +140,17 @@ spec:
...
       - name: init-collectstatic
         image: "{{ .Values.backend.image.repository }}:{{ .Values.backend.image.tag }}"
         command: ['python', 'manage.py', 'collectstatic', '--noinput']
         env:
           - name: DJANGO_SETTINGS_MODULE
-            value: backend.settings.{{ .Values.backend.settings }}
+            value: backend.settings.prod
```
AlexandrePicosson, line 148: Why is this now a fixed value ?
And is it necessary to keep it because we pass this as `--env DJANGO_SETTINGS_MODULE=backend.settings.{{ .Values.backend.settings }}` at these lines https://github.com/SubstraFoundation/substra-backend/blob/e6a0cb62abcb3dab2d5839fcf0094d55a700dabb/charts/substra-backend/templates/deployment-server.yaml#L37 and https://github.com/SubstraFoundation/substra-backend/blob/e6a0cb62abcb3dab2d5839fcf0094d55a700dabb/charts/substra-backend/templates/deployment-server.yaml#L39

docker/celerybeat/Dockerfile
```diff
@@ -10,6 +10,7 @@ WORKDIR /usr/src/app
 COPY ./backend/requirements.txt /usr/src/app/.
 
 RUN pip3 install -r requirements.txt
+RUN pip3 install watchdog argh
```
AlexandrePicosson, line 13: Should this be before `RUN pip3 install -r requirements.txt`, maybe even before `RUN pip3 install -r requirements.txt`  because we expect this layer to change less often so we want to keep this as a cached layer and not rebuild it every time we change the requirements.

docker/celeryworker/Dockerfile
```diff
@@ -10,6 +10,7 @@ WORKDIR /usr/src/app
 COPY ./backend/requirements.txt /usr/src/app/.
 
 RUN pip3 install -r requirements.txt
+RUN pip3 install watchdog argh
```
AlexandrePicosson, line 13: Same comment, maybe put this before in the file.

docker/flower/Dockerfile
```diff
@@ -10,7 +10,7 @@ WORKDIR /usr/src/app
 COPY ./backend/requirements.txt /usr/src/app/.
 
 RUN pip3 install -r requirements.txt
-RUN pip3 install flower
+RUN pip3 install flower watchdog argh
```
AlexandrePicosson, line 13: Same comment, maybe put this before in the file.

charts/substra-backend/templates/deployment-server.yaml
Kelvin-M: This is not a fix value. As said previously in the Aurelien's comment, we need to use `prod` settings to generate static files, as we expose now the server with uswgi in both cases

docker/celeryworker/Dockerfile
Kelvin-M: I don't know if we should add this to the requirements.txt as it's only a requirement for the dev settings (to be able to do the hot reload). Feel free to update.

docker/flower/Dockerfile
Kelvin-M: I don't know if we should add this to the requirements.txt as it's only a requirement for the dev settings (to be able to do the hot reload). Feel free to update.

docker/celerybeat/Dockerfile
Kelvin-M: Yes, but I don't know if we want to add this as a requirement or not.

charts/substra-backend/templates/deployment-server.yaml
Kelvin-M: @AlexandrePicosson Good for you ?

docker/celerybeat/Dockerfile
AlexandrePicosson: I was not suggesting to put it in the requirements.txt just adding this command to the line 9 of the dockerfile instead of the line 13 to keep it cached for docker builds as it will probably change less often.

Kelvin-M: Yes, it's my suggestion, if we put in in the requirements.txt we don't need to swap lines but only to remove one :)

charts/substra-backend/templates/deployment-server.yaml
AlexandrePicosson: Yes, I was just not sure of what this env variable would change and thought it might activate some security settings we do not want in dev.

docker/celerybeat/Dockerfile
AlexandrePicosson: As all images we build depend on this I think we should put it in the requirements.txt, I think it's better to manage all the python dependencies in the same place.
Adding it to the Dockerfile we make this a dependency of the images we publish even if it is not really required.

AurelienGasser:
^ rebased master

AurelienGasser:
Tested backend reload: :+1:

AurelienGasser:
@inalgnu would you have some time to review this?

PR #183, opened 2020-03-05
Streamline Model view

### LINKED ISSUE #160: Streamline model view
The model view has a lot of peculiarities:
- `/model` doesn't return models per se, but instead returns the list of all traintuples/composite traintuples/aggregatetuples each with the associated certified testtuple (even though this concept doesn't really exists anymore)
- `/model/<traintuple_key>` only works for traintuples and returns the traintuple and all the linked testtuples. It also creates a local cache of each outModel. It fails with a 500 if the key doesn't match a traintuple but rather a composite traintuple for example.
- `/model/<tuple_key>/details` (where `tuple_key` is either a `composite_traintuple_key`, a `traintuple_key` or an `aggregate_key`) returns the tuple and all the linked testtuples
- `/model/<model_hash>/file` streams the content of an outModel

A streamlined schema could be:

- `/model` returns the list of all traintuples/composite traintuples/aggregatuples
- `/model/<tuple_key>` returns the matching traintuple / composite traintuple / aggregatetuple with all linked testuples
- `/model/<tuple_key>/file` streams the content of the outModel (for traintuple / aggregatetuple) or the content of the outTrunkModel (for composite traintuple)

Kelvin-M:
Related to #160 
Companion PR : https://github.com/SubstraFoundation/substra-frontend/pull/43

Merge retrieve and details view

This PR doesn't change the `model/<model_hash>/file` path because:
- this path is only use between node
- the ledger call is different that other path : `queryModelPermissions ` and not `queryModelDetails`
- model file are registered in the db (and the ledger via storage address) with the hash of the model. To change it, we will need to change the chaincode.

### DIFF
--- a/backend/events/apps.py
+++ b/backend/events/apps.py
@@ -188,7 +188,7 @@ def wait():
                                                              'chaincode-updates',
                                                              onEvent=on_event)
 
-                    logger.error(f'Connect to Channel Event Hub')
+                    logger.error('Connect to Channel Event Hub')
                     loop.run_until_complete(stream)
 
                 except RpcError as e:
--- a/backend/substrapp/serializers/__init__.py
+++ b/backend/substrapp/serializers/__init__.py
@@ -2,14 +2,13 @@
 
 from .datasample import DataSampleSerializer
 from .objective import ObjectiveSerializer
-from .model import ModelSerializer
 from .datamanager import DataManagerSerializer
 from .algo import AlgoSerializer
 from .compositealgo import CompositeAlgoSerializer
 from .aggregatealgo import AggregateAlgoSerializer
 from .ledger import *
 
-__all__ = ['DataSampleSerializer', 'ObjectiveSerializer', 'ModelSerializer',
+__all__ = ['DataSampleSerializer', 'ObjectiveSerializer',
            'DataManagerSerializer', 'AlgoSerializer', 'CompositeAlgoSerializer',
            'LedgerObjectiveSerializer', 'LedgerModelSerializer',
            'LedgerDataSampleSerializer', 'LedgerDataSampleUpdateSerializer',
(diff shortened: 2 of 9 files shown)


jmorel:
This may not be related to this PR but while testing I saw that accessing `/model/<testtuple_key>` with a valid testtuple key returns an empty testtuple:
```
{
    "testtuple": {
        "algo": null,
        "certified": false,
        "computePlanID": "",
        "creator": "",
        "dataset": null,
        "key": "",
        "log": "",
        "objective": null,
        "rank": 0,
        "status": "",
        "tag": "",
        "traintupleKey": "",
        "traintupleType": ""
    },
    "nonCertifiedTesttuples": null
}
```
I guess passing a testtuple should raise an error (we only accept composite, aggregate and train tuple keys)

jmorel:
I tested commenting the `serializer_class = ModelSerializer` in the `ModelView` and had no trouble at all. This definition was only here so that the get_serializer method in the _retrieve would work. Since we don't use it at all in our views (and don't use the default viewset views) I think it should be safe to remove not only the line but also the whole ModelSerializer from the codebase.

jmorel:
I made a companion PR for the frontend: https://github.com/SubstraFoundation/substra-frontend/pull/43

Kelvin-M:
@jmorel I've updated the PR :)

Kelvin-M:
@jmorel @samlesu @inalgnu  :  Let's sync to merge this one or there is something that prevent from doing it ?

jmorel:
It's fine on my side. Since this was written quite some time ago we should rebase it and run the e2e tests just to be sure, but otherwise I'm fine with merging it.

Kelvin-M:
@jmorel I rebase it and do a e2e tests, everything works :)

PR #184, opened 2020-03-16
Add service account for worker

AlexandrePicosson:
Add ServiceAccount for the worker making it able to list, retrieve and create secrets in his workspace.
This PR could be adapted to add different Kubernetes ressources the worker would need to access.

### DIFF
--- a/charts/substra-backend/templates/rbac.yaml
+++ b/charts/substra-backend/templates/rbac.yaml
@@ -1,0 +1,46 @@
+{{- if .Values.celeryworker.rbac.enable }} 
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "substra.fullname" . }}
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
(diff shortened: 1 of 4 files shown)


charts/substra-backend/templates/rbac.yaml
```diff
@@ -0,0 +1,46 @@
+{{- if .Values.celeryworker.rbac.enable }} 
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "substra.fullname" . }}
```
ClementGautier: If the SA is specific to the celeryworker I'd add a suffix to the SA name. It looks like it's a global SA the way it is now.
(If its the other way - if the SA can be used by all pods - then you might want to change celeryworker.rbac.enable for rbac.enable and make use of it on all other pods)

AlexandrePicosson: Yes at the moment this SA would be used only by the worker, I will add a suffix.

AlexandrePicosson:
Forgot to merge this PR so I rebased the branch on current master and changed the chart version.

PR #186, opened 2020-03-18
Add user metadata

maeldebon:
🔗 **Related PRs:**
- [substra-tests](https://github.com/SubstraFoundation/substra-tests/pull/90)
- [substra-chaincode](https://github.com/SubstraFoundation/substra-chaincode/pull/111)
- [substra](https://github.com/SubstraFoundation/substra/pull/143)
- [substra-frontend (coming soon)](#)

The goal of these PRs is to add a `metadata` field in every asset in order to be able to store information related to this asset. It could be, for instance, a creation_ts, an epoch, a project_name, ...

### DIFF
--- a/backend/substrapp/serializers/ledger/objective/serializer.py
+++ b/backend/substrapp/serializers/ledger/objective/serializer.py
@@ -17,7 +17,7 @@ class LedgerObjectiveSerializer(serializers.Serializer):
     test_data_manager_key = serializers.CharField(max_length=256, allow_blank=True, required=False, allow_null=True)
     permissions = PermissionsSerializer()
     metrics_name = serializers.CharField(min_length=1, max_length=100)
-    metadata = DictField(child=CharField(), required=False)
+    metadata = DictField(child=CharField(), required=False, allow_null=True)
 
     def create(self, validated_data):
         instance = self.initial_data.get('instance')
@@ -26,7 +26,7 @@ class LedgerObjectiveSerializer(serializers.Serializer):
         permissions = validated_data.get('permissions')
         test_data_manager_key = validated_data.get('test_data_manager_key', '')
         test_data_sample_keys = validated_data.get('test_data_sample_keys', [])
-        metadata = validated_data.get('metadata', {})
+        metadata = validated_data.get('metadata')
 
         # TODO, create a datamigration with new Site domain name when we will know the name of the final website
         current_site = getattr(settings, "DEFAULT_DOMAIN")
--- a/backend/substrapp/serializers/ledger/objective/serializer.py
+++ b/backend/substrapp/serializers/ledger/objective/serializer.py
@@ -24,6 +24,7 @@ class LedgerObjectiveSerializer(serializers.Serializer):
         permissions = validated_data.get('permissions')
         test_data_manager_key = validated_data.get('test_data_manager_key', '')
         test_data_sample_keys = validated_data.get('test_data_sample_keys', [])
+        metadata = validated_data.get('metadata', {})
 
         # TODO, create a datamigration with new Site domain name when we will know the name of the final website
         current_site = getattr(settings, "DEFAULT_DOMAIN")
@@ -42,6 +43,7 @@ class LedgerObjectiveSerializer(serializers.Serializer):
             'permissions': {'process': {
                 'public': permissions.get('public'),
                 'authorizedIDs': permissions.get('authorized_ids'),
-            }}
+            }},
+            'metadata': metadata
         }
         return ledger.create_objective(args, instance.pkhash)
--- a/backend/substrapp/serializers/ledger/objective/serializer.py
+++ b/backend/substrapp/serializers/ledger/objective/serializer.py
@@ -17,7 +17,7 @@ class LedgerObjectiveSerializer(serializers.Serializer):
     test_data_manager_key = serializers.CharField(max_length=256, allow_blank=True, required=False, allow_null=True)
     permissions = PermissionsSerializer()
     metrics_name = serializers.CharField(min_length=1, max_length=100)
-    metadata = DictField(child=CharField(), required=False, allow_null=True)
+    metadata = DictField(child=CharField(), required=False, allow_null=True, allow_empty=True)
 
     def create(self, validated_data):
         instance = self.initial_data.get('instance')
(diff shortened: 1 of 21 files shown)


AurelienGasser:
LGTM. However I think the create/update datetime should not be part of the metadata. For good consistency and separation of responsibilities, the client (or backend) should not be able to set the creation/update time: it should be set by the chaincode instead,  based on the transaction proposal timestamp. Another reason to exclude it from the metadata is types: it is a good practice to avoid manipulating dates as strings if possible.

maeldebon:
> LGTM. However I think the create/update datetime should not be part of the metadata. For good consistency and separation of responsibilities, the client (or backend) should not be able to set the creation/update time: it should be set by the chaincode instead, based on the transaction proposal timestamp. Another reason to exclude it from the metadata is types: it is a good practice to avoid manipulating dates as strings if possible.

I'm agree! It was the example provided but it could indeed be automatically defined. 👍

samlesu:
> LGTM. However I think the create/update datetime should not be part of the metadata. For good consistency and separation of responsibilities, the client (or backend) should not be able to set the creation/update time: it should be set by the chaincode instead, based on the transaction proposal timestamp. Another reason to exclude it from the metadata is types: it is a good practice to avoid manipulating dates as strings if possible.

I fully agree with your comment. That's however something we could expect from our users if we don't provide it at a system level.

backend/substrapp/tests/views/tests_views_algo.py
```diff
@@ -204,14 +204,19 @@ def test_algo_create(self):
...
+                    "bar": "foo"
+                }}
 
         with mock.patch.object(LedgerAlgoSerializer, 'create') as mcreate:
 
             mcreate.return_value = {}
 
             response = self.client.post(url, data=data, format='multipart', **self.extra)
 
+        assert response.data['metadata'] is not None
```
jmorel: Since you know what the metadata must be, you could assert that:

```python
assert response.data['metadata'] ==  {
                    "foo": "bar",
                    "bar": "foo"
                }
```

AurelienGasser:
General question: do we have an immediate use-case for this feature? Or maybe planned use-cases?

thibaultrobert:
> I fully agree with your comment. That's however something we could expect from our users if we don't provide it at a system level.
Well I think that an issue for the chaincode then.

samlesu:
> Well I think that an issue for the chaincode then.

Could you explain why?

samlesu:
> General question: do we have an immediate use-case for this feature?

There is no immediate use-case for this field. It's an enhancement of the current tag field. The tag field is already used in some projects and we could already see it's limitations: how to use a single tag field to store multiple informations?

backend/substrapp/tests/views/tests_views_algo.py
```diff
@@ -204,12 +204,15 @@ def test_algo_create(self):
...
                 'permissions_public': True,
-                'permissions_authorized_ids': []}
+                'permissions_authorized_ids': [],
+                'metadata_one': 'foo',
+                'metadata_two': 'bar'
+                }
 
         with mock.patch.object(LedgerAlgoSerializer, 'create') as mcreate:
 
+            assert data['metadata_one'] == 'foo'
```
jmorel: This only tests the literal declaration of data above, it doesn't test what data is transmitted to the the ledger through the create method of LedgerAlgoSerializer.

You could test this after the `post` (from L217 and on), by inspecting the call_args of mcreate (see https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args) and making sure the dict contains `{'metadata': {'one': 'foo', 'two': 'bar'}}`.

backend/substrapp/views/utils.py
```diff
@@ -149,6 +149,17 @@ def _download_remote_file(self, ledger_field, asset):
         return response
 
 
+def get_metadata_as_dict(data):
```
jmorel, line 152: Could you add a docstring explaining what this function does?

backend/substrapp/views/aggregatealgo.py
```diff
@@ -87,7 +88,8 @@ def _create(self, request, file):
         serializer = self.get_serializer(data={
             'pkhash': pkhash,
             'file': file,
-            'description': request.data.get('description')
+            'description': request.data.get('description'),
+            'metadata': request.data.get('metadata')
```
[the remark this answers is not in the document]
samlesu: As pointed by @jmorel, this comment needs to be addressed.

backend/substrapp/serializers/ledger/objective/serializer.py
```diff
@@ -16,6 +17,7 @@ class LedgerObjectiveSerializer(serializers.Serializer):
     test_data_manager_key = serializers.CharField(max_length=256, allow_blank=True, required=False, allow_null=True)
     permissions = PermissionsSerializer()
     metrics_name = serializers.CharField(min_length=1, max_length=100)
+    metadata = DictField(child=CharField(), required=False, allow_null=True, allow_empty=True)
```
jmorel, line 20: This serializer uses allow_empty=True unlike all other serializers. Should it be removed?

backend/substrapp/serializers/ledger/objective/serializer.py
```diff
@@ -24,6 +26,7 @@ def create(self, validated_data):
         permissions = validated_data.get('permissions')
         test_data_manager_key = validated_data.get('test_data_manager_key', '')
         test_data_sample_keys = validated_data.get('test_data_sample_keys', [])
+        metadata = validated_data.get('metadata')
```
jmorel, line 29: You use `metadata = validated_data.get('metadata', {})` everywhere else, which one should it be?

backend/substrapp/tests/views/tests_views_algo.py
```diff
@@ -207,6 +207,10 @@ def test_algo_create(self):
                 'permissions': {
                     'public': True,
                     'authorized_ids': [],
+                },
+                'metadata': {
+                    "one": "foo",
+                    "two": "bar"
```
jmorel, line 213: It looks like this is never tested in the return value. Since just adding these lines doesn't do anything maybe we should leave this file untouched?

maeldebon: I agree, I'll remove this!

backend/substrapp/serializers/ledger/objective/serializer.py
maeldebon: The `{}` wasn't necessary, I removed it.

backend/substrapp/serializers/ledger/objective/serializer.py
maeldebon: Yes indeed, thanks!

PR #189, opened 2020-03-24
Prevent aggregatuples with less than 2 inmodels

### LINKED ISSUE #187: Create an aggregatetuple without in model key
# Original issue from the `substra` repository

> It is possible to create an aggregatetuple without passing any --in-model-key in the command.

# Issue in the backend

The `in_models_keys` field of an aggregatetuple isn't required and has a minimum length of 0. 

As the sole purpose of the aggregatetuple is to aggregate at least 2 models together, the `in_models_keys` should be required and should have a minimum length of 2.

jmorel:
Fixes #187 

Prevent the creation of aggregatetuples with no in models or just 1 inmodel, since they don't make much sense (no aggregation to do).

Companion PR: https://github.com/SubstraFoundation/substra/pull/149


samlesu:
After discussing with @camillemarini, it seems better to not add this limitation. 

There is no reason for restricting the aggregation of nothing or the aggregation of a single model.

PR #192, opened 2020-03-27
Prevent from path traversal when saving models or extracting archives

Kelvin-M:
Filenames should be validated when storing asset to prevent from path/directory traversal.
Companion PR :  https://github.com/SubstraFoundation/substra/pull/151

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -22,7 +22,7 @@ from celery.task import Task
 import boto3
 
 from backend.celery import app
-from substrapp.utils import get_hash, get_owner, create_directory, uncompress_content
+from substrapp.utils import get_hash, get_owner, create_directory, uncompress_content, raise_if_path_traversal
 from substrapp.ledger_utils import (log_start_tuple, log_success_tuple, log_fail_tuple,
                                     query_tuples, LedgerError, LedgerStatusError, get_object_from_ledger)
 from substrapp.tasks.utils import (ResourcesManager, compute_docker, get_asset_content, get_and_put_asset_content,
@@ -181,6 +181,7 @@ def fetch_model(parent_tuple_type, authorized_types, input_model, directory):
         raise TasksError(f'{parent_tuple_type.capitalize()}: invalid input model: type={tuple_type}')
 
     model_dst_path = path.join(directory, f'model/{input_model["traintupleKey"]}')
+    raise_if_path_traversal([model_dst_path], path.join(directory, 'model/'))
 
     if tuple_type == TRAINTUPLE_TYPE:
         get_and_put_model_content(
@@ -259,6 +260,7 @@ def prepare_composite_traintuple_input_models(directory, tuple_):
         raise TasksError(f'CompositeTraintuple: invalid head input model: type={tuple_type}')
     # get the output head model
     head_model_dst_path = path.join(directory, f'model/{PREFIX_HEAD_FILENAME}{head_model_key}')
+    raise_if_path_traversal([head_model_dst_path], path.join(directory, 'model/'))
     get_and_put_local_model_content(
         head_model_key, metadata['outHeadModel']['outModel'], head_model_dst_path
     )
@@ -267,6 +269,7 @@ def prepare_composite_traintuple_input_models(directory, tuple_):
     trunk_model_key = trunk_model['traintupleKey']
     tuple_type, metadata = find_training_step_tuple_from_key(trunk_model_key)
     trunk_model_dst_path = path.join(directory, f'model/{PREFIX_TRUNK_FILENAME}{trunk_model_key}')
+    raise_if_path_traversal([trunk_model_dst_path], path.join(directory, 'model/'))
     # trunk model must refer to a composite traintuple or an aggregatetuple
     if tuple_type == COMPOSITE_TRAINTUPLE_TYPE:  # get output trunk model
         get_and_put_model_content(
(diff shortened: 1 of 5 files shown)


backend/substrapp/utils.py
```diff
@@ -108,6 +108,25 @@ def create_directory(directory):
...
+        os.path.realpath(to_directory),
+        ''
+    )
+
+    if not isinstance(requested_paths, list):
+        raise TypeError(f'requested_paths argument should be a list not a {type(requested_paths)}')
+
+    for requested_path in requested_paths:
+        real_requested_path = os.path.realpath(requested_path)
+        path_traversal = os.path.commonprefix([real_requested_path, safe_directory]) != safe_directory
```
AlexandrePicosson: Same comment as in https://github.com/SubstraFoundation/substra/pull/151 i'm not sure that this code really prevent from path traversal.

backend/substrapp/tests/tests_misc.py
```diff
@@ -20,9 +27,39 @@ def write(self, s):
...
+
+    def getnames(self):
+        return self.files
+
+
 class MiscTests(TestCase):
     """Misc tests"""
 
+    def setUp(self):
+        if not os.path.exists(DIRECTORY):
```
samlesu: You could use a temporary folder using the tempfile library (`/tmp` might not be available and it would be safer): https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp ?

backend/substrapp/tests/tests_misc.py
```diff
@@ -99,3 +136,29 @@ def test_query_tuples(self):
...
+            mock_is_tarfile.return_value = True
+            mock_tarfile.return_value = MockArchive()
+
+            self.assertRaises(Exception,
+                              uncompress_path('', DIRECTORY))
+
+        # Models
+        with self.assertRaises(Exception):
+            model_dst_path = os.path.join(DIRECTORY, f'model/../../hackermodel')
+            raise_if_path_traversal([model_dst_path], os.path.join(DIRECTORY, 'model/'))
```
samlesu: What about zipping and unzipping (same for tar) real archives instead of using mocking? 
It could be a little easier to maintain these tests and would allow to test a bit more?

backend/substrapp/utils.py
```diff
@@ -108,6 +108,25 @@ def create_directory(directory):
         os.makedirs(directory)
 
 
+def raise_if_path_traversal(requested_paths, to_directory):
+    # Inspired from https://stackoverflow.com/a/45188896
+
+    safe_directory = os.path.join(
+        os.path.realpath(to_directory),
+        ''
+    )
```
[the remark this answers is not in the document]
Kelvin-M: Maybe I should add a comment for this !

backend/substrapp/utils.py
```diff
@@ -108,6 +108,25 @@ def create_directory(directory):
         os.makedirs(directory)
 
 
+def raise_if_path_traversal(requested_paths, to_directory):
```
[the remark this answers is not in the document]
Kelvin-M: Can you explain why ?
I use requested paths because it's paths requested by the archive for extraction or the model from web request :)

backend/substrapp/utils.py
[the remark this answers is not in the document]
Kelvin-M: Done

backend/substrapp/tests/tests_misc.py
Kelvin-M: Yes, but other test use also `/tmp/xxx` like  `tests_task.py`.
But I don't use it in the test so I will remove it :)

backend/substrapp/tests/tests_misc.py
Kelvin-M: I thought about that ! But as we only care about filename path it easier to use a mock !

backend/substrapp/utils.py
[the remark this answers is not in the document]
samlesu: I thought that this function could be used outside of this context. My comment does not really make sense, let's stick with `requested_paths`, it's good.

backend/substrapp/tests/tests_misc.py
samlesu: Makes sense.

backend/substrapp/tests/tests_misc.py
samlesu: I prefer to avoid mocking when I can. Seems ok to me.

jmorel:
Would it be possible to also perform this check when adding an algo or an objective . This way invalid assets would be rejected right away with an error message instead of "just" making tuples that use them fail mysteriously.

Kelvin-M:
@jmorel  Yes we can, can you list all assets where the add should be validated ?

jmorel:
All the assets that use an archive, so :
* algo,
* composite algo
* aggregate algo
* objective

backend/substrapp/tasks/tasks.py
```diff
@@ -181,6 +181,7 @@ def fetch_model(parent_tuple_type, authorized_types, input_model, directory):
         raise TasksError(f'{parent_tuple_type.capitalize()}: invalid input model: type={tuple_type}')
 
     model_dst_path = path.join(directory, f'model/{input_model["traintupleKey"]}')
+    raise_if_path_traversal([model_dst_path], path.join(directory, 'model/'))
```
jmorel: Since we're creating the model_dst_path just above from the directory and the (validated) traintupleKey, can't we just drop this check? I'm not seeing how it could fail and raise.

jmorel: The same would apply to all checks in this file, I'm clearly missing something, could you explain it to me?

Kelvin-M: As we build the destination path base on the external value `input_model["traintupleKey"]` we (the backend) have no guarantee that `input_model["traintupleKey"]` is not path traversal, so should verify it :)

PR #194, opened 2020-03-27
Enable browsable API in dev mode only

jmorel:
This has been tested against substra-tests in dev mode with success.

I tried running it against skaffold modified as follows to emulate a prod mode. I only had one test failing: test_execution_retry_on_fail. But it looks like it is a known bug not linked to these changes: https://github.com/SubstraFoundation/substra-tests/issues/48

```diff
diff --git a/skaffold.yaml b/skaffold.yaml
index d79ccb9d..6951f670 100644
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -39,7 +39,7 @@ deploy:
           secrets:
             fabricConfigmap: network-org-1-hlf-k8s-fabric
           backend:
-            settings: dev
+            settings: prod
             defaultDomain: http://substra-backend.node-1.com
             ingress:
               enabled: true
@@ -81,7 +81,7 @@ deploy:
           secrets:
             fabricConfigmap: network-org-2-hlf-k8s-fabric
           backend:
-            settings: dev
+            settings: prod
             defaultDomain: http://substra-backend.node-2.com
             ingress:
               enabled: true
```

The frontend isn't working with this pseudo-prod mode, but I've checked that it's also an issue with the master branch, so not introduced here. To be more precise, I can login using the frontend, but then calls to /objective, /data_manager or /algo all fail with the following error:
```json
{
  "detail": "Given token not valid for any token type",
  "code": "token_not_valid",
  "messages": [
    {
      "token_class": "AccessToken",
      "token_type": "access",
      "message": "Token is invalid or expired"
    }
  ]
}
```

### DIFF
--- a/backend/backend/settings/dev.py
+++ b/backend/backend/settings/dev.py
@@ -9,10 +9,10 @@ DEBUG = True
 # LEDGER_CALL_RETRY = False  # uncomment to overwrite the ledger setting value
 
 # Enable Browsable API
-REST_FRAMEWORK['DEFAULT_RENDERER_CLASSES'] = REST_FRAMEWORK['DEFAULT_RENDERER_CLASSES'] + (
+REST_FRAMEWORK['DEFAULT_RENDERER_CLASSES'] += (
     'rest_framework.renderers.BrowsableAPIRenderer',
 )
-REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES'] = REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES'] + [
+REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES'] += [
     'libs.session_authentication.CustomSessionAuthentication',
 ]
 
--- a/backend/backend/settings/dev.py
+++ b/backend/backend/settings/dev.py
@@ -8,6 +8,14 @@ from .deps.restframework import *
 DEBUG = True
 # LEDGER_CALL_RETRY = False  # uncomment to overwrite the ledger setting value
 
+# Enable Browsable API
+REST_FRAMEWORK['DEFAULT_RENDERER_CLASSES'] = REST_FRAMEWORK['DEFAULT_RENDERER_CLASSES'] + (
+    'rest_framework.renderers.BrowsableAPIRenderer',
+)
+REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES'] = REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES'] + [
+    'libs.session_authentication.CustomSessionAuthentication',
+]
+
 # Database
 # https://docs.djangoproject.com/en/2.0/ref/settings/#databases
 
--- a/backend/substrapp/views/utils.py
+++ b/backend/substrapp/views/utils.py
@@ -2,11 +2,11 @@ import os
 
 
 from django.http import FileResponse, HttpResponse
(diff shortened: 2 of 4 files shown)


backend/backend/settings/dev.py
```diff
@@ -8,6 +8,14 @@
 DEBUG = True
 # LEDGER_CALL_RETRY = False  # uncomment to overwrite the ledger setting value
 
+# Enable Browsable API
+REST_FRAMEWORK['DEFAULT_RENDERER_CLASSES'] = REST_FRAMEWORK['DEFAULT_RENDERER_CLASSES'] + (
```
samlesu:
```suggestion
REST_FRAMEWORK['DEFAULT_RENDERER_CLASSES'] += (
```
Should work but not tested.

backend/backend/settings/dev.py
```diff
@@ -8,6 +8,14 @@
...
+    'rest_framework.renderers.BrowsableAPIRenderer',
+)
+REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES'] = REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES'] + [
```
samlesu:
```suggestion
REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES'] += [
```
Idem

backend/substrapp/views/utils.py
```diff
@@ -54,12 +52,7 @@ def node_has_process_permission(asset):
 
 
 class PermissionMixin(object):
-    authentication_classes = [
-        BasicAuthentication,  # for node to node
-        SecureJWTAuthentication,  # for user from front/sdk/cli
-        TokenAuthentication,  # for user from front/sdk/cli
-        CustomSessionAuthentication,  # for user on drf web browsable api
-    ]
+    authentication_classes = api_settings.DEFAULT_AUTHENTICATION_CLASSES + [BasicAuthentication]
```
samlesu: Could you explain the purpose of this change? `api_settings.DEFAULT_AUTHENTICATION_CLASSES` seems not equivalent to `(BasicAuthentication, TokenAuthentication)`.

jmorel: `api_settings.DEFAULT_AUTHENTICATION_CLASSES` will contain `SecureJWTAuthentication` and `ExpiryTokenAuthentication` (and `CustomSessionAuthentication` in dev mode). By adding BasicAuth to it, we're equivalent to the existing list when in dev mode and just miss the `CustomSessionAuthentication ` in prod mode.

There is one change though: `ExpiryTokenAuthentication` instead of `TokenAuthentication`. Not an issue since they are fully compatible, and we get the added benefit of checking the expiry date of the token (which we didn't, and I think this was a bug).

backend/backend/settings/dev.py
jmorel: 👍

backend/backend/settings/dev.py
jmorel: 👍

backend/substrapp/views/utils.py
samlesu: Thanks for the explanation, yes my main concern was regarding the difference between `ExpiryTokenAuthentication` vs `TokenAuthentication`. As you said, it looks like a previous bug. LGTM.

jmorel:
@Kelvin-M @samlesu I implemented the change and also added a bit of that removes the auth urls used by the browsable api in prod mode.

Issue #185, opened 2020-03-18
Fix swagger doc


jmorel:
The current auto-generated swagger documentation (`/doc`) is broken, probably as a result of our last DRF update.

The [library it relied on](https://marcgibbons.com/django-rest-swagger/) is also deprecated since mid-2019, so instead of just fixing it we'll need to switch to something new.

A potential replacement is https://github.com/axnsan12/drf-yasg

We need to make sure that the fix swagger doc specifies correctly how file uploads are handled. That was the purpose of the `SchemaGenerator` class in `backend/views.py`.

Kelvin-M:
Close as swagger has been removed : https://github.com/SubstraFoundation/substra-backend/pull/364
Feel free to re-open another issue to add api documentation if needed !

PR #196, opened 2020-03-30
Remove customized swagger documentation

jmorel:
The swagger documentation was broken and the library it relied on is deprecated (see #185 )


backend/requirements.txt
```diff
@@ -1,5 +1,4 @@
 Django==2.2.10
-PyYAML==5.3
```
Kelvin-M: PyYaml should not be removed as it is used, for instance, in `docker/start.py`

jmorel: This `start.py` is not part of the platform per se, it's more of an utils to help launch the platform locally. The requirements.txt is supposed the described the requirements of the platform itself since it is in the dockerfiles.

I think it's ok to remove it here but we should also create another requirements file for local uses (`start.py`, `populate.py` etc.)

jmorel:
@Kelvin-M @samlesu I just rebased the branch, could you re-review it?

PR #197, opened 2020-03-30
Use the secrets lib for secret generation

jmorel:
Our secrets were generated using the `random` lib, which is not suitable for cryptographic purposes. One suggestion was to use `os.urandom` instead, but I discovered the `secrets` library from the standard library: https://docs.python.org/3.6/library/secrets.html

As the `secrets` introduction states:
> The secrets module is used for generating cryptographically strong random numbers suitable for managing data such as passwords, account authentication, security tokens, and related secrets.

> In particularly, secrets should be used in preference to the default pseudo-random number generator in the random module, which is designed for modelling and simulation, not security or cryptography.

I opted for secrets instead of urandom as it made for a very simple use.

### DIFF
--- a/backend/libs/gen_secret_key.py
+++ b/backend/libs/gen_secret_key.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-from random import choice
+from secrets import choice
 
 
 def gen_secret_key(r):


backend/libs/gen_secret_key.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-from random import choice
+from secrets import choice
 
 
 def gen_secret_key(r):
```
samlesu: Minor: as this PR aims at increasing the security of our application, what about replacing choice(...) with [secrets.token_urlsafe](https://docs.python.org/3/library/secrets.html#secrets.token_urlsafe)? It seems safer as more characters are used to define the secret.

jmorel: I tried to be as minimal as possible, but sure, I could use this other method.

One question though : is it important that the secret is exactly 50 chars long or can it be any size (must be long enough of course). Since token_urlsafe takes a number of bytes as input, it's just to know if I need to write `token_urlsafe(50)[:50]` or just `token_urlsafe(40)` (doc says it's about 1.3 chars per byte).

samlesu: I'm not sure it's worth it changing this function.As you said, it's good to add as minimal change as possible. We could see later how to improve this code.

samlesu: Sorry for the noise :)

PR #201, opened 2020-03-31
Secure http settings

jmorel:
This PR enables 3 settings:
- Adds the `X-Content-Type-Options: nosniff ` header to all requests
- Makes cookies secure and http only 
- Enable in prod only the  HTTP Strict Transport Security (HSTS) header with a value set through an environment variable

@AlexandrePicosson I'm tagging you so that you can set the value where appropriate. The value is a number of seconds (31536000s in a year).


samlesu:
Is there a reason for deactivating by default the HSTS header? If possible I would prefer the opposite behaviour: activate it by default with the suggested value.

jmorel:
@samlesu I think I did it this way following a discussion with @AlexandrePicosson. Alexandre what do you think?

AlexandrePicosson:
Are we sure the django backend is the right place to send this header ?
I don't have the ressources to setup a full production environment but the actual way the platform works is the following : 
```
+----------+   https  +------------------+   http    +----------+
|          | -------> |                  | --------> |          |
|  Client  |   https  |  Ingress         |   http    |  Django  |
|          | <------- | (SSL termination)| <-------- |          | 
+----------+          +------------------+           +----------+
```
The documentation of the Django middleware you use says the following: 

> SecurityMiddleware will set this header for you on all HTTPS responses if you set the SECURE_HSTS_SECONDS setting to a non-zero integer value.

But the backend is working exclusively over HTTP so it may never set the header as this header can't be sent over http or it gets ignored by the browsers/clients.

I think the correct place to set this header may be in the Ingress configuration, or you may have to setup additional variables like mentioned in the [Django doc](https://docs.djangoproject.com/en/2.2/ref/settings/#secure-proxy-ssl-header).
By default the nginx ingress controller should add  the `x-forwarded-proto` header to the forwarded requests when it consume and terminate SSL so you can use this to ensure the request was sent using HTTPS.

Also according to the [documentation](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/) of the nginx ingress controller HSTS is already enabled by default so it was not an issue if you where using this ingress controller but may be with some other controllers.

Regarding @samlesu question, yes the default value should probably be a year.

jmorel:
Thanks @AlexandrePicosson, I updated the value default to 1 year and left it in place.

I realized that putting the session and csrf cookie in secure mode even in dev broke the browsable api, so I moved these settings to prod instead.

PR #202, opened 2020-03-31
Use substra.Client new login method

jmorel:
Companion PR to https://github.com/SubstraFoundation/substra/pull/138


Kelvin-M:
@jmorel Where are we on this one ? It seems that companion PRs need to be rebased :)

PR #208, opened 2020-04-06
Fix the fact that the requests module does not use the system CA bundle

AlexandrePicosson:
Without this env variable the requests python module will not use the private CA as by default this module use the root CA bundle provided by the module [certifi](https://pypi.org/project/certifi/) and not the OS trusted root CA bundle.


AlexandrePicosson:
Rebased after the merge of https://github.com/SubstraFoundation/substra-backend/pull/6

PR #212, opened 2020-04-10
Timeit

inalgnu:


backend/backend/settings/common.py
```diff
@@ -215,4 +215,11 @@ def to_bool(value):
 
 GZIP_MODELS = to_bool(os.environ.get('GZIP_MODELS', False))
 
+STATSD_ENABLED = to_bool(os.environ.get('STATSD_ENABLED', False))
+STATSD_HOST = os.environ.get('STATSD_HOST', 'localhost')
+STATSD_PORT = os.environ.get('STATSD_PORT', 8125)
```
Kelvin-M, line 220: `os.environ` return string, so you should cast in int if this what you need :)

backend/substrapp/ledger_utils.py
```diff
@@ -271,12 +273,20 @@ def _call_ledger(call_type, fcn, args=None, kwargs=None):
         except json.decoder.JSONDecodeError:
             raise LedgerInvalidResponse(response)
 
+        # Set the ledger inside timer as a metric
+        if '__metrics__' in response:
+            metrics = response.pop('__metrics__', None)
```
Kelvin-M, line 278: Why do you use `pop` here and not `get` ?

backend/substrapp/ledger_utils.py
```diff
@@ -271,12 +273,20 @@ def _call_ledger(call_type, fcn, args=None, kwargs=None):
...
+            duration = metrics.get('duration', None)
```
Kelvin-M, line 279: Will raise if metrics is `None`

skaffold.yaml
```diff
@@ -145,6 +145,14 @@ deploy:
             - name: "node-1"
               secret: "p@$swr0d44"
           extraEnv:
+            - name: STATSD_ENABLED
+              value: "false"
+            - name: STATSD_HOST
+              value: "graphite-0.graphite.monitoring.svc.cluster.local"
```
Kelvin-M, line 151: Host value with `svc.cluster.local` is strange here, how do you deploy statsd ?

backend/substrapp/ledger_utils.py
```diff
@@ -7,10 +7,12 @@
 from django.conf import settings
 from rest_framework import status
 from grpc import RpcError
+from substrapp.metrics import get_metrics_client
 
 
 LEDGER = getattr(settings, 'LEDGER', None)
 logger = logging.getLogger(__name__)
+metrics_client = get_metrics_client()
```
Kelvin-M, line 15: Why do you use an instance class for a decorator ?
It should be better to have a decorator with class method, no ?

backend/substrapp/metrics.py
```diff
@@ -0,0 +1,62 @@
...
+        def passthrough(f):
+            @wraps(f)
+            def _wrapped(*fargs, **fkwargs):
+                return f(*fargs, **fkwargs)
+
+            return _wrapped
+        return passthrough
+
+
+_statsd_client = None
```
Kelvin-M, line 32: I think it's better to use a Class with class methods to do this stuff no ?

backend/substrapp/ledger_utils.py
inalgnu: Good point, this was done by @thibaultrobert, T think the goal is only to get the metrics and remove them from the response as the information won't be used elsewhere, we can confirm tomorrow with Thibault

skaffold.yaml
inalgnu: using another private repo

backend/substrapp/ledger_utils.py
inalgnu: @Kelvin-M this is how the statsd lib is exposing the decoration and we inherited from this mechanism

backend/substrapp/metrics.py
inalgnu: Sorry I don't get the point here, could you provide more information please

backend/backend/settings/common.py
inalgnu: done

backend/substrapp/ledger_utils.py
thibaultrobert: It's exactly that yes. The `__metrics__` field here is not relevant outside the ledger call that's why I poped it. Is there a problem with this approach ?

backend/substrapp/ledger_utils.py
thibaultrobert: You're right about that. In the current chaincode implementation, I think it can't be and present and `None` but it's still an unsafe code.

PR #215, opened 2020-04-15
Improve deployment

Kelvin-M:
Add namespace in storage metadata (missing).
Add status check deadline in skaffold file: deadline for deployements stabilisation with `skaffold run` or `skaffold deploy`

### DIFF
--- a/charts/substra-backend/templates/storage.yaml
+++ b/charts/substra-backend/templates/storage.yaml
@@ -4,6 +4,7 @@ kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
   name: {{ template "substra.fullname" $ }}-{{ .name }}
+  namespace: {{ $.Release.Namespace }}
 spec:
   storageClassName: ""
   accessModes:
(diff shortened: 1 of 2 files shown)


charts/substra-backend/templates/storage.yaml
```diff
@@ -23,6 +24,7 @@ apiVersion: v1
 kind: PersistentVolume
 metadata:
   name: {{ template "substra.fullname" $ }}-{{ .name }}
+  namespace: {{ $.Release.Namespace }}
```
ClementGautier, line 27: `PersistentVolume` unlike pvcs aren't namespaced objects.

Kelvin-M: Okay !

Kelvin-M:
> If you add namespaces at one place I would recommend adding it everywhere it's needed. As I understand you are trying to work around this issue of helm ([helm/helm#5465](https://github.com/helm/helm/issues/5465)) that make the rendering of `helm template` not adding namespaces where it should compared to helm install / helm upgrade commands am I right? Then you should add it to all deployments / comfigmaps / secrets etc. so it's consistent. Only PVs are not namespaced in my knowledge so feel free to add it everywhere else.

Yes, sort of. I try also to make pvc/pv relation deterministic in case of multiple deployement on the same cluster. If we should add to all deployments/configmaps/secrets, the scope of this PR will be too big. Close it as it is not a priority.

PR #218, opened 2020-04-15
Add pod affinity to backend worker

AlexandrePicosson:
This way we can force the worker to be scheduled on the same node as the backend in a multi node setup.
This is required in a multi node environment because the backend and worker needs to share files using the host filesystem.

This PR has not been tested against a local environment.

### DIFF
--- a/charts/substra-backend/Chart.yaml
+++ b/charts/substra-backend/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: substra-backend
 home: https://substra.org/
-version: 1.0.0-alpha.27
+version: 1.0.0-alpha.28
 description: Main package for Substra
 icon: https://avatars1.githubusercontent.com/u/38098422?s=200&v=4
 sources:
--- a/charts/substra-backend/templates/deployment-worker.yaml
+++ b/charts/substra-backend/templates/deployment-worker.yaml
@@ -20,6 +20,16 @@ spec:
         app.kubernetes.io/name: {{ template "substra.name" . }}-worker
         app.kubernetes.io/instance: {{ .Release.Name }}
     spec:
+      affinity:
+        podAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+              - key: app.kubernetes.io/name
+                operator: In
+                values:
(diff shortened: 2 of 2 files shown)


Kelvin-M:
>  RequiredDuringScheduling affinity is not symmetric. That is, if the pods of S1 have a RequiredDuringScheduling affinity rule "run me on nodes that are running pods from S2," it is not required that there be S1 pods on a node in order to schedule a S2 pod onto that node. More specifically, if S1 has the aforementioned RequiredDuringScheduling affinity rule, then:

>if a node is empty, you can schedule S2 onto the node
if a node is empty, you cannot schedule S1 onto the node
if a node is running S2, you can schedule S1 onto the node
if a node is running S1+S2 and S1 terminates, S2 continues running
if a node is running S1+S2 and S2 terminates, the system terminates S1 (eventually)

What does `eventually` mean in this case ? That is S2 is re-scheduled on the same node no termination of S1, but if not S1 will terminate ?

I see lot of potential issues :

If `server` terminates, then, eventually but let assume it will, `worker` terminates:
- As we use PV `HostPath`, if `server` and `worker` are not scheduled on the same node we won't have access to the data anymore !
- Any computation on the worker will not be marked as failed (but same issue if the `worker` terminates itself alone, so it's a known issue)

AlexandrePicosson:
The use of eventually here is a bit strange here, as I see it it is because you have two options for `RequiredDuringScheduling`: `RequiredDuringExecution` and `IgnoredDuringExecution` so eventually you activated the option `RequiredDuringExecution` and the pod will get terminated ?

In this [documentation](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#node-affinity) they say the following :

> In the future we plan to offer requiredDuringSchedulingRequiredDuringExecution which will be just like requiredDuringSchedulingIgnoredDuringExecution except that it will evict pods from nodes that cease to satisfy the pods’ node affinity requirements.

Also the document you mention is just the design-proposal eventually could mean "if they can implement it".

AurelienGasser:
The way I understand "eventually" is "if the pod terminates for another reason (killed by user, reached completion etc)

In other words, k8s won't "kill" the pod, but if the pod dies for any reason, it won't try to reschedule it.

But that's just my interpretation, the documentation is definitely not clear!

Issue #223, opened 2020-04-20
Local setup (skaffold): Incorrect DNS assumptions


AurelienGasser:
The default "local" installation of the susbtra stack (`skaffold run`) assume that custom `/etc/hosts` entries present on the host are used during DNS resolution _from inside running pods_. 

That appears to be the case in some setups:
- Docker-for-mac
- minikube with virtualbox driver

However, it appears not to be the case in other setups: 
- some AWS setups (https://github.com/SubstraFoundation/substra/issues/150)
- minikube with docker driver (https://github.com/kubernetes/minikube/issues/7802)

So, effectively, the default local installation of the substra stack only works in some environments.

PR #224, opened 2020-04-20
Local setup (skaffold): don't assume host DNS propagation

AurelienGasser:
Fixes https://github.com/SubstraFoundation/substra-backend/issues/223

Note: the prefix `http://` is necessary because of the url validation on the chaincode side ([example](https://github.com/SubstraFoundation/substra-chaincode/blob/4744da151530e8ef4b3175690eefd37db5d9b315/chaincode/input.go#L36))
﻿

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -40,7 +40,7 @@ deploy:
             fabricConfigmap: network-org-1-hlf-k8s-fabric
           backend:
             settings: dev
-            defaultDomain: http://substra-backend.node-1.com
+            defaultDomain: http://backend-org-1-substra-backend-server.org-1:8000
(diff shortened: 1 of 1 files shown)


AurelienGasser:
Tested with `substra-tests` on minikube (docker driver): :heavy_check_mark:

jmorel:
If we want to go this route, we'll have to update:
* the doc examples
* populate.py
* generate_assets.py
* substra-tests
etc.

The `.node-X.com` are hardcoded in a lot of places, we'll need to replace them everywhere.

jmorel:
It looks I was mistaken about what this does, sorry about that.

PR #226, opened 2020-04-23
add pre save logs if exceptions happend

inalgnu:


backend/substrapp/tests/query/tests_query_datasample.py
```diff
@@ -486,7 +486,7 @@ def test_bulk_update_data(self):
                                                   description=self.data_description2,
                                                   data_opener=self.data_data_opener2)
 
-        d = DataSample(path=self.data_file)
+        d = DataSample(path=self.data_file_filename)
```
Kelvin-M, line 489: Why not using https://github.com/SubstraFoundation/substra-backend/blob/fc247da9557b69f241e6cfae653983e1d09f4fd3/backend/substrapp/tests/query/tests_query_datasample.py#L199:L201

backend/substrapp/signals/datasample/pre_save.py
```diff
@@ -1,18 +1,26 @@
...
 
     # try to make an hard link to keep a free copy of the data
     # if not possible, keep the real path location
     try:
-        shutil.copytree(normpath(instance.path), directory, copy_function=link)
-    except Exception:
-        shutil.rmtree(directory, ignore_errors=True)
+        shutil.copytree(src_path, destination_path, copy_function=link)
+    except Exception as e:
+        logger.error(f'error happend while copying data from {src_path} to {destination_path} {e}')
```
Kelvin-M, line 21: It may be better to use `logger.exception` to log the exception :)

backend/substrapp/tests/tests_tasks.py
```diff
@@ -220,7 +220,7 @@ def test_prepare_data_sample_zip(self):
 
     def test_prepare_data_sample_zip_fail(self):
 
-        data_sample = DataSample(pkhash='foo', path=self.data_sample)
+        data_sample = DataSample(pkhash='foo', path=self.data_sample_filename)
```
Kelvin-M, line 223: Why not using https://github.com/SubstraFoundation/substra-backend/blob/fc247da9557b69f241e6cfae653983e1d09f4fd3/backend/substrapp/tests/query/tests_query_datasample.py#L199:L201

backend/substrapp/signals/datasample/pre_save.py
inalgnu: done

backend/substrapp/tests/tests_tasks.py
inalgnu: could you give me more details? a filename here seems to be sufficient

Kelvin-M: A file name is a string, but not a path. It work as it will not raise, but it can be better to use a path to the file (even fake) to be consistent.

PR #228, opened 2020-04-24
Allow receiving data as json in multipart requests

jmorel:
Companion PR: https://github.com/SubstraFoundation/substra/pull/162

# Current situation

Until now, when sending files alongside data in a request, the data was encoded as form data. This prevented us from sending complex data structures, and for example forced us to send these permissions:
```json
{
  "public": True,
  "authorized_ids": []
}
```

as:
```
permissions_public = True,
permissions_authorized_ids = True
```

And then forced us to parse the upload to recreate the dict on the server side before passing it to the serializer.

# Proposed solution

In order to send files and data in one call, we can serialize all data in json format and pass it as the value of a "json" formdata. 

A new request parser is then written that inherits from the DRF's multipart parser that automatically handles the deserialization of the "json" formdata.

This change is almost transparent for the rest of the codebase, with 2 exceptions:
 * some views were expecting request.data to be a querydict and used request.data.getlist. Since request.data is now a dict, we have to use get() and provide a default list.
* serializers defined their optional values (such as tag) with `requested=False` (when the tag key is not present in the request) and with `allow_empty=True` (when tag is an empty string). They didn't need to have `allow_null=True` because null values were automagically stripped. Now they need to.

The biggest impact was for the tests as they needed to be updated so that they'd use this new json serialization.

### DIFF
--- a/backend/libs/json_multipart_parser.py
+++ b/backend/libs/json_multipart_parser.py
@@ -1,0 +1,21 @@
+from rest_framework.parsers import MultiPartParser, DataAndFiles
+from rest_framework.settings import api_settings
+from rest_framework.utils import json
+
+
+class JsonMultiPartParser(MultiPartParser):
+    media_type = 'multipart/form-data'
+    strict = api_settings.STRICT_JSON
+
+    def _load_json(self, data):
+        if 'json' not in data:
+            return {}
+
+        parse_constant = json.strict_constant if self.strict else None
+        return json.loads(data['json'], parse_constant=parse_constant)
+
+    def parse(self, stream, media_type, parser_context):
+        data_and_files = super().parse(stream, media_type, parser_context)
+        files = data_and_files.files.dict()
+        data = self._load_json(data_and_files.data)
+        return DataAndFiles(data, files)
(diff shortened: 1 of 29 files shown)


backend/libs/json_multipart_parser.py
```diff
@@ -0,0 +1,23 @@
+
+from django.http import QueryDict
+from rest_framework.parsers import MultiPartParser, DataAndFiles
+from rest_framework.settings import api_settings
+from rest_framework.utils import json
+
+
+class JsonMultiPartParser(MultiPartParser):
+    media_type = 'multipart/form-data'
+    strict = api_settings.STRICT_JSON
```
samlesu, line 10: It seems unused, we could remove it?

jmorel: It is used, on line 16.

jmorel: There was an unused import though, I removed it.

samlesu: Yes sorry, my comment was not very precise.

What's the goal of adding this option? It looks like that we need to support only one of the case. If yes, I think we could remove the option.

It's also hard to know what's the default value of this option (it should not be added to the settings/common.py file?).

jmorel: I went to DRF's JSONParser for inspiration and found this option that I wasn't aware of. I included it here as well so that no matter what the project's settings are, the behavior of json and multipart-json would be the same.

samlesu: Thanks for the explanation. Makes sense.

PR #233, opened 2020-04-30
Kubernetes tasks

AlexandrePicosson:
The goal of this PR is to enable the usage of Kubernetes to run tasks instead of using the docker host.

In order to achieve this goal we will have to go through the following steps:
- [x] Build images using kaniko
- [x] Add cache warmer for kaniko
- [ ] Run images using kubernetes API
- [ ] Improve security of the docker registry

The way this is expected to work is first building the archive corresponding to the tuple if it is not already present in the registry and push the task Dockerfile to the private docker registry.

Then we run the tasks pulling from the internal registry.

For this I made available new endpoints in the skaffold.yaml file that needs to be added to the user `/ect/hosts` file. 
- substra-registry.node-1.com
- substra-registry.node-2.com

This is necessary for the host docker being able to pull from these images registry.
It is also necessary to add these endpoints to the host docker config as trusted insecure images registry.

### DIFF
--- a/backend/substrapp/tasks/utils.py
+++ b/backend/substrapp/tasks/utils.py
@@ -4,6 +4,7 @@ import docker
 import logging
 import functools
 import threading
+import requests
 
 from subprocess import check_output
 from django.conf import settings
@@ -328,9 +329,8 @@ def build_docker_image(dockerfile_path, image_name):
     dockerfile_fullpath = os.path.join(dockerfile_path, 'Dockerfile')
     current_namespace = os.getenv('NAMESPACE')
     docker_cache = os.getenv('DOCKER_CACHE_PVC')
-
     config.load_incluster_config()
-
+    dockerfile_mount_path = os.path.join(os.path.realpath(os.getenv('MEDIA_ROOT')), 'subtuple')
     batch_v1 = client.BatchV1Api()
 
     container = client.V1Container(
@@ -344,7 +344,7 @@ def build_docker_image(dockerfile_path, image_name):
             "--insecure"
         ],
         volume_mounts=[
-            {'name': 'dockerfile', 'mountPath': f"{os.getenv('MEDIA_ROOT')}subtuple", 'readOnly': True},
+            {'name': 'dockerfile', 'mountPath': dockerfile_mount_path, 'readOnly': True},
             {'name': 'cache', 'mountPath': '/cache', 'readOnly': True}
         ]
     )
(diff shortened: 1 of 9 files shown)


backend/substrapp/tasks/utils.py
```diff
@@ -220,6 +220,87 @@ def list_files(startpath):
...
+        image="gcr.io/kaniko-project/executor:latest",
+        args=[
+            f"--dockerfile={dockerfile_fullpath}",
+            f"--context=dir://{dockerfile_path}",
+            f"--destination={os.getenv('REGISTRY')}:5000/{image_name}",
+            "--cache=true",
+            "--insecure"
+        ],
+        volume_mounts=[
+            {'name': 'dockerfile', 'mountPath': f"{os.getenv('MEDIA_ROOT')}subtuple", 'readOnly': True},
```
Kelvin-M, line 243: It may be better to build the mountPath with os.path.join
Moreover you can certified that MEDIA_ROOT has a '/' at the end with 
```python
# Get real path and ensure there is a suffix / at the end of the path
os.path.join(os.path.realpath(to_directory), '')
```

backend/substrapp/tasks/utils.py
```diff
@@ -220,6 +220,87 @@ def list_files(startpath):
         logger.info(f'{startpath} does not exist.')
 
 
+def build_docker_image(dockerfile_path, image_name):
+    dockerfile_fullpath = os.path.join(dockerfile_path, 'Dockerfile')
+    current_namespace = os.getenv('NAMESPACE')
```
Kelvin-M, line 225: You could get the namespac from the kubernetes python client :)

backend/substrapp/tasks/utils.py
```diff
@@ -220,6 +220,87 @@ def list_files(startpath):
...
+    docker_cache = os.getenv('DOCKER_CACHE_PVC')
```
Kelvin-M, line 226: It may be better to fetch it with kubernetes python client :)

backend/substrapp/tasks/utils.py
```diff
@@ -220,6 +220,87 @@ def list_files(startpath):
...
+        kind="Job",
+        metadata=client.V1ObjectMeta(name="kaniko"),
+        spec=spec
+    )
+
+    batch_v1.create_namespaced_job(body=job, namespace=current_namespace)
+
+    status = batch_v1.list_namespaced_job(current_namespace)
+
+    kanikoJob = None
```
Kelvin-M, line 277: We often use snake case for python variable instead of camel case (used for class)

backend/substrapp/tasks/utils.py
```diff
@@ -220,6 +220,87 @@ def list_files(startpath):
...
+
+    for job in status.items:
```
Kelvin-M, line 279: You can use a `filter` or list comprehension for that :)

backend/substrapp/tasks/utils.py
```diff
@@ -220,6 +220,87 @@ def list_files(startpath):
...
+        if job.metadata.name == "kaniko":
+            kanikoJob = job
+            break
+
+    while kanikoJob.status.succeeded is None:
+        if kanikoJob.status.failed == 4:
```
Kelvin-M, line 285: 4 is not explicit, it may be better to use a global variable or fetch this value from kubernetes client if exists !

backend/substrapp/tasks/utils.py
```diff
@@ -220,6 +220,87 @@ def list_files(startpath):
...
+            break
+        time.sleep(10)
+        status = batch_v1.list_namespaced_job(current_namespace)
+        for job in status.items:
```
Kelvin-M, line 289: Same as before !

backend/substrapp/tasks/utils.py
```diff
@@ -220,6 +220,87 @@ def list_files(startpath):
...
+            if job.metadata.name == "kaniko":
+                kanikoJob = job
+                break
+
+    batch_v1.delete_namespaced_job(
```
Kelvin-M, line 294: You should put that in a finally I think, to be sure to remove the job :)

backend/substrapp/tasks/utils.py
AlexandrePicosson: The only solution I found to do this does not work with the `config.load_incluster_config()` as we only get a token to interact with the API and not a full kubeconfig with context informations like the namespace.
As the env workaround was already used [here](https://github.com/SubstraFoundation/substra-backend/blob/master/backend/substrapp/tasks/tasks.py#L757) I figured that I should use it too.

backend/substrapp/tasks/utils.py
AlexandrePicosson: You would need its name in order to fetch it, or rely on something like `kubernetes.client.list_pvc()` with a filter matching a label `docker-cache`, I don't know what is the best solution for this.

backend/substrapp/tasks/utils.py
Kelvin-M: Yes, you're right, maybe rename this variable to make it explicit that this namespace will be use for that ! :)

backend/substrapp/tasks/utils.py
Kelvin-M: I would prefer this solution as it is less error-prone than a env variable

backend/substrapp/tasks/utils.py
```diff
@@ -272,41 +272,94 @@ def build_docker_image(dockerfile_path, image_name):
...
+            logger.info(f'kubernetes.images.build - elaps={elaps:.2f}ms')
 
 
 def compute_docker(client, resources_manager, dockerfile_path, image_name, container_name, volumes, command,
                    environment, remove_image=True, remove_container=True, capture_logs=True):
 
-    dockerfile_fullpath = os.path.join(dockerfile_path, 'Dockerfile')
-    if not os.path.exists(dockerfile_fullpath):
-        raise Exception(f'Dockerfile does not exist : {dockerfile_fullpath}')
+    get_dockerfile_fullpath(dockerfile_path)
```
Kelvin-M, line 360: should be `dockerfile_fullpath = get_dockerfile_fullpath(dockerfile_path)`

AlexandrePicosson: dockerfile_fullpath is unused later in the code leading to a lint error, it was just used to verify that the path exists early in the function.

AurelienGasser:
> This is necessary for the host docker being able to pull from these images registry.

This would resurrect the problem stated in [this issue](https://github.com/SubstraFoundation/substra-backend/issues/223), wouldn't it? 

If I understand correctly, the proposed change would make susbtra-backend non-functional in environments where the user's `/etc/hosts` isn't used within the cluster (e.g. minikube with docker driver). 

It would also make it non functional in environments where the kubernetes host's `/etc/host` isn't writable (users testing substra using a kubernetes cluster provisioned in the cloud because their local machine isn't powerful enough)

AlexandrePicosson:
Yes it would resurrect this issue.

However there is no way to work around this, the host container runtime need to resolve the registry hostname in order to download and start images when Kubernetes asks for it.

But as this would only be an optional way to run containers (you will still be able to use the old docker task runner which would be the default option) I thinks this is not an issue for people trying out the software.
It is more of an advanced feature for specific deployments where you have constraints that require you to use only kubernetes and not access the host docker socket as this is a potential security issue.

A potential workaround in the cloud would be to setup a local DNS for your hosts and add entries for the backend and the registry but I recognise that it is quite advanced.

Kelvin-M:
Can we use `svc.cluster.local` instead ?

AlexandrePicosson:
No as the hostname of the registry needs to be resolved by the host docker daemon which is unaware of kubernetes internal DNS.

Kelvin-M:
> No as the hostname of the registry needs to be resolved by the host docker daemon which is unaware of kubernetes internal DNS.

You're right !

AlexandrePicosson:
Closing this PR as this work is now done in https://github.com/SubstraFoundation/substra-backend/pull/249.

Issue #217, opened 2020-04-15
Worker: have a proper error if the medias volume is not mounted


samlesu:
If the medias volume is not mounted properly on the worker the tuple execution fails with the following message:

```
ERROR 2020-04-15 13:43:08,556 substrapp.tasks.tasks 708 139767323916096 [00-01-0126-1e30844]
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/celery/app/trace.py", line 385, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/celery/app/trace.py", line 650, in __protected_call__
    return self.run(*args, **kwargs)
  File "/usr/src/app/substrapp/tasks/tasks.py", line 550, in compute_task
    max_retries=int(getattr(settings, 'CELERY_TASK_MAX_RETRIES')))
  File "/usr/local/lib/python3.6/dist-packages/celery/app/task.py", line 704, in retry
    raise_with_context(exc)
  File "/usr/src/app/substrapp/tasks/tasks.py", line 543, in compute_task
    prepare_materials(subtuple, tuple_type)
  File "/usr/src/app/substrapp/tasks/utils.py", line 26, in timed
    result = function(*args, **kw)
  File "/usr/src/app/substrapp/tasks/tasks.py", line 580, in prepare_materials
    prepare_opener(directory, subtuple)
  File "/usr/src/app/substrapp/tasks/utils.py", line 26, in timed
    result = function(*args, **kw)
  File "/usr/src/app/substrapp/tasks/tasks.py", line 333, in prepare_opener
    raise Exception('DataOpener Hash in Subtuple is not the same as in local db')
Exception: DataOpener Hash in Subtuple is not the same as in local db
```

It doesn't help to find the cause of the error. This is ambiguous and does not help for debugging.

The function `get_hash` must be improved.

Issue #222, opened 2020-04-20
Algo fetch: HTTP errors aren't displayed in the logs


AurelienGasser:
HTTP errors occurring during the algo fetch aren't reported in the logs. This makes it hard to understand what is going on. 

For instance, [in this issue](https://github.com/SubstraFoundation/substra/issues/150), the logs only error show `hash doesn't match [...]` which is unhelpful: the real source of the problem is the HTTP 403 error which occurred prior to the hash computation.

samlesu:
It looks like it is similar to the issue #217. The cause of the failure is different, but it's the same ambiguous error message.

AurelienGasser:
After more investigation, the case I was looking at was not a HTTP 403 error. Despite the appearances, the server returns a 200.

![image](https://user-images.githubusercontent.com/168382/81079298-66083b00-8ebd-11ea-9757-13465c121ae4.png)

For context, this is what the worker pod receives when it tries to fetch e.g. `susbtra-backend.node-1.com/algo/....` but its DNS resolution points to the public internet, and not the k8s cluster.

Maybe another approach would be to look at the response content type. Thoughts @Kelvin-M

AurelienGasser:
I'm going to close this for now as there's no easy way to fix it.

We should keep in mind that errors like `hash doesn't match ...` are usually due to DNS misconfiguration. The solution is make sure that the address `substra-backend.node-{1-N}.com` resolves to the cluster IP, and not to the public internet. This could be part of a troubleshooting guide? cc @natct10

natct10:
> I'm going to close this for now as there's no easy way to fix it.
> 
> We should keep in mind that errors like `hash doesn't match ...` are usually due to DNS misconfiguration. The solution is make sure that the address `substra-backend.node-{1-N}.com` resolves to the cluster IP, and not to the public internet. This could be part of a troubleshooting guide? cc @natct10

You're right! I will add this to the troubleshooting section, thank you for pointing this one out @AurelienGasser

PR #237, opened 2020-05-04
Improve error handling

Kelvin-M:
Proposal Fix https://github.com/SubstraFoundation/substra-backend/issues/217

Some improvement to a better handling of those issues

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -338,9 +338,9 @@ def prepare_opener(directory, tuple_):
 
     datamanager = DataManager.objects.get(pk=data_opener_hash)
 
-    # verify that local db opener file exists
+    # verify that local storage opener file exists
     if not os.path.exists(datamanager.data_opener.path) or not os.path.isfile(datamanager.data_opener.path):
-        raise Exception(f'DataOpener file ({datamanager.data_opener.path}) is missing in local db ')
+        raise Exception(f'DataOpener file ({datamanager.data_opener.path}) is missing in local storage')
 
     # verify that local db opener file is not corrupted
     if get_hash(datamanager.data_opener.path) != data_opener_hash:
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -338,6 +338,10 @@ def prepare_opener(directory, tuple_):
 
     datamanager = DataManager.objects.get(pk=data_opener_hash)
 
+    # verify that local db opener file exists
+    if not os.path.exists(datamanager.data_opener.path) or not os.path.isfile(datamanager.data_opener.path):
+        raise Exception(f'DataOpener file ({datamanager.data_opener.path}) is missing in local db ')
+
     # verify that local db opener file is not corrupted
     if get_hash(datamanager.data_opener.path) != data_opener_hash:
         raise Exception('DataOpener Hash in Subtuple is not the same as in local db')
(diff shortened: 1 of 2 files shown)


backend/substrapp/tests/common.py
```diff
@@ -293,6 +293,10 @@ def __init__(self, status, content):
         self.status_code = status
         self.content = content
 
+    def raise_for_status(self):
+        if self.status_code != 200:
+            raise Exception(f'HTTP Exception : {self.status_code}')
+
```
AurelienGasser, line 299: I might be missing something, but doesn't that apply only to tests? I'm not seeing anything checking the response  code of actual requests.

Kelvin-M: Old code ! I will remove it

backend/substrapp/tasks/tasks.py
```diff
@@ -357,6 +361,13 @@ def prepare_data_sample(directory, tuple_):
     from substrapp.models import DataSample
     for data_sample_key in tuple_['dataset']['keys']:
         data_sample = DataSample.objects.get(pk=data_sample_key)
+
+        if not os.path.exists(data_sample.path) or not os.path.isdir(data_sample.path):
+            raise Exception(f'Data Sample ({data_sample.path}) is missing in local db')
+
+        if not os.listdir(data_sample.path):
+            raise Exception(f'Data Sample ({data_sample.path}) is empty in local db')
+
```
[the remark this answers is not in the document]
Kelvin-M: Not but the error here is more explicit I think :) (better twice than none)

AurelienGasser:
@Kelvin-M This doesn't seem to fix https://github.com/SubstraFoundation/substra-backend/issues/222 as the PR description implies

backend/substrapp/tasks/tasks.py
```diff
@@ -338,6 +338,10 @@ def prepare_opener(directory, tuple_):
 
     datamanager = DataManager.objects.get(pk=data_opener_hash)
 
+    # verify that local db opener file exists
+    if not os.path.exists(datamanager.data_opener.path) or not os.path.isfile(datamanager.data_opener.path):
+        raise Exception(f'DataOpener file ({datamanager.data_opener.path}) is missing in local db ')
```
AurelienGasser, line 343: I find the `in local db` part to be confusing: why "local db" and not "local filesystem"? 

I'm not sure what "local db" refers to. I would assume that the local db is postgres.

Kelvin-M:
> @Kelvin-M This doesn't seem to fix #222 as the PR description implies

I don't think this issue was caused by a 403
As you can see here https://github.com/SubstraFoundation/substra-backend/blob/master/backend/substrapp/utils.py#L242:L266 , we handle http issues by raising a Node Error
Do you have a setup to reproduce it ?

backend/substrapp/tasks/tasks.py
Kelvin-M: I will change it to local storage

AurelienGasser:
@KelvinM

- Re: issue #222. You're right, it's not caused by a 403. I updated the issue. Could you please remove it from this PR's description?
- Re: naming: You updated one place but not the two others.

Kelvin-M:
> @Kelvinm
> 
> * Re: issue #222. You're right, it's not caused by a 403. I updated the issue. Could you please remove it from this PR's description?
> * Re: naming: You updated one place but not the two others.

Done

Kelvin-M:
@samlesu Can you review if this solution if good for your issue (#217)

Kelvin-M:
> Thanks @Kelvin-M. This will help debugging when the "data volumes" are not available.
> 
> Using dedicated exceptions would have been better in my opinion. As it would be inconsistant with the existing code base, that's good for me :)

Yes, I think there is a potential improvement by reworking how exception are generated and raised in the backend.

PR #240, opened 2020-05-11
Update api token strategy

Kelvin-M:
- Create a new token for each login (remove the old one)
- Token validity is still 24h

### DIFF
--- a/backend/backend/views.py
+++ b/backend/backend/views.py
@@ -18,12 +18,9 @@ class ExpiryObtainAuthToken(ObtainAuthToken):
         serializer.is_valid(raise_exception=True)
         user = serializer.validated_data['user']
 
-        token, created = Token.objects.get_or_create(user=user)
-
         # token should be new each time, remove the old one
-        if not created:
-            token.delete()
-            token = Token.objects.create(user=token.user)
+        Token.objects.filter(user=user).delete()
+        token = Token.objects.create(user=user)
 
         return Response({
             'token': token.key,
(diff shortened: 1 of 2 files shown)


backend/substrapp/tests/views/test_views_authentication.py
```diff
@@ -102,11 +102,44 @@ def test_obtain_token(self):
...
+        token_old = response.json()['token']
+        self.assertTrue(token_old)
+
+        # token should be update after a second post
         response = self.client.post('/api-token-auth/',
                                     {'username': 'foo', 'password': 'bar'}, **self.extra)
         self.assertEqual(response.status_code, 200)
         token = response.json()['token']
-        self.assertTrue(token)
+        self.assertTrue(token_old)
```
jmorel, line 116: I think this should still be `self.assertTrue(token)`

Kelvin-M: Yes !

backend/backend/views.py
```diff
@@ -17,10 +17,13 @@ def post(self, request, *args, **kwargs):
...
         user = serializer.validated_data['user']
+
         token, created = Token.objects.get_or_create(user=user)
 
-        # token_expire_handler will check, if the token is expired it will generate new one
-        is_expired, token = token_expire_handler(token)
+        # token should be new each time, remove the old one
+        if not created:
+            token.delete()
+            token = Token.objects.create(user=token.user)
```
samlesu, line 26: What about trying to delete it if it exists and then create it in all cases? 
It seems easier to maintain: creation is done at a single place.

```
# token should be new each time, so remove old one if it exists
Token.objects.filter(user=user).delete()
token = Token.objects.create(user=user)
```
(this code has not been tested :))

Kelvin-M:
@jmorel Can you review it when you have time :) ?

PR #243, opened 2020-05-12
Objective unicity

### LINKED ISSUE #221: Objective unicity check
The unicity of an objective is currently only based on the description content. This allows objective with the same metrics to be uploaded multiple time so that they are associated with different datasets.

This however prevents the creation of an objective with the same description but a different metrics archive.

We could have the unicity check include both description and metrics, which would allow the same flexibility when it comes to re-uploading a metrics with another description but would also allow to upload multiple metrics with the same description.

Kelvin-M:
Fix #221


samlesu:
Thanks @Kelvin-M. LGTM

Kelvin-M:
@jmorel  I'm closing it, feel free to take the code when you want to do this update :)

PR #249, opened 2020-05-19
K8s tasks

Kelvin-M:
Inherit from https://github.com/SubstraFoundation/substra-backend/pull/233 
Companion PR : https://github.com/SubstraFoundation/substra-tests/pull/109
The goal of this PR is to enable the usage of Kubernetes to run tasks instead of using the docker host.

In order to achieve this goal we will have to go through the following steps:
- [x] Option to choose between docker host or k8s compute backend
- [x] Build images using kaniko
- [x] Add cache warmer for kaniko
- [x] Prepare k8s volume for k8s backend
- [x] Run images using kubernetes API
- [x] Fetch results from k8s pods (model, pred, perf)
- [x] Switch from nginx file exposure to RWO pods (worker - job)
- [x] Local volume for compute plan
- [x] Value for registry address (https://github.com/SubstraFoundation/substra-backend/pull/251#pullrequestreview-418896390)
- [x] Memory Limit, Cpu Limit, GPU allocation
- [x] Service Account for job/pod creation (Need @AurelienGasser validation)
- [x] Improve security of the docker registry (Need @AurelienGasser validation)
- [x] Secure ML Task job (remains : Disable Network)
- [ ] Refactoring & Optimization

The way this is expected to work is first building the archive corresponding to the tuple if it is not already present in the registry and push the task Dockerfile to the private docker registry.

Then we run the tasks pulling from the internal registry.

For NetworkPolicy you need to install calico :
```
kubectl apply -f https://docs.projectcalico.org/manifests/calico.yaml
```
~~For this I made available new endpoints in the skaffold.yaml file that needs to be added to the user /ect/hosts file.~~

~~substra-registry.node-1.com~~
~~substra-registry.node-2.com~~
~~This is necessary for the host docker being able to pull from these images registry.~~
~~It is also necessary to add these endpoints to the host docker config as trusted insecure images registry.~~
```
  "insecure-registries": [
    "substra-registry.node-1.com",
    "substra-registry.node-2.com"
  ]
```

### DIFF
--- a/charts/substra-backend/templates/deployment-worker.yaml
+++ b/charts/substra-backend/templates/deployment-worker.yaml
@@ -82,8 +82,12 @@ spec:
             - name: REQUESTS_CA_BUNDLE
               value: /etc/ssl/certs/ca-certificates.crt
             {{- end }}
-            - name: SUBTUPLE_PVC
-              value: {{ include "substra.fullname" $ }}-subtuple
+            {{- range .Values.persistence.volumes }}
+            - name: {{ .name | upper }}_PVC
+              value: {{ include "substra.fullname" $ }}-{{ .name }}
+            {{- end }}
+            - name: SERVERMEDIAS_PVC
+              value: {{ template "substra.fullname" $ }}-servermedias
             - name: REGISTRY
               value: {{ .Release.Name }}-docker-registry
             - name: NAMESPACE
--- a/charts/substra-backend/templates/deployment-worker.yaml
+++ b/charts/substra-backend/templates/deployment-worker.yaml
@@ -74,7 +74,7 @@ spec:
             - name: LEDGER_CONFIG_FILE
               value: /conf/{{ .Values.organization.name }}/substra-backend/conf.json
             - name: MEDIA_ROOT
-              value: {{ .Values.persistence.hostPath }}/medias/
+              value: /var/substra/medias/
             - name: PYTHONUNBUFFERED
               value: "1"
             - name: "CELERYWORKER_IMAGE"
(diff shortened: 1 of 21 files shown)


charts/substra-backend/templates/rbac.yaml
```diff
@@ -25,7 +25,13 @@ rules:
...
     verbs: ["get", "watch", "list", "create", "update", "patch"]
----
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+  - apiGroups: ["batch", "extensions"]
+    resources: ["jobs"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+  ---
 kind: RoleBinding
```
Kelvin-M, line 35: @AurelienGasser Can you confirm if this rbac is enough ?

AurelienGasser: @Kelvin-M I get this in the worker. The wrong service account is getting used apparently
```
│ HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"jobs.batch  │
│ is forbidden: User \"system:serviceaccount:org-1:default\" cannot create resource \"jobs\" in API group \"batc │
│ h\" in the namespace \"org-1\"","reason":"Forbidden","details":{"group":"batch","kind":"jobs"},"code":403}
```

charts/substra-backend/templates/deployment-worker.yaml
```diff
@@ -112,8 +112,10 @@ spec:
 {{ toYaml . | indent 12 }}
           {{- end }}
           volumeMounts:
+            {{- if .Values.backend.dockerSocket.enabled }}
             - name: dockersocket
               mountPath: /var/run/docker.sock
```
inalgnu, line 117: seems that it should be updated also here {{ .Values.backend.dockerSocket.hostPath }}

Kelvin-M: This variable is here to know where the docker socket is on the host.
As docker socket on pods should be in /var/run/docker.sock there is no issue here.

inalgnu:
closing this PR in favor of #267

PR #257, opened 2020-06-01
Add prefix functionality for S3 bucket upload

AurelienGasser:
Currently, generated tar.gz files can only be uploaded at the root of an s3 bucket.

This PR adds support for `.tar.gz` upload into s3 prefixes, which are the s3 equivalent of subfolders.


AurelienGasser:
Released as `0.0.17-alpha.1`

AurelienGasser:
Released as `0.0.17-alpha.2`

PR #258, opened 2020-06-03
add export volume

inalgnu:


Kelvin-M:
You need to create the export folder here :) 
https://github.com/SubstraFoundation/substra-backend/blob/master/backend/substrapp/tasks/tasks.py#L395

PR #259, opened 2020-06-03
Disable CORS_ALLOW_CREDENTIALS by default

Kelvin-M:
Disable CORS_ALLOW_CREDENTIALS
But for front-end you need to set it to true to allow it to login in the backend !


Kelvin-M:
@jmorel  Yes, I thought I've done that ! Fixed now !

PR #264, opened 2020-06-16
don't use hostPath for PVC

AurelienGasser:
Changes:

- Make `hostPath` value optional + introduce `storageClassName` value
- Only `chown` data folders when hostPath is used. I can't see a benefit to chown in non-hostPath scenarios
- Use `/var/substra` instead of `/substra` for the mount path (see note below)

--- 

Note: I chose `/var/substra` because  1. /var is a more standard place to store data, and 2.  there are 100+ occurrences of the string /substra in the project, which makes future refactors painful. 

:warning:  If we merge this we'll need to update `/substra` to `/var/substra` in `script-cleanup.sh` (melloddy-deploy)

### DIFF
--- a/charts/substra-backend/templates/deployment-server.yaml
+++ b/charts/substra-backend/templates/deployment-server.yaml
@@ -71,7 +71,7 @@ spec:
           - name: LEDGER_CONFIG_FILE
             value: /conf/{{ .Values.organization.name }}/substra-backend/conf.json
           - name: MEDIA_ROOT
-            value: {{ .Values.persistence.hostPath }}/medias/
+            value: /var/substra/medias/
           - name: PYTHONUNBUFFERED
             value: "1"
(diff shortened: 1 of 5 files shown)


AurelienGasser:
Tested with one end-to-end test locally

```
pytest tests/test_execution.py -k same_node
```

- with hostPath enabled: :heavy_check_mark: 
- with hostPath disabled: :heavy_check_mark:

Kelvin-M:
Nice work, but it should be done in k8s-tasks too no ?

AlexandrePicosson:
I think you forgot to bump the chart version

AurelienGasser:
committed into `hlf-k8s` tasks manually: https://github.com/SubstraFoundation/substra-backend/commit/d851c92d3651b56dd19cbbc1c6c1d266daae40b6

PR #271, opened 2020-06-18
ecr + kaniko wip

AurelienGasser:
Maybe a terrible idea anyway


AurelienGasser:
Doesn't work yet.

I think we need to do this : https://docs.docker.com/registry/spec/auth/token/

AurelienGasser:
Closing in favor of https://github.com/SubstraFoundation/substra-backend/pull/278

PR #273, opened 2020-06-22
Add metadata to compute plan

maeldebon:
:link: **Related Pull Requests :**
• [substra](https://github.com/SubstraFoundation/substra/pull/189)
• [substra-tests](https://github.com/SubstraFoundation/substra-tests/pull/112)
• [substra-chaincode](https://github.com/SubstraFoundation/substra-chaincode/pull/121)

This PR aims to add a metadata field to the compute plan asset.

### DIFF
--- a/backend/substrapp/serializers/ledger/computeplan/serializer.py
+++ b/backend/substrapp/serializers/ledger/computeplan/serializer.py
@@ -78,11 +78,10 @@ class LedgerComputePlanSerializer(serializers.Serializer):
                 'dataSampleKeys': data_traintuple['train_data_sample_keys'],
                 'algoKey': data_traintuple['algo_key'],
                 'id': data_traintuple['traintuple_id'],
+                'metadata': data_traintuple.get('metadata'),
             }
             if 'in_models_ids' in data_traintuple:
                 traintuple['inModelsIDs'] = data_traintuple['in_models_ids']
-            if 'metadata' in data_traintuple:
-                traintuple['metadata'] = data_traintuple['metadata']
             if 'tag' in data_traintuple:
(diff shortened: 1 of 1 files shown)


backend/substrapp/serializers/ledger/computeplan/serializer.py
```diff
@@ -75,6 +81,8 @@ def get_args(self, data):
             }
             if 'in_models_ids' in data_traintuple:
                 traintuple['inModelsIDs'] = data_traintuple['in_models_ids']
+            if 'metadata' in data_traintuple:
+                traintuple['metadata'] = data_traintuple['metadata']
```
jmorel, line 85: Is there a reason to only include metadata if present vs always including it like we do for the other assets (datasets, objectives etc.) ?

backend/substrapp/tests/views/tests_views_algo.py
```diff
@@ -218,6 +222,7 @@ def test_algo_create(self):
             mcreate.return_value = {}
             response = self.client.post(url, data=data, format='multipart', **self.extra)
 
+        assert response.data['metadata'] is not None
```
jmorel, line 225: Why not test that it is equal to the actual value ?

maeldebon: I just checked, we cannot get the metadata's value by using the response object. I'll remove this test for now.

backend/substrapp/serializers/ledger/computeplan/serializer.py
maeldebon: Well, it doesn't work when I change this, so I suppose it's the right way to do it.

PR #274, opened 2020-06-22
Use RUN chown instead of COPY --chown

Kelvin-M:
Using `COPY --chown` does not seem to work with gcloud build (CI with substra-tests).


Kelvin-M:
> LGTM.
> 
> Side note: another approach might have been to actually create user 1001?

@AurelienGasser What is the procedure for that on gcp for instance ?

AurelienGasser:
@Kelvin-M Just `RUN useradd` !

Kelvin-M:
Ah I thought you were talking about adding it in gcp for the CI :D

PR #275, opened 2020-06-22
Update builder setup

Kelvin-M:


backend/substrapp/tasks/k8s_backend.py
```diff
@@ -263,19 +263,24 @@ def k8s_build_image(path, tag, rm):
...
         command = None
         mount_path_dockerfile = '/makisu-context'
-        mount_path_cache = None
+        mount_path_cache = '/makisu-storage'
         args = [
             'build',
             f'--push={REGISTRY}',
             f'-t={tag}:substra',
             '--modifyfs=true',
+            f'--storage="{mount_path_cache}"'
```
AurelienGasser, line 283: Nit: I don't think you need this line

```
 --storage string                  Directory that makisu uses for temp files and cached layers. Mount this path for better caching performance. If modifyfs is set, default to /makisu-storage; Otherwise default to /tmp/makisu-storage
```

Kelvin-M: I prefer to do it explicitly as if we change modifyfs, makisu storage will be in /tmp/makisu-storage

PR #276, opened 2020-06-23
prepopulate local registry

AurelienGasser:
Pre-populate the local registry from an arbitrary source registry (with auth)

---

I intentionally separated it from the existing cache warmer, because it serves a different purpose : 

- the cache warmer **populates the cache volume** (builder-dependent)
- the registry prepopulate **populates the docker registry** (builder-independent)

### DIFF
--- a/backend/substrapp/tasks/k8s_backend.py
+++ b/backend/substrapp/tasks/k8s_backend.py
@@ -254,7 +254,7 @@ def k8s_build_image(path, tag, rm):
     if IMAGE_BUILDER == 'kaniko':
         # kaniko build can be launched without privilege but
         # it needs some capabilities and to be root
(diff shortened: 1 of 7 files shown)


backend/substrapp/tasks/k8s_backend.py
```diff
@@ -24,7 +24,7 @@
 FS_GROUP = os.getenv('FS_GROUP')
 IMAGE_BUILDER = os.getenv('IMAGE_BUILDER')
 KANIKO_IMAGE = os.getenv('KANIKO_IMAGE')
-KANIKO_MIRROR = os.getenv('KANIKO_MIRROR')
+KANIKO_MIRROR = os.getenv('KANIKO_MIRROR').lower() == 'true'
```
inalgnu, line 27: there is a to_bool helper in the settings common.py file that you can use to convert boolean config values into a python True/False value and then treat it as a bool everywhere

AurelienGasser:
closed in favor of https://github.com/SubstraFoundation/substra-backend/pull/278

PR #279, opened 2020-06-26
Concurrency k8s

Kelvin-M:
Companion PRs:

- https://github.com/SubstraFoundation/hlf-k8s/pull/87


charts/substra-backend/templates/job-image-builder-warmer.yaml
```diff
@@ -130,20 +161,52 @@ spec:
...
           - "/makisu-context"
         volumeMounts:
         - name: makisu-cache
           mountPath: /makisu-storage
+          subPath: warm-{{ $index }}
           readOnly: False
         - name: dockerfile
           mountPath: /makisu-context
           readOnly: true
+      - name: copy-to-cache
```
AurelienGasser, line 180: Note: this won’t  work if we preload mutiple images. 

To fix it: Instead of adding multiple jobs, add multiple “copy-to-cache” init containers run sequentially, all pointing to the same cache folder (no subpath)

AurelienGasser:
rebased `melloddy`

charts/substra-backend/templates/job-image-builder-warmer.yaml
Kelvin-M: Is it still an issue ?

AurelienGasser:
I think we should go back to an exponential backoff, with a reasonable maximum delay of, say, 5 to 30 minutes.

The linear backoff only retries for a few minutes, which is sometimes not enough for the application to recover its stable state after a partial outage.

Using an exponential backoff would allow us to recover in those scenarios where part of the application is down for say 5-10 minutes. Example scenario: another peer is down, but the current training task needs to download an in-model from it. In that scenario, assuming the peer comes back up within 5-10 minutes, if we retry for long enough, the training task will succeed. If not, the training task will fail, possibly failing a whole compute plan.

backend/substrapp/tasks/tasks.py
```diff
@@ -707,11 +707,11 @@ def _do_task(subtuple_directory, tuple_type, subtuple, compute_plan_id, rank, or
             subtuple_key=subtuple["key"],
             compute_plan_id=compute_plan_id,
             dockerfile_path=f'{subtuple_directory}/metrics',
-            image_name=f'substra/metrics_{subtuple["key"][0:8]}'.lower(),
+            image_name=f'substra/metrics_{subtuple["objective"]["hash"][0:8]}'.lower(),
             job_name=f'{tuple_type}-{subtuple["key"][0:8]}-eval'.lower(),
             volumes=common_volumes,
             command=f'--output-perf-path {OUTPUT_PERF_PATH}',
-            remove_image=not(settings.TASK['CACHE_DOCKER_IMAGES']),
+            remove_image=not(compute_plan_id is not None or settings.TASK['CACHE_DOCKER_IMAGES']),
```
AurelienGasser, line 714: @Kelvin-M Thoughts on this commit?

Kelvin-M: Nope :)

Kelvin-M:
> I think we should go back to an exponential backoff, with a reasonable maximum delay of, say, 5 to 30 minutes.
> 
> The linear backoff only retries for a few minutes, which is sometimes not enough for the application to recover its stable state after a partial outage.
> 
> Using an exponential backoff would allow us to recover in scenarios where part of the application is down for say 5-10 minutes.
> 
> Example scenario: another node is down, but the current training task needs to download an in-model from the backend of that node. In that scenario, if we retry for long enough, the node will come back up and the training task will succeed. If we don't wait for long enough, the training task will fail, possibly failing a whole compute plan.

Maybe an exponential backoff with `max_time_to_wait` and `max_time_total` (for instance 5 minutes max between  and 30/60 minutes total) can be a good idea

PR #280, opened 2020-07-03
Use nginx with uswgi

Kelvin-M:
We should use nginx in front of uwgsi
https://uwsgi-docs.readthedocs.io/en/latest/tutorials/Django_and_nginx.html#basic-nginx

### DIFF
--- a/charts/substra-backend/templates/configmap-server.yaml
+++ b/charts/substra-backend/templates/configmap-server.yaml
@@ -38,7 +38,6 @@ data:
     }
 
 ---
-
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -46,7 +45,6 @@ metadata:
 data:
   uwsgi.ini: |
     [uwsgi]
-    # chdir           = /usr/src/app
     module          = backend.wsgi
     env             = DJANGO_SETTINGS_MODULE=backend.settings.server.{{ .Values.backend.settings }}
     static-map      = /static=/usr/src/app/backend/statics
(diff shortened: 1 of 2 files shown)


Kelvin-M:
@ClementGautier Can you review the nginx configuration. Should I add more options ?

charts/substra-backend/templates/configmap-server.yaml
```diff
@@ -38,27 +38,61 @@ data:
...
+    worker_processes  auto;
+
+    events {
+      # worker_connections  4096;  ## Default: 1024
+    }
+    http {
+
+      # the upstream component nginx needs to connect to
+      upstream django {
+          server unix:///usr/src/app/backend/socket/substra-backend.sock; # for a file socket
```
inalgnu, line 83: This should be passed as a variable and also should be the path to statics below

charts/substra-backend/templates/deployment-server.yaml
```diff
@@ -99,6 +95,8 @@ spec:
             mountPath: /var/substra/servermedias
           - name: statics
             mountPath: /usr/src/app/backend/statics
+          - name: uwsgi-socket
+            mountPath: /usr/src/app/backend/socket
```
AurelienGasser, line 99: Minor: /usr/src is an uncommon folder for a socket, isn’t it?
 What about a more standard folder, like /var/substra ?

Kelvin-M: Yes !

charts/substra-backend/templates/configmap-server.yaml
Kelvin-M: Those paths are hardcoded in deployment-server, I don't think we should let them configurable or we need to make everything configurable (path to statics, socket and medias) which is not the aim of this PR.
I keep it like this for consistency :)

charts/substra-backend/templates/deployment-server.yaml
ClementGautier: sockets usually live in `/var/run/xxx.sock` if you are looking for something standard ;)

Kelvin-M: Yes, good point !

Kelvin-M:
> I'm not completely confident we should add a webserver here... Every configuration (like max upload size, number of open connexion, forwarding X-forwarded headers etc.) will have to be duplicated between the loadbalancer and here and that gave me headache already.
> 
> I understand we need a WSGI server to translate the http request into something python can understand but I don't see the added value here. If you think uwsgi isn't performing well why not trying gunicorn for example ?
> 
> I would stick to http if it change nothing for you.


In fact, we should not expose directly uwsgi (even gunicorn) directly. It always needs to have a webserver to server the request.

We run into something like this issue : https://medium.com/@gabriel.m.troy/aws-alb-docker-uwsgi-502-bad-gateway-16d0a36f6240
 
https://uwsgi-docs.readthedocs.io/en/latest/tutorials/Django_and_nginx.html
> Now normally we won’t have the browser speaking directly to uWSGI. That’s a job for the webserver, which will act as a go-between.

But here https://github.com/nginxinc/kubernetes-ingress/issues/143, it says that we do not need a webserver between ingress and nginx by adding more configuration in uwsgi.

ClementGautier:
You don't expose uwsgi directly, the Nginx ingress controller is the webserver in our case.
My point is the same as this: https://github.com/nginxinc/kubernetes-ingress/issues/143#issuecomment-347814243

If you still need to add a proxy Nginx, could you test that configuration put at ingress level isn't overwritten by this one:
```
# nginx controller chart values
config:
  server-tokens: 'false'
  use-forwarded-headers: 'true'
  compute-full-forwarded-for: 'true'
extraArgs:
  enable-ssl-passthrough: ''

# ingress annotations
nginx.ingress.kubernetes.io/client-body-buffer-size: 100m
nginx.ingress.kubernetes.io/proxy-body-size: 100m
```

Kelvin-M:
@ClementGautier I will try to update the uwsgi settings along the github issue if it solves the 502 bad gateway.

Kelvin-M:
We will try to handle it with ingress and uwsgi (http-socket directly)

PR #285, opened 2020-07-10
Fix subtuple check when replaying events

Kelvin-M:


AurelienGasser:
@Kelvin-M @inalgnu This appears to have broken the e2e tests. For some reason, the tuple is _sometimes_ not found by `find_training_step_tuple_from_key`.

I fixed it in `multi-channel` by [wrapping it](https://github.com/SubstraFoundation/substra-backend/blob/90ca03f34d685160796e9bae6330b4b85a9111f6/backend/substrapp/tasks/tasks.py#L490-L500) in a `try/except/pass`, but there's probably a better way to fix it.

Kelvin-M:
It may "break" e2e tests when launching them concurrently. Indeed, I think, when multiple jobs call the backend-server (and consequently invoke the ledger), blocks are not already received. And we the backend-worker launches a train/test after receiving an event, it may fail as the subtuple does not exist at this moment.
We can conclude that this tuple is new, and ignore the exception as it surely a 'todo' tuple

Kelvin-M:
@AurelienGasser Can you add it as a fix for master branch ? :)

backend/substrapp/ledger_utils.py
```diff
@@ -147,7 +147,7 @@ def _wrapper(*args, **kwargs):
                     return fn(*args, **kwargs)
                 except exceptions_to_retry as e:
                     _nbtries -= 1
-                    if not nbtries:
+                    if not _nbtries:
```
samlesu, line 150: 🐛

PR #289, opened 2020-07-15
Ignore error on subtuple retrieval in prepare_tuple

Kelvin-M:
See discussion [here](https://github.com/SubstraFoundation/substra-backend/pull/285)

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -487,10 +487,15 @@ def prepare_tuple(subtuple, tuple_type):
     # Can happen if we re-process all events
     # But we need to fetch the subtuple again to get the last
     # version of it
-    _, subtuple_check = find_training_step_tuple_from_key(subtuple['key'])
-    if subtuple_check['status'] != 'todo':
-        logger.error(f'Tuple task ({tuple_type}) not in "todo" state ({subtuple_check["status"]}).\n{subtuple_check}')
-        return
+    try:
+        _, subtuple_check = find_training_step_tuple_from_key(subtuple['key'])
+        if subtuple_check['status'] != 'todo':
+            logger.error(f'Tuple task ({tuple_type}) not in "todo" state ({subtuple_check["status"]}).'
+                         f'\n{subtuple_check}')
+            return
+    except Exception:
+        # use the provided subtuple if the previous call fail
+        pass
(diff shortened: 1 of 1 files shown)


backend/substrapp/tasks/tasks.py
```diff
@@ -487,10 +487,15 @@ def prepare_tuple(subtuple, tuple_type):
...
-        return
+    try:
+        _, subtuple_check = find_training_step_tuple_from_key(subtuple['key'])
+        if subtuple_check['status'] != 'todo':
+            logger.error(f'Tuple task ({tuple_type}) not in "todo" state ({subtuple_check["status"]}).'
+                         f'\n{subtuple_check}')
+            return
+    except Exception:
+        # use the provided subtuple if the previous call fail
+        pass
```
Esadruhn, line 498: does the provided subtuple have a status ? Or is it not reliable to check the `subtuple['status'] == 'todo'` ?

Kelvin-M: Subtuple provide a status. It's not always reliable in case of backend-server restart and replaying events from the begining. Indeed, a replayed event gives the worker a subtuple from an old state of the ledger. This can be seen as a sanity check for replayed event. New events do not need this sanity check :)

backend/substrapp/tasks/tasks.py
```diff
@@ -487,10 +487,15 @@ def prepare_tuple(subtuple, tuple_type):
...
-    if subtuple_check['status'] != 'todo':
-        logger.error(f'Tuple task ({tuple_type}) not in "todo" state ({subtuple_check["status"]}).\n{subtuple_check}')
-        return
+    try:
+        _, subtuple_check = find_training_step_tuple_from_key(subtuple['key'])
+        if subtuple_check['status'] != 'todo':
+            logger.error(f'Tuple task ({tuple_type}) not in "todo" state ({subtuple_check["status"]}).'
+                         f'\n{subtuple_check}')
+            return
+    except Exception:
```
inalgnu, line 496: shouldn't we catch only `TasksError` that could be raised by `find_training_step_tuple_from_key`? I mean no catch-all exception?

Kelvin-M: Yes, I'm currently testing it !

PR #290, opened 2020-07-15
Fix crash when trying to save identical trunk/head models

AurelienGasser:
For a composite traintuple, if the head and the trunk out models are identical, the saving fails.
﻿
That's because not only do the two models have the same composite traintuple key (expected), they ALSO have the same value.

This PR adds a salt to avoid pkhash conflicts in that case.

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -925,6 +925,7 @@ def save_models(subtuple_directory, tuple_type, subtuple_key):
                 subtuple_directory,
                 subtuple_key,
                 filename=filename,
+                extraKey=type_model
             )
 
             models[type_model] = {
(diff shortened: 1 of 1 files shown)


AurelienGasser:
Actually, this approach would be a breaking change. 

1. The pkhash calculation algorithm also needs to be updated in `get_and_put_local_model_content`

https://github.com/SubstraFoundation/substra-backend/blob/19ec017d26f5c58d50e8e8300b2b961e41b94fd3/backend/substrapp/tasks/tasks.py#L158-L160

2. Any existing deployment with composite traintuples in the ledger would lead to potential crashes

`get_and_put_local_model_content` expects the pkhash algorithm to be constant. (see code above)

With the proposed change, the new pkhash algorithm (with the salt) would be different from the old algorithm (without the salt). 

So assets saved in the ledger with the old algorithm would be not be usable on a backend using the new algorithm.

Kelvin-M:
I think it was a good way to solve this issue ! It is difficult here to keep compatibility. We can do that by using old and new pkhash strategy if we need to keep retro-compatibility

samlesu:
I also think this a good solution !
Correct me if I'm wrong, but AFAIK a change in the pkhash generation is not a blocking point.

Kelvin-M:
> I also think this a good solution !
> Correct me if I'm wrong, but AFAIK a change in the pkhash generation is not a blocking point.

Depends if we support retro-compatibility

AurelienGasser:
@Kelvin-M 

> We can do that by using old and new pkhash strategy if we need to keep retro-compatibility

Great idea, that would do it. We would need a timeline for deprecation (a future version when we'll stop accepting the old pkhash), but we can keep supporting both for the time being.

Issue #291, opened 2020-07-15
[Edge case] Crash when composite traintuple head and trunk models are identifcal


AurelienGasser:
For a composite traintuple, if the head and the trunk out models are identical, the saving fails.
﻿
That's because not only do the two models have the same composite traintuple key (expected), they ALSO have the same value.

The hash is computed from the traintuplekey and the value. So the hashes are the same for the head and the trunk model, which leads to a pkhash conflict.

There was [an aborted attempt at fixing this issue](https://github.com/SubstraFoundation/substra-backend/pull/290).

Esadruhn:
Also once the fix is done in the default mode we can fix this in the debug mode (on the substra repo).

PR #298, opened 2020-07-27
(2nd attempt) Fix crash when trying to save identical trunk/head models

AurelienGasser:
New iteration of PR https://github.com/SubstraFoundation/substra-backend/pull/290

Fixes https://github.com/SubstraFoundation/substra-backend/issues/291

--- 

For a composite traintuple, if the head and the trunk out models are identical, the saving fails.
﻿
That's because not only do the two models have the same composite traintuple key (expected), they ALSO have the same value.

This PR adds a salt to avoid pkhash conflicts in that case.


AurelienGasser:
I ran a few tests:

Train a composite traintuple using the `master` branch (old algo), then try to use it as an in-model on `fix-model-save-composite` (new algo): 
- With `SUPPORT_LEGACY_MODEL_SALT` = `False`, **fails**, as expected :heavy_check_mark: 
- With `SUPPORT_LEGACY_MODEL_SALT` = `True`, **succeeds** as expected :heavy_check_mark: 


e2e tests OK for both `False` and `True`

backend/backend/settings/common.py
```diff
@@ -216,3 +216,9 @@ def to_bool(value):
 GZIP_MODELS = to_bool(os.environ.get('GZIP_MODELS', False))
 
 ENABLE_REMOVE_LOCAL_CP_FOLDERS = to_bool(os.environ.get('ENABLE_REMOVE_LOCAL_CP_FOLDERS', True))
+
+# The algorithm used to compute out-model hashes was changed to support more use-cases.
+# The following flag ensures backward compatibility: it makes it so that hashes computed
+# using the 'old' algorithm are still treated as valid.
+# This flag should ultimately set to False, and the associated code should be removed.
+SUPPORT_LEGACY_MODEL_SALT = True
```
Kelvin-M, line 224: Maybe we can add something like `os.environ.get('SUPPORT_LEGACY_MODEL_SALT', True)` to enable or disable it with env variable ?

backend/substrapp/tasks/tasks.py
```diff
@@ -137,34 +142,43 @@ def find_training_step_tuple_from_key(tuple_key):
...
-    if get_hash(model.file.path, tuple_key) != out_model['hash']:
+    def validate_hash(path, key, expected):
+        if get_hash(model.file.path, key) == expected:
+            return True
+
+        if settings.SUPPORT_LEGACY_MODEL_SALT and key.endswith(HASH_SALT_HEAD):
+            key = key[:-len(HASH_SALT_HEAD)]
+            return get_hash(model.file.path, key) == expected
+        else:
+            return False
```
Kelvin-M, line 172: You should create `validate_hash` function outside as it will be create each time you call the function `get_and_put_local_model_content` !

backend/substrapp/tasks/tasks.py
```diff
@@ -137,34 +142,43 @@ def find_training_step_tuple_from_key(tuple_key):
...
+def get_and_put_model_content(tuple_type, hash_key, tuple_, out_model, model_dst_path):
     """Get out model content."""
     owner = tuple_get_owner(tuple_type, tuple_)
     return get_and_put_asset_content(
         out_model['storageAddress'],
         owner,
         out_model['hash'],
         content_dst_path=model_dst_path,
-        salt=tuple_key
+        hash_key=hash_key
```
Kelvin-M, line 153: I don't understand the change from `salt` to hash_key for the function arguments, can you explain :) ?

AurelienGasser: This is to be consistent with the names of the arguments of the `get_hash` function. It doesn't have a concept of "salt", only of a key, so I fixed the inconsistency.

AurelienGasser:
Backwards compatibility has been removed as it's not required. This significantly reduced complexity.

e2e: :heavy_check_mark:

backend/substrapp/tasks/tasks.py
```diff
@@ -32,6 +32,9 @@
 PREFIX_HEAD_FILENAME = 'head_'
 PREFIX_TRUNK_FILENAME = 'trunk_'
 
+HASH_SALT_HEAD = 'HEAD'
+HASH_SALT_TRUNK = 'TRUNK'
```
Kelvin-M, line 36: Following  
> This is to be consistent with the names of the arguments of the `get_hash` function. It doesn't have a concept of "salt", only of a key, so I fixed the inconsistency.

Is this still relevant to use `SALT` for those variables (and `_salt` references in the same file) ?

AurelienGasser: I think so. Since it's something that is appended to the cache key, I think calling it "salt" makes sense. Do you have another idea in mind?

Kelvin-M: `EXTENSION` or something in the same family :)

Kelvin-M:
Look like the CI is failing now
`/substrapp/tasks/tasks.py:934:121: E501 line too long (121 > 120 characters)`

backend/substrapp/tasks/tasks.py
```diff
@@ -930,15 +930,15 @@ def save_models(subtuple_directory, tuple_type, subtuple_key):
         }
 
     elif tuple_type == COMPOSITE_TRAINTUPLE_TYPE:
-        for type_model, filename, hash_salt in [('end_head_model', OUTPUT_HEAD_MODEL_FILENAME, HASH_KEY_SUFFIX_HEAD),
-                                                ('end_trunk_model', OUTPUT_TRUNK_MODEL_FILENAME, HASH_KEY_SUFFIX_TRUNK)]:
+        for type, filename, hash_salt in [('end_head_model', OUTPUT_HEAD_MODEL_FILENAME, HASH_KEY_SUFFIX_HEAD),
```
Kelvin-M, line 933: Using `type` is shadowing a builtin function (see https://docs.python.org/3/library/functions.html#type)

Nevertheless, it's not a reserved word (http://docs.python.org/reference/lexical_analysis.html#keywords ), but it's generally not considered as a good practice to shadow any builtin even if, here, it's reduced to the scope of the function.

AurelienGasser:
e2e: :heavy_check_mark:

Kelvin-M (approved):

backend/substrapp/tasks/tasks.py
```diff
@@ -939,16 +944,15 @@ def save_models(subtuple_directory, tuple_type, subtuple_key):
         }
 
     elif tuple_type == COMPOSITE_TRAINTUPLE_TYPE:
-
-        for type_model, filename in [('end_head_model', OUTPUT_HEAD_MODEL_FILENAME),
-                                     ('end_trunk_model', OUTPUT_TRUNK_MODEL_FILENAME)]:
+        for t_type, filename, hash_salt in [('end_head_model', OUTPUT_HEAD_MODEL_FILENAME, HASH_KEY_SUFFIX_HEAD),
```
inalgnu, line 947: could you explain what does `t_type` stands for and why `type_model` was renamed?

AurelienGasser: `type_model` has been renamed because of flake8 (line length).

The new name `t_type` means "tuple type" but that's not a good name. Let me rename it to `m_type` (model type) which is more accurate.

AurelienGasser: done https://github.com/SubstraFoundation/substra-backend/pull/298/commits/c11b42b54388441d941d144e542708e893e37bda

inalgnu (approved):

PR #299, opened 2020-07-28
K8s backend

Kelvin-M:
Non-mellody branch
Include:
- k8s-tasks
- security improvements
- fix concurrency

Companion PR :
https://github.com/SubstraFoundation/hlf-k8s/pull/86
https://github.com/SubstraFoundation/substra-tests/pull/126

Related : 
https://github.com/SubstraFoundation/substra-backend/pull/267
https://github.com/SubstraFoundation/substra-backend/pull/279

PR #304, opened 2020-07-31
K8S Backend fixup

Kelvin-M:
Multiple fixes after k8s-backend #299 was merged :
- Remove docker socket mount in substra-backend chart 
- Add server medias hostPath value


Kelvin-M:
Re-activate hostPath for servermedias (see conversation in https://github.com/SubstraFoundation/substra-tests/pull/126)

PR #307, opened 2020-08-17
Use pvc for docker registry data.

Kelvin-M:
Use pvc for docker registry data storage instead of emptyDir that may create issue if storage on node is greater than 80% (pod eviction by kubernetes)

### DIFF
--- a/charts/substra-backend/Chart.yaml
+++ b/charts/substra-backend/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: substra-backend
 home: https://substra.org/
-version: 1.0.0-alpha.35
+version: 1.0.0-alpha.36
 description: Main package for Substra
 icon: https://avatars1.githubusercontent.com/u/38098422?s=200&v=4
 sources:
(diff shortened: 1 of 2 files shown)


Kelvin-M:
> I'm not a big fan of enabling everything by default. By default you want Helm chart to be as simple as possible for development purpose so it can run without extra controller on k8s. What you can do is provide a values-production.yaml for exemple that describe what a production environment should look like.

Yes, but as we use pvc for all storage (you can see it in substra-backend volumes), it's more consistence and avoid potential issue with host storage.

ClementGautier:
> Yes, but as we use pvc for all storage (you can see it in substra-backend volumes), it's more consistent and avoid potential issue with host storage.

I 100% agree that pvc is to be prefered over hostpath but it's not what this PR says. This PR says "by default, this chart will know requires a persistent storage" and I kind of disagree. On the other hand having a more robust default set of values for production is gladely appreciated but maybe in another file, `values-production.yaml` is usually pick as name for that. For example : https://github.com/helm/charts/tree/94cdd6c65013dd1a54511aaf59868e0fe1a78bd6/stable/redis have two set of values.

As I'm writting this I see that lots of charts have switched to a more "robust/secure by default" approach so my comment may be a complete falacy. I'm approving this.

ClementGautier (approved):

inalgnu (approved):

PR #267, opened 2020-06-17
MELLODDY - K8S tasks, Security improvements

inalgnu:
This branch is based on #249 and adding some security improvements

Companion PR : SubstraFoundation/substra-tests#109

PR #311, opened 2020-08-27
Reuse metrics image in compute plans

AurelienGasser:
In compute plans, the metrics image is typically the same for multiple ranks.

This PR makes it so:
- the metrics image name is built from the objective hash and not the tuple hash.
- the metrics image is not deleted after use (compute plans only)

As a result, if a compute plan uses the same objective for multiple iterations, the same metrics image will be reused without the need to rebuild it for every testtuple.

-- 

Note that this code is already merged in melloddy https://github.com/SubstraFoundation/substra-backend/pull/267. The goal of this PR is to reflect the changes in master.

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -704,11 +704,11 @@ def _do_task(subtuple_directory, tuple_type, subtuple, compute_plan_id, rank, or
             subtuple_key=subtuple["key"],
             compute_plan_id=compute_plan_id,
             dockerfile_path=f'{subtuple_directory}/metrics',
-            image_name=f'substra/metrics_{subtuple["key"][0:8]}'.lower(),
+            image_name=f'substra/metrics_{subtuple["objective"]["hash"][0:8]}'.lower(),
             job_name=f'{tuple_type}-{subtuple["key"][0:8]}-eval'.lower(),
             volumes=common_volumes,
             command=f'--output-perf-path {OUTPUT_PERF_PATH}',
-            remove_image=not(settings.TASK['CACHE_DOCKER_IMAGES']),
+            remove_image=not(compute_plan_id is not None or settings.TASK['CACHE_DOCKER_IMAGES']),
             remove_container=settings.TASK['CLEAN_EXECUTION_ENVIRONMENT'],
             capture_logs=settings.TASK['CAPTURE_LOGS'],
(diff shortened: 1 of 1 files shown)


AurelienGasser:
e2e: :heavy_check_mark:

backend/substrapp/tasks/tasks.py
```diff
@@ -704,11 +704,11 @@ def _do_task(subtuple_directory, tuple_type, subtuple, compute_plan_id, rank, or
             subtuple_key=subtuple["key"],
             compute_plan_id=compute_plan_id,
             dockerfile_path=f'{subtuple_directory}/metrics',
-            image_name=f'substra/metrics_{subtuple["key"][0:8]}'.lower(),
+            image_name=f'substra/metrics_{subtuple["objective"]["hash"][0:8]}'.lower(),
             job_name=f'{tuple_type}-{subtuple["key"][0:8]}-eval'.lower(),
             volumes=common_volumes,
             command=f'--output-perf-path {OUTPUT_PERF_PATH}',
-            remove_image=not(settings.TASK['CACHE_DOCKER_IMAGES']),
+            remove_image=not(compute_plan_id is not None or settings.TASK['CACHE_DOCKER_IMAGES']),
```
Kelvin-M, line 711: This boolean formula can be simplified to 
```python
compute_plan_id is None and not settings.TASK['CACHE_DOCKER_IMAGES']
```
if we compute the truth table and/or use boolean algebra :
|compute_plan_id is not None - [1]  | CACHE_DOCKER_IMAGES - [2]  | 1 or 2  - [3] | not(3) | not(1) and not (2) | 
|---|---|---|---|---|
| False  |  False  | True  | False  |   False  |
|  True | False  | False  | True  |  True |
| False  | True  | True  | False  |   False  |
| True  | True  |  True | False  |  False   |

with 
```python
not (compute_plan_id is not None) = compute_plan_id is None
```

inalgnu (approved):

backend/substrapp/tasks/tasks.py
AurelienGasser: done https://github.com/SubstraFoundation/substra-backend/pull/311/commits/314d9eb2a8a5705e8d82c1d53f94bf9a1c631aaa

Kelvin-M (approved):

PR #312, opened 2020-09-03
Return key instead of pkhash

Kelvin-M:
The goal of this multi-repo PRs is to make sure that all the asset creation endpoints returns only the "key" of the asset. 
Indeed, currently `create` and 'get' an asset return, respectively, `pkhash` and `key`. As `pkhash` is only a substra-backend concept, we need to move from `pkhash` to `key` when dealing with substra cli/sdk and frontend.

Consequently, when client create a new asset, the substra-backend will now only return the `key` of the created asset.

There is an exception for compute plan which need to be returned in its complete form because client need to get the ID_to_key mapping and it’s returned only when client create this asset.

The client will take care of doing an extra GET in order to retrieve the complete asset.

PRs :
    https://github.com/SubstraFoundation/substra-backend/pull/312
    https://github.com/SubstraFoundation/substra/pull/220
    https://github.com/SubstraFoundation/substra-tests/pull/140

### DIFF
--- a/backend/substrapp/views/utils.py
+++ b/backend/substrapp/views/utils.py
@@ -217,3 +217,15 @@ def get_channel_name(request):
         return request.headers['Substra-Channel-Name']
 
     raise exceptions.BadRequestError('Could not determine channel name')
+
+
+def data_to_data_response(data):
+    # Transform data to a data_response with only key
+
+    if data and isinstance(data, list) and 'pkhash' in data[0]:
+        return [{'key': d['pkhash']} for d in data]
+
+    if 'pkhash' in data:
+        return {'key': data['pkhash']}
+
+    return data
--- a/backend/substrapp/tests/assets.py
+++ b/backend/substrapp/tests/assets.py
@@ -14,45 +14,24 @@ In order to update this file:
 
 objective = [
     {
-        "key": "1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3",
-        "name": "Skin Lesion Classification Objective",
+        "key": "48cc5ce4d352b859e202adb0d639ee092c78cae7a2a1df0fccd88c63b3466ad8",
+        "name": "9224cad96c964b7b8ed2f7eb5872c27a_global - Objective 0",
         "description": {
-            "hash": "1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3",
-            "storageAddress": "http://testserver/objective/1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3/description/"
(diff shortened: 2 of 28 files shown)


backend/substrapp/views/utils.py
```diff
@@ -217,3 +217,15 @@ def get_channel_name(request):
         return request.headers['Substra-Channel-Name']
 
     raise exceptions.BadRequestError('Could not determine channel name')
+
+
+def data_to_data_response(data):
```
inalgnu, line 222: It looks like we want to call this before every response.
Wouldn't be more appropriate to use middleware in that particular case?

Advantages :
 - No need to modify all the views
 - Once the chaincode is adopting this change (replace pkhash with key) we only need to remove the middleware.

What do you think?

Kelvin-M: We use this function for `create` asset call responses as we merge information from django and chaincode response.
I'm not familiar with middleware for this kind of stuff (overwrite one route response) and I don't know if it can be used for this kind of problem.
The pkhash concept is currently linked with use the django psql db. I don't know if it will be easy to migrate from it to something else :)
In addition, there are lot of similar code parts in multiple views in the backend that surely need to be refactor to something simpler.

Esadruhn:
When running the tests from substra-tests locally, I get an error (`test_debug.py::test_debug_compute_plan_aggregate_composite`):

```
>   ???
E   pydantic.error_wrappers.ValidationError: 1 validation error for CompositeAlgo
E   pkhash
E     extra fields not permitted (type=value_error.extra)

pydantic/main.py:346: ValidationError
```

it says that the composite algo returned by the backend contains an extra field 'pkhash'. I should be on the right branch for substra-backend, substra and substra-tests

Kelvin-M:
Yes, as those tests use a container with a image without the modification of substra sdk/cli

backend/substrapp/views/utils.py
AurelienGasser: I have another view on this. 

I think the backend should own the return values of its interfaces. Instead, right now the backend proxies the chaincode response (and this PR adds a modification to this proxying).

I think a better solution would be to remove the `data_to_data_response` function entirely, and let each interface **explicitely** compute and return its return value. For instance:

```python
class TrainTupleViewSet(mixins.CreateModelMixin,
   [...]
   data = self._create(request) // from the chaincode
   response = { 'key': data['pkhash'] }
   return Response(response)
```

IMO this would be better because it's more explicit.

Kelvin-M: @AurelienGasser Yep, it could be a better approach but looks like a bigger work than solving the issue of this PR ^^

Esadruhn (approved):
not a lot of experience with the backend but looks good to me and works for our purpose :)

inalgnu (approved):

PR #314, opened 2020-09-03
Improve worker logs

AurelienGasser:
This is a first pass at improving worker logs.

Example without error:

image

Example with error:

image
image
image


AurelienGasser:
Here's a first draft, it needs many improvements but a few things are here:

- prepend each log entry with the task ID (useful when tasks running in parallel)
- log directory contents as a single json line (takes less space)
- log tuple key when starting "prepare materials"

backend/libs/formatters.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def format(self, record):
+        task = self.get_current_task()
+        if task and task.request:
+            record.__dict__.update(task_id=task.request.id,
+                                   task_name=task.name)
+        else:
+            record.__dict__.setdefault('task_name', '')
+            record.__dict__.setdefault('task_id', '')
+        return super().format(record)
```
AurelienGasser, line 23: Stolen from https://www.distributedpython.com/2018/11/06/celery-task-logger-format/

backend/substrapp/tasks/k8s_backend.py
```diff
@@ -651,7 +658,15 @@ def k8s_remove_local_volume(volume_id):
         spec=spec
     )
 
-    k8s_client.create_namespaced_pod(body=pod, namespace=NAMESPACE)
+    try:
+        k8s_client.create_namespaced_pod(body=pod, namespace=NAMESPACE)
+    except kubernetes.client.rest.ApiException as e:
+        if e.status == 409 and '"reason":"AlreadyExists"' in e.body:
+            logger.info(f'Cannot create pod {NAMESPACE}/{job_name}: Already exists. Skipping local volume deletion.')
+            return
```
AurelienGasser, line 666: @inalgnu @Kelvin-M somehow the `clean-xxx` pod sometimes already exists. If that's the case, I suggest to just skip the whole function. Agree?

inalgnu: @AurelienGasser yes agree, the pod name depends on the volume_id which depends on the compute plan id. For the same compute plan this could be skipped.

AurelienGasser:
e2e :heavy_check_mark:

backend/substrapp/ledger_utils.py
```diff
@@ -338,7 +338,10 @@ def call_ledger(channel_name, call_type, fcn, *args, **kwargs):
         # add a log even if the function raises an exception
         te = time.time()
         elaps = (te - ts) * 1000
-        logger.info(f'smartcontract {call_type}:{fcn}; elaps={elaps:.2f}ms; error={error}')
+        if error is None:
+            logger.info(f"(smartcontract) {call_type}:{fcn} took {elaps:.2f} ms")
+        else:
+            logger.info(f"(smartcontract) {call_type}:{fcn} took {elaps:.2f} ms. Error: {error}")
```
inalgnu, line 344: shouldn't use logger.error here?

backend/substrapp/tasks/k8s_backend.py
```diff
@@ -90,7 +90,14 @@ def watch_pod(name, watch_init_container=False):
...
     watch_container = not watch_init_container
 
-    logger.error(f'Checking for pod {name} init: {watch_init_container} main: {watch_container}')
+    logger.info(f'Waiting for pod {name}...')
+
+    def get_error(state):
+        error = state.reason
+        if state.message is not None:
+            error += f' ({state.message})'
+        return error
```
Kelvin-M, line 99: I think this function can be created outside this one :)

backend/substrapp/ledger_utils.py
AurelienGasser: I'd argue that this log line is for informational purposes only, and that since the error is re-raised, it's the caller's responsibility to report the error as such. 

If we put `logger.error` here, then the logs will have **two** big red "ERROR" lines, suggesting that two different errors occurred. 

@inalgnu @Kelvin-M thoughts?

backend/substrapp/tasks/k8s_backend.py
AurelienGasser: https://github.com/SubstraFoundation/substra-backend/pull/314/commits/5c3a741ead9333de4d44a89c3a3bba0e9daec5ad

Kelvin-M (approved):

backend/substrapp/ledger_utils.py
inalgnu: I think it's better to differentiate logs that contain errors, especially if some exceptions are completely silent this would help to see that errors are happening even if it duplicates the information. 

However, I don't have a strong opinion on that so do whatever helps better-investigating errors.

inalgnu (approved):

PR #315, opened 2020-09-09
Switch output from camel case to snake case

Kelvin-M:
The backend responses should be more consistent. The goal of this PR is to make sure that all inputs fields and response fields of backend endpoints are in snake_case format.

PRs:
https://github.com/SubstraFoundation/substra-backend/pull/315
https://github.com/SubstraFoundation/substra/pull/222
https://github.com/SubstraFoundation/substra-tests/pull/142

### DIFF
--- a/backend/substrapp/ledger_utils.py
+++ b/backend/substrapp/ledger_utils.py
@@ -3,6 +3,7 @@ import functools
 import json
 import logging
 import time
+import re
 
 from django.conf import settings
 from rest_framework import status
@@ -101,6 +102,9 @@ _STATUS_TO_EXCEPTION = {
     status.HTTP_409_CONFLICT: LedgerConflict,
 }
 
+CAMEL_TO_SNAKE_PATTERN = re.compile(r'(.)([A-Z][a-z]+)')
+CAMEL_TO_SNAKE_PATTERN_2 = re.compile(r'([a-z0-9])([A-Z])')
+
 
 def _raise_for_status(response):
     """Parse ledger response and raise exceptions in case of errors."""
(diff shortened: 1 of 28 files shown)


backend/substrapp/ledger_utils.py
```diff
@@ -157,6 +161,44 @@ def _wrapper(*args, **kwargs):
...
+    new_d = {}
+    for key, value in d.items():
+        if isinstance(value, dict):
+            value = _replace_dict_keys(value, converter)
+        elif isinstance(value, list):
+            if all([isinstance(v, dict) for v in value]):
+                value = [_replace_dict_keys(v, converter) for v in value]
+
+        new_d[converter(key)] = value
+    return new_d
```
jmorel, line 186: there is a "metadata" property for all assets that the user can fill with a dict. The keys of this dict shouldn't be converted to snake_case since they are user input.

jmorel (changes requested):

Esadruhn:
Should this be in snake case ? https://github.com/SubstraFoundation/substra-backend/pull/315/files#diff-1c7f95fdd9883397a5489522f326ccb7R42

Kelvin-M:
@Esadruhn 
Nice catch for the `authorizedIDs`
For `compositeTraintuple`, the chaincode return some asset with : `"traintuple_type": "compositeTraintuple"`. So we first need to update the chaincode before changing this value I think

Esadruhn (dismissed):

jmorel (approved):
Tested while developing the frontend part of this and working as expected!

backend/substrapp/ledger_utils.py
```diff
@@ -157,6 +161,50 @@ def _wrapper(*args, **kwargs):
     return _retry
 
 
+def to_snake_case(camel_str):
+    name = CAMEL_TO_SNAKE_PATTERN.sub(r'\1_\2', camel_str)
+    name = CAMEL_TO_SNAKE_PATTERN_2.sub(r'\1_\2', name).lower()
+    return name.replace('_i_d', '_id')
```
inalgnu, line 167: I think something like 
```python 
def to_snake_case(camel_str):
    if len(camel_str) <= 2:
        return camel_str.lower()
    
    name = CAMEL_TO_SNAKE_PATTERN.sub(r'\1_\2', camel_str)
    name = CAMEL_TO_SNAKE_PATTERN_2.sub(r'\1_\2', name).lower()
    return name
```
Will englobe more cases and not only 'iD', what do you think?

Kelvin-M: Nice idea ! But as we will do the update (camel to snake case) in the chaincode, this function will be removed in the future.
I don't know if it is worth to do this update in that context !

Kelvin-M: And this case is to handle `IDToKey` and `ComputePlanID`, so you suggestion won't work in those cases no ?

inalgnu: It looks that it works, but anyway if this is a temporary code, fine for me also to keep it like this
```python
import re

CAMEL_TO_SNAKE_PATTERN = re.compile(r'(.)([A-Z][a-z]+)')
CAMEL_TO_SNAKE_PATTERN_2 = re.compile(r'([a-z0-9])([A-Z])')

def to_snake_case(camel_str):
    if len(camel_str) <=2:
        return camel_str.lower()
    
    name = CAMEL_TO_SNAKE_PATTERN.sub(r'\1_\2', camel_str)
    name = CAMEL_TO_SNAKE_PATTERN_2.sub(r'\1_\2', name).lower()
    return name


print(to_snake_case('iD'))
print(to_snake_case('IDToKey'))
print(to_snake_case('ComputePlanID'))
```

Kelvin-M: Yes, I found this was made for `'authorizedIDs'`

inalgnu: I understand now :-). Let forget about my comment, and remove this after the chaincode update

Kelvin-M: Removal incoming: https://github.com/SubstraFoundation/substra-chaincode/pull/125

inalgnu (approved):

PR #316, opened 2020-09-10
Remove nested list in api result

Kelvin-M:
Companion PR:
https://github.com/SubstraFoundation/substra/pull/223
https://github.com/SubstraFoundation/substra-frontend/pull/81

### DIFF
--- a/backend/substrapp/views/filters_utils.py
+++ b/backend/substrapp/views/filters_utils.py
@@ -1,4 +1,6 @@
 import logging
+import itertools
+
 from urllib.parse import unquote
 
 from substrapp.ledger_utils import query_ledger
@@ -98,6 +100,14 @@ def _get_model_tuple(model):
         return model['traintuple']
     else:
         raise NotImplementedError
+
+
+def flatten_without_duplicates(list_of_list):
+    res = []
+    for item in itertools.chain.from_iterable(list_of_list):
+        if item not in res:
+            res.append(item)
+    return res
 
 
 def filter_list(channel_name, object_type, data, query_params):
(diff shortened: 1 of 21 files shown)


jmorel (approved):

jmorel (commented):
Tested through the companion PR in the frontend!

backend/substrapp/views/filters_utils.py
```diff
@@ -100,6 +102,14 @@ def _get_model_tuple(model):
         raise NotImplementedError
 
 
+def flatten_without_duplicates(list_of_list):
```
Esadruhn, line 105: If speed is important, might be better to do it with a set:

```python
def flatten_set(list_of_list):
    res = set()
    for item in itertools.chain.from_iterable(list_of_list):
        if item not in res:
            res.add(item)
    return list(res)
```

```python
data = [[j for j in range(100)] for i in range(1000)]
%timeit flatten_without_duplicates(data)
%timeit flatten_set(data)
```
--> with list: 10 loops, best of 5: 241 ms per loop
--> with set: 100 loops, best of 5: 8.6 ms per loop

Kelvin-M: I don't know if it's a bottleneck of the application :)

Esadruhn (approved):
the comment on set is not blocking

backend/substrapp/views/filters_utils.py
Esadruhn: Probably not ! Not blocking for the merge

inalgnu (approved):

PR #319, opened 2020-09-11
GRPC: add keepalive (120 sec)

AurelienGasser:
Add a keepalive (120 sec) to all the GRPC links﻿.

This prevents links from being closed by network software/equipment because of inactivity.

### DIFF
--- a/charts/substra-backend/values.yaml
+++ b/charts/substra-backend/values.yaml
@@ -58,6 +58,11 @@ backend:
     #   cpu: 100m
     #   memory: 128Mi
 
+  grpc:
+    keepalive:
+      timeMs: 120000
+
+
 outgoingNodes: []
   # - name: nodeId
   #   secret: nodeSecret
@@ -316,6 +321,3 @@ securityContext:
   runAsUser: 1001
   runAsGroup: 1001
 
-grpc:
-  keepalive:
-    timeMs: 120000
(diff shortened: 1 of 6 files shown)


inalgnu (approved):

charts/substra-backend/values.yaml
```diff
@@ -315,3 +315,7 @@ securityContext:
   fsGroup: 1001
   runAsUser: 1001
   runAsGroup: 1001
+
+grpc:
+  keepalive:
+    timeMs: 120000
```
AlexandrePicosson, line 321: As this is only a backend setting and not enforced globally (currently only the backend use GRPC but we never know) shouldn't this be located under the `backend:` key ?

AurelienGasser: That makes sense. 

Note: We have plans to separate the event app from the backend. If we do so, then the key would have to be moved again.

AlexandrePicosson: Yes, but if you move the event app out of the backend you would also need to move it from here to an `eventApp:` key !

Kelvin-M: I agree with Alexandre points.
We may first need to put this key under `backend`

AurelienGasser: https://github.com/SubstraFoundation/substra-backend/pull/319/commits/072bfbf83d2f1a66d11d88ce036e4014077f6f1e

PR #331, opened 2020-09-19
Faster startup in case of initialization error + more verbose logs

AurelienGasser:
This PR makes the startup faster and more verbose in case of initialization error (e.g. peer not in channel yet)

1. If an exception occurs when calling the ledger in the application initialization (register-node), do not crash the app. 

Instead, retry the ledger call indefinitely. This is faster than crashing the pod and waiting for it to restart. Note that kubernetes will still crash the pod if the readiness probe fails (i.e. initialization repeatedly fails for > 30 secs or so)

2. With the `log-master=true`﻿ option, no logs are output until the app initialization is complete or the pod is crashed. In other words, if the app initialization takes a long time, we can't see the logs until either the initialization finally succeeds, or the container is stopped by kubernetes.

image

...wait...

...wait more...

...what is going on?....

After a good 30 seconds all the logs arrive at once

image

image

This PR removes this option. As a result, logs aren't delayed anymore.

### DIFF
--- a/backend/node-register/apps.py
+++ b/backend/node-register/apps.py
@@ -4,8 +4,6 @@ from django.apps import AppConfig
 from django.conf import settings
 from substrapp.ledger_utils import invoke_ledger
 
-from aiogrpc import RpcError
-
 logger = logging.getLogger(__name__)
 LEDGER = getattr(settings, 'LEDGER', None)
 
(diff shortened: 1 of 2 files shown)


AurelienGasser:
e2e tests :heavy_check_mark:

Kelvin-M (approved):

PR #332, opened 2020-09-21
Simplify local folder deletion

AurelienGasser:
Currently, the local folder is cleaned by creating a kubernetes pod. 

This has several disadvantages : 

- First, it is very expensive. Starting a pod and waiting for its completion takes a very long time and a lot of server resources.

- When the backend is restarted, all the events are replayed and we attempt to delete the local folder for all the compute plans that were ever registered on the platform. Since each deletion creates a pod and waits for its completion, **the worker is busy for a very long time**, and cannot be used for computation tasks.

- Only the folder contents are deleted, not the folder itself

This PR addresses the above issues by mounting the local folder root in the worker pod, and executing the deletion from the worker pod. 

`remove_local_folders` now takes <1ms.


Kelvin-M (approved):

backend/substrapp/tasks/tasks.py
```diff
@@ -457,11 +458,14 @@ def remove_local_folders(compute_plan_id):
...
-    logger.info(f'Deleting local volume for compute plan {compute_plan_id}')
-
-    volume_id = get_volume_id(compute_plan_id)
-
-    remove_local_volume(volume_id)
+    try:
+        local_folder = get_local_folder(compute_plan_id)
+        logger.info(f'Deleting local folder {local_folder}')
+        shutil.rmtree(local_folder)
+    except FileNotFoundError:
```
inalgnu, line 465:
```suggestion
    except FileNotFoundError:
           logger.info(f'No local folder with name {local_folder} found')
```

inalgnu (approved):

PR #333, opened 2020-09-28
Extract values for skaffold

Kelvin-M:
In order to have a simpler skaffold file and add the possibility to deploy with helm only, we extract helm values from skaffold file into separate values files.

@natct10 Do we need to update the documentation for that part ?

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -84,6 +84,7 @@ deploy:
       - name: backend-org-1
         chartPath: charts/substra-backend
         namespace: org-1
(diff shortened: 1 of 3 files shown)


natct10:
Hi, just to be sure: this allows to use both `skaffold dev # or run` and helm with something like `helm install --name backend-org-1 --namespace org-1 -f values/backend-org-1.yaml`?
I think this is really easier to read, thank you @Kelvin-M

Kelvin-M:
@natct10  Yes, that's the goal :)

inalgnu (approved):

PR #335, opened 2020-09-29
Remove flower dependency

inalgnu:
## Remove flower 🌹 

This PR removes flower dependency, which is not used at all


charts/substra-backend/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: substra-backend
 home: https://substra.org/
-version: 1.2.1
+version: 1.3.1
```
inalgnu, line 4: This is a breaking change, maybe I should have upgrade the chart version to 2.0.0. I'm not sure about the impact here. cc @Kelvin-M

Kelvin-M: I don't think is a breaking change as we juste remove an unused pod :) 
The 1.3.1 is good for me !

Kelvin-M (approved):

PR #338, opened 2020-09-30
Clean-up ledger API

AurelienGasser:
This is a step towards a more modular backend.

Changes:

- Don't execute any HLF code at settings initialization time. Move ledger connection code from the `settings/` folder to a package in `substrapp/`.
- Cleanup + make settings consistent: all the settings live in `settings/deps/ledger.py`, which is populated from a new ConfigMap (`substra-backend-ledger`).
- Re-structure HLF code into a `substrapp/ledger` package: `api.py`, `assets.py`, `connection.py` and `exceptions.py`


AurelienGasser:
e2e tests: :heavy_check_mark:

backend/backend/settings/common.py
```diff
@@ -224,6 +224,18 @@ def to_bool(value):
...
 LOGGING_USE_COLORS = to_bool(os.environ.get('LOGGING_USE_COLORS', True))
+
+
+# Without this import statement, hfc.fabric outputs DEBUG-level logs regardless of LOGGING configuration.
+# Importing the module early ensures these steps happen in order:
+# 1. hfc.fabric is imported and all the associated loggers are configured  (log level, handler, etc...)
+# 2. Loggers are re-configured with the `LOGGING` statement, overriding the configuration from step 1
+try:
+    import hfc.fabric
+except:
```
Kelvin-M, line 235: What exception you can have here ? You should catch the right one :)

backend/backend/settings/test.py
```diff
@@ -5,5 +5,5 @@
 import logging
 logging.disable(logging.CRITICAL)
 
-LEDGER_CALL_RETRY = False
 ORG_NAME = 'OrgTestSuite'
+LEDGER_SYNC_ENABLED = True
```
Kelvin-M, line 9: Why do you change the test setting from `LEDGER_CALL_RETRY = False` to `LEDGER_SYNC_ENABLED = True` ?

Kelvin-M: Okay `LEDGER_SYNC_ENABLED = True` moved to test settings :)

AurelienGasser: `LEDGER_CALL_RETRY` has no effect in the tests: it's [only used here](https://github.com/SubstraFoundation/substra-backend/commit/f0d39f4dadbfdef6a3b23c8ffc37c09ddf2e5010#diff-342b571689fc99df2ba094e899fbd715R35) and this line is never executed during test runs (tested)

backend/backend/settings/common.py
AurelienGasser: Fixed: https://github.com/SubstraFoundation/substra-backend/pull/338/commits/c8600c1e9b8d95f64ab4ec25fc06fec7ca596e13

This try/catch was intended to show that not importing fabric is a valid use case. 

However, if we want to conditionally stop using fabric in the future, we probably will want to still do the `import`.

Kelvin-M (approved):

inalgnu (approved):
Great Job!

PR #341, opened 2020-10-06
move substra.tasks to worker

AurelienGasser:


AurelienGasser:
@inalgnu @Kelvin-M I tried to moved `substrapp/tasks` to a new folder `worker/`. It's pretty straightforward.

Is it a step in the right direction?

PR #342, opened 2020-10-07
Use random UUIDs for asset keys (instead of content hashes)

### LINKED ISSUE #221: Objective unicity check

AurelienGasser:
Fixes #221 

Companion PRs: 
- https://github.com/SubstraFoundation/substra-chaincode/pull/128
- https://github.com/SubstraFoundation/substra/pull/235
- https://github.com/SubstraFoundation/substra-tests/pull/153

# Background

Historically, asset keys (`pkhash`) have been computed by hashing the assets contents/properties:

- for instance, a Traintuple's `pkhash` is a hash of its `(AlgoKey, InModels, DataManagerKey, etc...)`
- an Algo's `pkhash` is a hash of its algo file

In other words, an asset's **primary key** and **checksum** are the _same thing_, stored in the `pkhash` field. Depending on the asset type, `pkhash` are calculated by either the backend or the chaincode.

# This PR

This PR moves away from the `pkhash` model:

- Each asset's primary key (`key`) is a [globally unique UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier), aka UUID v4 or GUID, which is computed randomly. In other words, each asset has a completely random `key` which is independent from the asset's properties. *All* keys are now generated in the backend: the chaincode doesn't calculate any key, and instead honors the keys provided by the backend.
- Assets hashes are renamed to `checksum`. They are only calculated for asset types associated with a _file_: `Algo` (algo), `DataManager` (opener), `Datasample` (data sample), `Objective` (description), etc. For non-file asset types (e.g. `Traintuple`), no checksum is calculated. Checksums are used exclusively for content integrity checks, i.e. `assert hash(downloaded_file) == checksum`

## Effects on asset duplication

It is now possible to create duplicate assets, without triggering an error.

 For instance, registering the exact same Algo twice will succeed. It will **not** trigger a `HTTP 409` response, or an `AlreadyExists` exception. Two distinct assets will be created, the only difference between them being their `key`s.

# Future work

To simplify diffs in this PR, some work has been postponed to a future PR (which should happens shortly after this PR is merged):

## Rename fields

To simplify APIs and make them consistent, rename the following fields

- `dataset.keys` => `dataset.data_sample_keys`
- `compute_plan_id` => `key` / `compute_plan_key`


AurelienGasser:
cc @inalgnu @Kelvin-M 

/!\ WIP

As discussed with @inal, there's a functional "OK" to implement this change.

This change has multiple benefits: one of them is that it removes the restriction on duplicate assets, which will help in the case of melloddy - avoid redeploys.

**For now** this PR copies the value of `pkhash` into a new field `checksum` (see discussion on naming below). `pkhash` remains the primary key, and `checksum` is used for, well, checksums. 

**I have not yet implemented the part where `pkhash = new guid()`. So far basically `pkhash == checksum`.** 

Question: I'm thinking of two renamings, please let me know what you think:

1. pkhash -> key. Rationale: I think this change has already occurred in the CLI + once we move to GUIDs, the name `pkhash` becomes inaccurate
2. For the new field, I propose `checksum`: Rationale: While the code currently uses the term "hash" (I don't suggest changing that), "checksum" is basically the same thing but with an extra piece of information: "the intent is to perform an integrity check". Note that in the current state of the PR, I've named the field "checksum", but it's quite easy to rename it.

Any other general thought welcome :)

AurelienGasser:
e2e tests: :heavy_check_mark:

backend/substrapp/migrations/0004_key_checksum.py
```diff
@@ -0,0 +1,23 @@
+from django.apps import apps
+from django.db import migrations, models
+from django.db.models import F
```
inalgnu, line 3: looks like is not used

AurelienGasser: Fixed in https://github.com/SubstraFoundation/substra-backend/pull/342/commits/9186a1de0490d28af75f6dd9c5228e4a2c217ff5

backend/substrapp/utils.py
```diff
@@ -313,3 +313,7 @@ def timed(*args, **kw):
         return result
 
     return timed
+
+
+def new_uuid():
```
inalgnu, line 318: I'm not sure this is necessary, If I need to create a new UUID my first thought would be to use the uuid.uuid4() library, not the utils function.
I think It can be removed as it changes nothing to the library behavior

backend/substrapp/views/utils.py
```diff
@@ -143,43 +143,11 @@ def _download_remote_file(self, channel_name, ledger_field, asset):
...
-    return None
-
-
 def validate_pk(pk):
-    if len(pk) != 64:
-        raise exceptions.BadRequestError(f'Wrong pk {pk}')
-
     try:
-        int(pk, 16)  # test if pk is correct (hexadecimal)
+        uuid.UUID(pk)
```
inalgnu, line 148: do we still need to validate_pk?

backend/substrapp/utils.py
AurelienGasser: That makes sense. I was unsure whether or not to keep this function :)

AurelienGasser: Fixed by https://github.com/SubstraFoundation/substra-backend/pull/342/commits/f23943fe9fdaefa65da74a40d845297a43bcd358

backend/substrapp/views/utils.py
AurelienGasser: This function helps in the sense that it provides a consistent HTTP response for invalid input keys. 

Without this function, the key would be propagated to a ledger invoke, a ledger query, a DB call etc, resulting in different error types. I tried running the unit tests without this function and they fail

```
AttributeError: 'Settings' object has no attribute 'LEDGER_CALL_RETRY'
```

That's because we try to do an invoke using an invalid key.

Note: I renamed the function to `validate_key` because it makes more sense than `validate_pk` IMO https://github.com/SubstraFoundation/substra-backend/pull/342/commits/41738f71c316fe6866798bada0d3bbefbe6c9628

backend/substrapp/tasks/tasks.py
```diff
@@ -181,7 +178,7 @@ def get_and_put_local_model_content(hash_key, out_model, model_dst_path):
     """Get local model content."""
     from substrapp.models import Model
 
-    model = Model.objects.get(pk=out_model['hash'])
+    model = Model.objects.get(pk=out_model['key'])
```
Kelvin-M, line 181: I think you can use `key` instead of `pk` here :)

backend/substrapp/tasks/tasks.py
```diff
@@ -217,7 +214,7 @@ def fetch_model(channel_name, parent_tuple_type, authorized_types, input_model,
         get_and_put_model_content(
             channel_name,
             tuple_type,
-            input_model['traintuple_key'] + HASH_KEY_SUFFIX_TRUNK,
+            input_model['traintuple_key'],
```
Kelvin-M, line 217: Can you explain why we can safely remove the hash suffix here ?

backend/substrapp/tasks/tasks.py
```diff
@@ -288,9 +285,7 @@ def prepare_composite_traintuple_input_models(channel_name, directory, tuple_):
     # get the output head model
     head_model_dst_path = path.join(directory, f'model/{PREFIX_HEAD_FILENAME}{head_model_key}')
     raise_if_path_traversal([head_model_dst_path], path.join(directory, 'model/'))
-    get_and_put_local_model_content(
-        head_model_key + HASH_KEY_SUFFIX_HEAD, metadata['out_head_model']['out_model'], head_model_dst_path
-    )
+    get_and_put_local_model_content(head_model_key, metadata['out_head_model']['out_model'], head_model_dst_path)
```
Kelvin-M, line 288: Can you explain why we can safely remove the hash suffix here ?

backend/substrapp/tasks/tasks.py
```diff
@@ -302,7 +297,7 @@ def prepare_composite_traintuple_input_models(channel_name, directory, tuple_):
         get_and_put_model_content(
             channel_name,
             tuple_type,
-            trunk_model_key + HASH_KEY_SUFFIX_TRUNK,
+            trunk_model_key,
```
Kelvin-M, line 300: Same as before :)

backend/substrapp/tasks/tasks.py
```diff
@@ -334,15 +329,14 @@ def prepare_testtuple_input_models(channel_name, directory, tuple_):
         metadata = get_object_from_ledger(channel_name, traintuple_key, 'queryCompositeTraintuple')
         head_model_dst_path = path.join(directory, f'model/{PREFIX_HEAD_FILENAME}{traintuple_key}')
         raise_if_path_traversal([head_model_dst_path], path.join(directory, 'model/'))
-        get_and_put_local_model_content(traintuple_key + HASH_KEY_SUFFIX_HEAD, metadata['out_head_model']['out_model'],
-                                        head_model_dst_path)
+        get_and_put_local_model_content(traintuple_key, metadata['out_head_model']['out_model'], head_model_dst_path)
```
Kelvin-M, line 332: same as before

backend/substrapp/tasks/tasks.py
```diff
@@ -334,15 +329,14 @@ def prepare_testtuple_input_models(channel_name, directory, tuple_):
...
 
         model_dst_path = path.join(directory, f'model/{PREFIX_TRUNK_FILENAME}{traintuple_key}')
         raise_if_path_traversal([model_dst_path], path.join(directory, 'model/'))
         get_and_put_model_content(
             channel_name,
             traintuple_type,
-            traintuple_key + HASH_KEY_SUFFIX_TRUNK,
+            traintuple_key,
```
Kelvin-M, line 339: Same as before

backend/substrapp/tasks/tasks.py
```diff
@@ -377,9 +371,10 @@ def prepare_models(channel_name, directory, tuple_type, tuple_):
 def prepare_opener(directory, tuple_):
     """Prepare opener for tuple execution."""
     from substrapp.models import DataManager
+    dataset_key = tuple_['dataset']['key']
     data_opener_hash = tuple_['dataset']['opener_hash']
 
-    datamanager = DataManager.objects.get(pk=data_opener_hash)
+    datamanager = DataManager.objects.get(pk=dataset_key)
```
Kelvin-M, line 377: I think you can use key here no ?

backend/substrapp/tasks/tasks.py
```diff
@@ -963,24 +958,26 @@ def save_models(subtuple_directory, tuple_type, subtuple_key):
     models = {}
 
     if tuple_type in [TRAINTUPLE_TYPE, AGGREGATETUPLE_TYPE]:
-        file, file_hash = save_model(subtuple_directory, subtuple_key)
+        model, storage_address = save_model(subtuple_directory, subtuple_key)
         models['end_model'] = {
-            'file': file,
-            'hash': file_hash
+            'key': model.pk,
```
Kelvin-M, line 963: You can use `model.key` here no ?

backend/substrapp/tasks/tasks.py
```diff
@@ -1009,54 +1009,54 @@ def save_model(subtuple_directory, hash_key, filename='model'):
...
-    end_model_file_hash = get_hash(end_model_path, hash_key)
-    instance = Model.objects.create(pkhash=end_model_file_hash, validated=True)
+    checksum = get_hash(end_model_path, hash_key)
+    instance = Model.objects.create(checksum=checksum, validated=True)
 
     with open(end_model_path, 'rb') as f:
         instance.file.save('model', f)
     current_site = getattr(settings, "DEFAULT_DOMAIN")
-    end_model_file = f'{current_site}{reverse("substrapp:model-file", args=[end_model_file_hash])}'
+    storage_address = f'{current_site}{reverse("substrapp:model-file", args=[instance.pk])}'
```
Kelvin-M, line 1018: I think you should change `pk` to `key` here too no ?

backend/substrapp/tasks/tasks.py
```diff
@@ -1009,54 +1009,54 @@ def save_model(subtuple_directory, hash_key, filename='model'):
...
 
 
-def remove_intermediary_models(model_hashes):
+def remove_intermediary_models(model_keys):
     from substrapp.models import Model
 
-    models = Model.objects.filter(pkhash__in=model_hashes, validated=True)
-    filtered_model_hashes = [model.pk for model in models]
+    models = Model.objects.filter(key__in=model_keys, validated=True)
+    filtered_model_keys = [str(model.pk) for model in models]
```
Kelvin-M, line 1038: I think you should change `pk` to `key` here too no ?

backend/substrapp/tasks/tasks.py
```diff
@@ -963,24 +958,26 @@ def save_models(subtuple_directory, tuple_type, subtuple_key):
...
     elif tuple_type == COMPOSITE_TRAINTUPLE_TYPE:
-        for m_type, filename, hash_salt in [('end_head_model', OUTPUT_HEAD_MODEL_FILENAME, HASH_KEY_SUFFIX_HEAD),
-                                            ('end_trunk_model', OUTPUT_TRUNK_MODEL_FILENAME, HASH_KEY_SUFFIX_TRUNK)]:
-            file, file_hash = save_model(
+        for m_type, filename in [('end_head_model', OUTPUT_HEAD_MODEL_FILENAME),
+                                 ('end_trunk_model', OUTPUT_TRUNK_MODEL_FILENAME)]:
+            model, storage_address = save_model(
                 subtuple_directory,
-                subtuple_key + hash_salt,
+                subtuple_key,
```
Kelvin-M, line 973: Can you explain why we don't need to use hash_salt anymore here ?

Kelvin-M (changes requested):
Nice work !

backend/substrapp/tests/assets.py
```diff
@@ -14,22 +14,22 @@
 
 objective = [
     {
-        "key": "48cc5ce4d352b859e202adb0d639ee092c78cae7a2a1df0fccd88c63b3466ad8",
+        "key": "48cc5ce4-d352-b859-e202-adb0d639ee09",
```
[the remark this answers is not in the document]
AurelienGasser: it doesn't work anymore :(

backend/substrapp/tests/tests_misc.py
```diff
@@ -87,8 +87,9 @@ def test_log_success_tuple(self):
         with patch('substrapp.ledger.api.update_ledger') as mupdate_ledger:
             mupdate_ledger.return_value = None
             res = {
-                'end_model_file_hash': 'hash',
-                'end_model_file': 'storage_address',
+                'end_model_key': '<some_key>',
+                'end_model_checksum': 'hash',
```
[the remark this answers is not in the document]
AurelienGasser: Not yet. But this dict is independent from the chaincode: it's used only in backend tasks.

backend/substrapp/views/aggregatetuple.py
```diff
@@ -23,17 +25,24 @@ def get_queryset(self):
...
+    def commit(self, serializer, channel_name):
         # create on ledger
         try:
-            data = serializer.create(channel_name, serializer.validated_data)
+            data = serializer.validated_data
+            if data['rank'] == 0 and not data['compute_plan_id']:
+                # Auto-create compute plan
+                res = create_compute_plan(channel_name, data={})
+                data['compute_plan_id'] = res['compute_plan_id']
+            data = serializer.create(channel_name, data)
```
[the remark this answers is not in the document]
AurelienGasser: I think the serializer should take care only of its own type (Aggregate Tuple in this case), and not manipulate other types (Compute Plan). I'd argue that's the responsibility of the view.

> Is it a behavior we want to have ?

I simply kept on supporting this use-case (create a compute plan automatically) :) It was implemented in the chaincode prior to this PR but now it's implemented in the backend (because the backend generates keys). We could consider dropping this feature in the future cc @inalgnu

backend/substrapp/tasks/tasks.py
AurelienGasser: It is not needed anymore because primary keys are not based on model contents anymore. As a result, there's no possibility of a primary key conflict, hence no need for a hash salt.

This feature (hash suffixes) was implemented to support the edge case when the trunk and head models are identical.

More generally, it's now ok to save duplicate models.

backend/substrapp/tests/assets.py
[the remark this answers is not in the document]
Kelvin-M: So we need to fix it no ?

backend/substrapp/tests/tests_misc.py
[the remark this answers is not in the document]
Kelvin-M: Okay

backend/substrapp/tests/assets.py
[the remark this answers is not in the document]
AurelienGasser: Yes although that's independent from this PR

Kelvin-M (approved):

AurelienGasser:
@Kelvin-M would you mind reviewing the last 2 commits? 

A bug was found in Titanic examples, these 2 commits fix the bug (and similar potential bugs) and add new tests.

AurelienGasser:
@Kelvin-M would you mind reviewing the last commit?

A bug was found using the Titanic examples: the commit fixes the bug (and similar potential bugs) and add new tests.

PR #344, opened 2020-10-27
Data samples order

jmorel:
:warning: In order for this PR to be merged, https://github.com/SubstraFoundation/substra-tools/pull/43 has to be merged first and the 0.7.0 release done.

With this change, the relative order of data samples is added to the command executed by the algo container during training.

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -48,6 +48,7 @@ TUPLE_COMMANDS = {
     AGGREGATETUPLE_TYPE: 'aggregate',
 }
 
+DATA_FOLDER = '/sandbox/data'
 MODEL_FOLDER = '/sandbox/model'
 OUTPUT_MODEL_FOLDER = '/sandbox/output_model'
 OUTPUT_PERF_PATH = '/sandbox/perf/perf.json'
@@ -780,7 +781,7 @@ def prepare_volumes(subtuple_directory, tuple_type, compute_plan_key, compute_pl
                 symlinks_volume[real_path] = {'bind': f'{real_path}', 'mode': 'ro'}
 
     volumes = {
-        data_path: {'bind': '/sandbox/data', 'mode': 'ro'},
+        data_path: {'bind': DATA_FOLDER, 'mode': 'ro'},
         opener_path: {'bind': '/sandbox/opener', 'mode': 'ro'}
     }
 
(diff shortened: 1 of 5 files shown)


AurelienGasser:
@jmorel Wouldn't it be simpler to _always_ respect the order of datasamples?

In `prepare_materials`, we could create a `data_samples.json` file, with contents `subtuple['dataset']['keys']` (ordered list of keys). Substra-tools could read from that file and always respect the order (which is the order specified when creating the traintuple).

samlesu:
> @jmorel Wouldn't it be simpler to _always_ respect the order of datasamples?
> 
> In `prepare_materials`, we could create a `data_samples.json` file, with contents `subtuple['dataset']['keys']` (ordered list of keys). Substra-tools could read from that file and always respect the order (which is the order specified when creating the traintuple).

I think that's already the case in this PR: the backend always set the data samples order. 

As this change is not backward compatible (the backend won't work with a previous version of substratools) I think it would be better to simplify the current substratools interface: have an argument specifying the data sample folder names (or paths) to load in the correct order (IMO substratools should not have to list the subfolders of the rootfolder).

jmorel:
Actually there is one place where this PR doesn't enforce data samples order: for testtuples.

As for the interface with substra-tools, my reasoning was that adding an extra argument would allow the new substra tools to work with earlier versions of the backend, which could be nice for environments that can't be updated easily.

And about the json file, while I do agree that it could be simpler that having an extra long command (is there a limit to the size of a command btw?) it would introduce a second way of sending instructions to substra tools. In that case wouldn't it be better to remove all commands args and just populate a json file with all information?

samlesu:
> Actually there is one place where this PR doesn't enforce data samples order: for testtuples.
> 
> As for the interface with substra-tools, my reasoning was that adding an extra argument would allow the new substra tools to work with earlier versions of the backend, which could be nice for environments that can't be updated easily.
> 
> And about the json file, while I do agree that it could be simpler that having an extra long command (is there a limit to the size of a command btw?) it would introduce a second way of sending instructions to substra tools. In that case wouldn't it be better to remove all commands args and just populate a json file with all information?

+1 for the json file (or yaml?), either use a single file for all parameters or use only command arguments. I do think using a file will make things easier to maintain, but I don't think it should be done in this PR.

I think we should focus now on code simplicity/maintainability rather than on compatibility but that's my personal point of view.

AurelienGasser:
@jmorel Thanks for the details! 

I don't have a strong opinion about files v.s. argument, my concern was more about the duplication of information about which data samples to use.

> I think it would be better to simplify the current substratools interface: have an argument specifying the data sample folder names (or paths) to load in the correct order (IMO substratools should not have to list the subfolders of the rootfolder).

+1, I think this would make sense.

jmorel:
@AurelienGasser @samlesu Done!

AurelienGasser (approved):
LGTM

AurelienGasser:
@jmorel FYI there's a test failing. Probably a case of adding `keys` to a test dataset.

jmorel:
Thanks @AurelienGasser, there was an incomplete feature, it's fixed!

Kelvin-M (approved):

samlesu (approved):

jmorel:
Hello @AurelienGasser, I'm very happy to report that the tests are passing! (there was a bug following the rebase on master).

PR #346, opened 2020-11-04
Add LedgerNotFound exception in retry update_ledger

Kelvin-M:
Add LedgerNotFound exception in retry `update_ledger`
A org may try to launch a traintuple (i.e. `log_start_tuple` via an invoke). During this procedure, the transaction is endorsed by each peer (Another org here).
But, even if the first org had already the block with the traintuple metadata, the second org may not have it at this moment so the endorsement request will fail with a `no asset for key` error.
In order to fix this potential issue, we can use the retry mecanism and add LedgerNotFound exception)


AurelienGasser:
Nice! Maybe a good opportunity to rename `LedgerNotFound` to `LedgerAssetNotFound`?

AurelienGasser (approved):

ClementGautier (approved):

inalgnu (approved):

PR #351, opened 2020-11-24
Change from Thread to Process

Kelvin-M:
## Description
It looks like using threads to download models in parallel is slower than Process.
For 11 * 2 Go, 

**Thread** 
```
#1 -  49.08039045333862 seconds - 2097152000                                                               
#5 - 67.21909880638123 seconds - 2097152000                                                               
#2 - 69.7494068145752 seconds - 2097152000                                                                
#4 - 73.38624572753906 seconds - 2097152000                                                               
#3 - 76.63423752784729 seconds - 2097152000                                                               
#6 - 78.72280073165894 seconds - 2097152000                                                               
#7 - 81.33725309371948 seconds - 2097152000                                                               
#11 - 81.71029543876648 seconds - 2097152000                                                              
#8 - 84.1978645324707 seconds - 2097152000                                                                
#9 - 84.2302668094635 seconds - 2097152000                                                                
#10 - 82.81430339813232 seconds - 2097152000  
```
**Process**
```
#1 - 12.778266668319702 seconds - 2097152000
#7 - 11.798902988433838 seconds - 2097152000
#2 - 13.3548104763031 seconds - 2097152000
#3 - 12.40976595878601 seconds - 2097152000
#4 - 11.725588083267212 seconds - 2097152000
#6 - 12.409850358963013 seconds - 2097152000
#5 - 13.262670040130615 seconds - 2097152000
#8 - 13.058980703353882 seconds - 2097152000
#10 - 11.18153691291809 seconds - 2097152000
#9 - 13.346837043762207 seconds - 2097152000
#11 - 12.924207925796509 seconds - 2097152000
```

## Closes issue(s)
None

## Companion PRs
None

## How to test / repro
None

## Screenshots / Trace
None

## Changes include
- [x] Migrate Thread to Process

## Checklist
None

## Other comments
None

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -237,7 +237,7 @@ def fetch_models(channel_name, tuple_type, authorized_types, input_models, direc
 
     for proc, args in models:
         proc.join()
-        if proc._popen.returncode != 0:
+        if proc.exitcode != 0:
             exceptions.append(Exception(f'fetch model failed for args {args}'))
 
     if exceptions:
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -10,6 +10,7 @@ import logging
 import tarfile
 
 import kubernetes
+from billiard import Process
 
 from django.conf import settings
 from rest_framework.reverse import reverse
@@ -26,8 +27,7 @@ from substrapp.ledger.api import (log_start_tuple, log_success_tuple, log_fail_t
                                   query_tuples, get_object_from_ledger)
 from substrapp.ledger.exceptions import LedgerError, LedgerStatusError
 from substrapp.tasks.utils import (compute_job, get_asset_content, get_and_put_asset_content,
-                                   list_files, do_not_raise, ExceptionThread,
-                                   get_or_create_local_volume, remove_image)
+                                   list_files, do_not_raise, get_or_create_local_volume, remove_image)
 
 from substrapp.tasks.exception_handler import compute_error_code
 
@@ -227,25 +227,21 @@ def fetch_model(channel_name, parent_tuple_type, authorized_types, input_model, 
 def fetch_models(channel_name, tuple_type, authorized_types, input_models, directory):
 
     models = []
-
-    for input_model in input_models:
-        proc = ExceptionThread(target=fetch_model,
-                               args=(channel_name, tuple_type, authorized_types, input_model, directory))
-        models.append(proc)
-        proc.start()
-
-    for proc in models:
-        proc.join()
-
     exceptions = []
 
-    for proc in models:
-        if hasattr(proc, "_exception"):
-            exceptions.append(proc._exception)
-            logger.exception(proc._exception)
-    else:
-        if exceptions:
-            raise Exception(exceptions)
+    for input_model in input_models:
+        args = (channel_name, tuple_type, authorized_types, input_model, directory)
+        proc = Process(target=fetch_model, args=args)
+        models.append((proc, args))
+        proc.start()
+
+    for proc, args in models:
+        proc.join()
+        if proc._popen.returncode != 0:
+            exceptions.append(Exception(f'fetch model failed for args {args}'))
+
+    if exceptions:
+        raise Exception(exceptions)
 
 
 def prepare_traintuple_input_models(channel_name, directory, tuple_):
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -27,7 +27,8 @@ from substrapp.ledger.api import (log_start_tuple, log_success_tuple, log_fail_t
                                   query_tuples, get_object_from_ledger)
 from substrapp.ledger.exceptions import LedgerError, LedgerStatusError
 from substrapp.tasks.utils import (compute_job, get_asset_content, get_and_put_asset_content,
-                                   list_files, do_not_raise, get_or_create_local_volume, remove_image)
+                                   list_files, do_not_raise, get_or_create_local_volume, remove_image,
+                                   authenticate_worker)
 
 from substrapp.tasks.exception_handler import compute_error_code
 
@@ -162,7 +163,7 @@ def get_tuple_status(channel_name, tuple_type, key):
     return metadata['status']
 
 
-def get_and_put_model_content(channel_name, tuple_type, hash_key, tuple_, out_model, model_dst_path):
+def get_and_put_model_content(channel_name, tuple_type, hash_key, tuple_, out_model, model_dst_path, auth=None):
     """Get out model content."""
     owner = tuple_get_owner(tuple_type, tuple_)
     return get_and_put_asset_content(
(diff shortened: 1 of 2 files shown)


Kelvin-M:
Launching e2e tests !

AlexandrePicosson (approved):

Kelvin-M:
It looks like we cannot launch processes under celery tasks :(

Kelvin-M:
using `Process` from `billiard` (celery multiprocessing fork) works 🎉

AurelienGasser:
@Kelvin-M Nice work! Could you please add a condition to bypass the Process creation in the case when there's only one model to fetch?

It would also be interesting to test it if the approach in solution 6 [here](https://stackoverflow.com/a/35819773/1370722) would be even more performant, although this can be done separately from this PR.

Kelvin-M:
I would rather prefer to keep the same behavior as before. Moreover, using the same function for 1, 2 or n `in_models`, as it is a list, is preferable than handling each case. If we bypass the Process creation for one model, maybe for 2 models, Threads are faster and for n > 1000 asyncio/aiohttp is the best solution. So having Process for all, as it looks to perform well, is, imo, the good change.

For solutions proposed in the stackoverflow, it optimises RPS, which is not the case for us. I try an experiment with asyncio and it was clearly slower :(

backend/substrapp/tasks/tasks.py
```diff
@@ -227,25 +227,21 @@ def fetch_model(channel_name, parent_tuple_type, authorized_types, input_model,
...
+        args = (channel_name, tuple_type, authorized_types, input_model, directory)
+        proc = Process(target=fetch_model, args=args)
+        models.append((proc, args))
         proc.start()
 
-    for proc in models:
+    for proc, args in models:
         proc.join()
+        if proc._popen.returncode != 0:
+            exceptions.append(Exception(f'fetch model failed for args {args}'))
```
samlesu, line 241: Why do you create a list of exceptions ?

You could create a list of errors directly as an Exception is already raised line 244.

It would be great to add also the return code value in the error message.

backend/substrapp/tasks/tasks.py
```diff
@@ -227,25 +227,21 @@ def fetch_model(channel_name, parent_tuple_type, authorized_types, input_model,
...
-        models.append(proc)
+        args = (channel_name, tuple_type, authorized_types, input_model, directory)
+        proc = Process(target=fetch_model, args=args)
+        models.append((proc, args))
         proc.start()
 
-    for proc in models:
+    for proc, args in models:
         proc.join()
+        if proc._popen.returncode != 0:
```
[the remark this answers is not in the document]
Kelvin-M: Ah, you're right ! I missed this attribute !

backend/substrapp/tasks/tasks.py
```diff
@@ -227,25 +227,21 @@ def fetch_model(channel_name, parent_tuple_type, authorized_types, input_model,
 def fetch_models(channel_name, tuple_type, authorized_types, input_models, directory):
```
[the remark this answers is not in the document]
Kelvin-M: Yes, maybe :)  But like Aurélien suggestion, it's not the aim of this PR to change the current behavior !

backend/substrapp/tasks/tasks.py
Kelvin-M: I keep the old behavior: raise an Exception with the list of exceptions in each parallel job.
Return code is useless as Process log already the python Traceback :)

samlesu: I don't think it makes sense to encapsulate the error messages in exceptions, but that's fine for me.

samlesu (approved):

backend/substrapp/tasks/tasks.py
```diff
@@ -227,25 +227,21 @@ def fetch_model(channel_name, parent_tuple_type, authorized_types, input_model,
...
 
     models = []
```
AurelienGasser, line 229: @Kelvin-M just to be clear, my suggestion is to use a function call (not a thread) when there's only one model to download:


```suggestion
    if len(input_models) == 1:
        # Don't create sub-processes if there's only one model to download
        fetch_model(channel_name, tuple_type, authorized_types, input_model, directory)
        return

    models = []
```

Kelvin-M: Thank you :)
As answered [here](https://github.com/SubstraFoundation/substra-backend/pull/351#issuecomment-733540836), I would prefer to use the same solution for all cases than handle them with a special behavior for each (here 2, but we can do an analysis to know when it's better to use sequential calls or parallel calls) as it's not "slower" to spawn a Process

AurelienGasser: I just ran a quick benchark and process creation takes 6ms on my machine, nothing to argue about here :)

backend/substrapp/tasks/utils.py
```diff
@@ -40,8 +39,12 @@ def get_asset_content(channel_name, url, node_id, content_checksum, salt=None):
     return get_remote_file_content(channel_name, url, authenticate_worker(node_id), content_checksum, salt=salt)
 
 
-def get_and_put_asset_content(channel_name, url, node_id, content_checksum, content_dst_path, hash_key):
-    return get_and_put_remote_file_content(channel_name, url, authenticate_worker(node_id), content_checksum,
+def get_and_put_asset_content(channel_name, url, node_id, content_checksum, content_dst_path, hash_key, auth=None):
```
AurelienGasser, line 42: Nit: IMO it would be clearer if `auth` was a required parameter

AurelienGasser (approved):
LGTM

backend/substrapp/tasks/utils.py
Kelvin-M: This modification was done to handle the particular issue of using `Process` to fetch all models in parallel which seems to mess the usage of Django db model to get OutGoing authentication properly.
For that, we need to get and give the authentication before launching the function in the `Process`, but we may want to keep the old behavior (get authentication in the function) for other call.

backend/substrapp/tasks/tasks.py
```diff
@@ -227,25 +232,30 @@ def fetch_model(channel_name, parent_tuple_type, authorized_types, input_model,
...
+    exceptions = []
 
     for input_model in input_models:
-        proc = ExceptionThread(target=fetch_model,
-                               args=(channel_name, tuple_type, authorized_types, input_model, directory))
-        models.append(proc)
+        # Authentification should be retrieved before sending the function fetch model to a subProcess
+        # as, fetching it from django models give random results
+        # Don't know why
```
inalgnu, line 240: Just curious @Kelvin-M what kind of random results you get here? I think even we can merge this we should do further investigations in orderer to understand what is happening

Kelvin-M: Retrieving sometime a wrong password for a particular `node_id` instead of the good one :(

Kelvin-M:
Do not merge right now, I will try to see if we can fix the auth problem without modifying the authentication code part !

PR #352, opened 2020-11-26
Handle chaincode pagination

Kelvin-M:
## Description
The aim of this PR is to handle chaincode pagination for query[Asset]s.

## Closes issue(s)
None

## Companion PRs
https://github.com/SubstraFoundation/substra-chaincode/pull/131

## How to test / repro
Launch e2e tests, no regression should appear.

## Screenshots / Trace
None

## Changes include
- [x] Loop over pagination in `call_ledger`

## Checklist
- [x] I have tested this code

## Other comments
None

### DIFF
--- a/backend/substrapp/ledger/api.py
+++ b/backend/substrapp/ledger/api.py
@@ -168,7 +168,19 @@ def call_ledger(channel_name, call_type, fcn, *args, **kwargs):
     ts = time.time()
     error = None
     try:
-        return _call_ledger(channel_name, call_type, fcn, *args, **kwargs)
+        response = _call_ledger(channel_name, call_type, fcn, *args, **kwargs)
+
+        if isinstance(response, dict) and 'bookmarks' in response:
+            results = response['result']  # first result
+            while response['result']:
+                kwargs['args'] = {'bookmarks': response['bookmarks']}
+                response = _call_ledger(channel_name, call_type, fcn, *args, **kwargs)
+                results.extend(response['result'])  # following results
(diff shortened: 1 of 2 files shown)


backend/substrapp/ledger/api.py
```diff
@@ -168,7 +168,19 @@ def call_ledger(channel_name, call_type, fcn, *args, **kwargs):
     ts = time.time()
     error = None
     try:
-        return _call_ledger(channel_name, call_type, fcn, *args, **kwargs)
+        response = _call_ledger(channel_name, call_type, fcn, *args, **kwargs)
+
+        if isinstance(response, dict) and 'bookmark' in response:
+            results = response['results']  # first results
+            while response['results'] and len(response['bookmark']) > 0:
```
inalgnu, line 175:
```suggestion
            while len(response['bookmark']) > 0:
```
should be suffisent, no?

inalgnu (commented):
would be nice to add a unit test that covers a bookmark use case from the ledger

backend/substrapp/ledger/api.py
Kelvin-M: Nope, `response['results']` should not be empty as, if the query returns, at least, a bookmark, it will always return a bookmark (end bookmark), even if `response['results']` is empty.

inalgnu: that make sens, thank's

inalgnu (approved):

PR #358, opened 2020-12-16
Migrate from Travis CI to GitHub Actions

AlexandrePicosson:
## Description
This PR adds a new GH Actions workflow to test the python code it is basically the same workflow as the one that was run on TravisCI.

It also adds a GH Actions workflow to test and publish the Helm chart. The Helm chart is only tested on modifications to the `charts/` dir. The chart is only published if the `charts/` dir was modified on a push to the `master` branch and the lint job succeeded. This should replicate the tests and publish scenarios we had before with TravisCI.

## Motivation and Context
We switch from TravisCI to GH Actions due to the fact that https://travis-ci.org/ will close at the end of december ([source](https://docs.travis-ci.com/user/migrate/open-source-repository-migration#q-when-will-the-migration-from-travis-ciorg-to-travis-cicom-be-completed)).

## How Has This Been Tested?

You can look at the [Checks](https://github.com/SubstraFoundation/substra-backend/pull/358/checks) tab to see the workflow logs.

I tested the cache performance, there is not a significant increase in speed that comes by caching the pip cache. This is due to the fact that the longest step in our `pip install` is not the download of the modules but their installation.
Another solution could be to cache the whole installed packages dir, this would significantly speed up everything but i'm not 100% confident in the fact that there would be no side effects.

Links:
- [Helm pipeline](https://github.com/SubstraFoundation/substra-backend/actions/runs/425703721)
- [Python pipeline](https://github.com/SubstraFoundation/substra-backend/actions/runs/425703720)

## Screenshots:
Preview of the checks of a PR with chart modifications (ignore travis):
Screenshot 2020-12-16 at 14 32 30
Preview of the Python action logs
Screenshot 2020-12-16 at 14 34 50
Preview of the Helm action pipeline
Screenshot 2020-12-16 at 14 35 57

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.


.github/workflows/python.yml
```diff
@@ -0,0 +1,43 @@
...
+          pip install --upgrade --upgrade-strategy eager -r backend/requirements.txt
+      - name: Install tools
+        run: pip install flake8 pep8-naming
+      - name: Lint
+        run: flake8 backend/
+      - name: Test
+        run: |
+          cd backend/
+          coverage run manage.py test
+          cd ..
```
mblottiere, line 42: This doesn't seem necessary since there are no other actions:
```suggestion
```

mblottiere (approved):

inalgnu (approved):
LGTM

.github/workflows/python.yml
AlexandrePicosson: Yes, you are right.

PR #369, opened 2021-01-18
[doc] prefered way to deploy with images from repo

mblottiere:
## Motivation

Rebuilding some images now leads to broken backend. While this has been fixed by #367, previous releases cannot be rebuilt correctly. However, images in [registry](https://hub.docker.com/r/substrafoundation/substra-backend) are fine, so this only really impacts local usage of previous tags.

In order to run a previous version locally, instead of rebuilding the image, this PR introduce a new skaffold config file to run an arbitrary backend version.

## How to address this ?

Skaffold allows to deploy without building (`deploy` command), so let's just leverage that.
Here is the snippet added to the readme:
```bash
SUBSTRA_BACKEND_VERSION=0.1.6
git checkout $SUBSTRA_BACKEND_VERSION
docker pull substrafoundation/substra-backend:$SUBSTRA_BACKEND_VERSION
skaffold deploy --images substrafoundation/substra-backend:$SUBSTRA_BACKEND_VERSION
```

- it will work as long as images are in the repository
- this doesn't add any maintenance burden
- as opposed to the previous implementation with a separate skaffold file, since we can checkout the tag values files will be compatible with the release

ℹ️ Here the backend is impacted but other components of the stack may face the same rebuild issue at some point. Besides, having a harmonized way to run specific tags throughout the stack is a better developer experience. So it might be desirable to implement the same strategy for other projects, what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,14 @@ Provided you have correctly setup your [network configuration](https://doc.subst
 Make sure you deploy this backend with a compatible ecosystem (chaincode, hlf-k8s, etc).
 Always refer to the [compatibility table](https://github.com/SubstraFoundation/substra#compatibility-table).
 
+The recommended way to run a specific version (0.1.6) of substra-backend is to execute:
+
+```bash
+SUBSTRA_BACKEND_VERSION=0.1.6
+git checkout $SUBSTRA_BACKEND_VERSION
+skaffold deploy --images substrafoundation/substra-backend:$SUBSTRA_BACKEND_VERSION
+```
+
 ## Companion repositories
 
 The Substra platform is built from several components (see the [architecture](https://doc.substra.ai/architecture.html) documentation for a comprehensive overview):


Kelvin-M (approved):

Kelvin-M:
For the chart, you can use the `chartPath` to use a helm hub like artifacthub.io. But it would also need to change or adapt the value file too !

Yes, the same strategy can be use for other repo like hlf-k8s !

Kelvin-M (approved):

README.md
```diff
@@ -66,6 +66,15 @@ Provided you have correctly setup your [network configuration](https://doc.subst
 Make sure you deploy this backend with a compatible ecosystem (chaincode, hlf-k8s, etc).
 Always refer to the [compatibility table](https://github.com/SubstraFoundation/substra#compatibility-table).
 
+The recommended way to run a specific version (0.1.6) of substra-backend is to execute:
+
+```bash
+export SUBSTRA_BACKEND_VERSION=0.1.6
```
Kelvin-M, line 72: export may not be needed like here : https://github.com/SubstraFoundation/hlf-k8s/pull/117/files#r560370670

mblottiere: Thanks for the feedback, but unless I'm missing something it is indeed necessary (see my answer in the other PR).

Kelvin-M (approved):

PR #373, opened 2021-01-21
Improve local volume handling in case of retries

Kelvin-M:
## Description
The aim of this PR is to be able to retry composite traintuple in a deterministic way. Indeed, as we mount the local folder as it, in case of files modification inside it with the algorithm **and** a fail the composite traintuple, the retry was done with everything clean and restore except the local folder.

## Closes issue(s)
None
## Companion PRs
None

## How to test / repro
Launch a compute plan which use the local folder to share hyperparameters between composite traintuples.

## Screenshots / Trace
None

## Changes include

## Checklist
- [x] I have tested this code

## Other comments
None

### DIFF
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -732,7 +732,7 @@ def do_task(channel_name, subtuple, tuple_type):
         # Copy content of local folder from subtuple directory to local folder in pvc
         # to fetch modification of local folder content
         distutils.dir_util.copy_tree(local_path, local_path_in_pvc)
-        logger.info(f'Content from {local_path} has been copied to {local_path_in_pvc}')
+        logger.info(f'Content from sandbox {local_path} has been copied to pvc {local_path_in_pvc}')
 
     # Evaluation
     if tuple_type == TESTTUPLE_TYPE:
@@ -822,7 +822,7 @@ def prepare_volumes(subtuple_directory, tuple_type, compute_plan_key, compute_pl
         # local data.
         if os.path.exists(local_path_in_pvc):
             distutils.dir_util.copy_tree(local_path_in_pvc, local_path)
-            logger.info(f'Content from {local_path_in_pvc} has been copied to {local_path}')
+            logger.info(f'Content from pvc {local_path_in_pvc} has been copied to sandbox {local_path}')
 
         mode = 'ro' if tuple_type == TESTTUPLE_TYPE else 'rw'
         model_volume[local_path] = {'bind': LOCAL_FOLDER, 'mode': mode}
--- a/backend/substrapp/tasks/tasks.py
+++ b/backend/substrapp/tasks/tasks.py
@@ -22,13 +22,13 @@ import boto3
 
 from backend.celery import app
 from substrapp.utils import (get_hash, get_owner, create_directory, uncompress_content, raise_if_path_traversal,
-                             get_dir_hash, get_local_folder_name, get_subtuple_directory, get_chainkeys_directory,
-                             get_local_folder, timeit)
+                             get_dir_hash, get_subtuple_directory, get_chainkeys_directory,
+                             get_cp_local_folder, timeit)
 from substrapp.ledger.api import (log_start_tuple, log_success_tuple, log_fail_tuple,
                                   query_tuples, get_object_from_ledger)
 from substrapp.ledger.exceptions import LedgerError, LedgerStatusError
 from substrapp.tasks.utils import (compute_job, get_asset_content, get_and_put_asset_content,
-                                   list_files, do_not_raise, get_or_create_local_volume, remove_image)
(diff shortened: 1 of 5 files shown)


AlexandrePicosson (approved):

backend/substrapp/tasks/tasks.py
```diff
@@ -720,6 +722,18 @@ def do_task(channel_name, subtuple, tuple_type):
...
 
+    # Handle local volume
+    if compute_plan_key is not None:
+        local_path = path.join(subtuple_directory, 'local')
+
+        # Get CP local folder from local pvc
+        local_path_in_pvc = get_local_folder(compute_plan_key)
+
+        # Copy content of local folder from subtuple directory to local folder in pvc
+        # to fetch modification of ocal folder content
```
mblottiere, line 733:
```suggestion
        # to fetch modification of local folder content
```

backend/substrapp/tasks/tasks.py
```diff
@@ -797,11 +811,21 @@ def prepare_volumes(subtuple_directory, tuple_type, compute_plan_key, compute_pl
...
+        # Get CP local folder from local pvc
+        local_path_in_pvc = get_local_folder(compute_plan_key)
+        get_or_create_local_volume(get_local_folder_name(compute_plan_key))
+
+        # Copy content of local folder from pvc to local folder in subtuple directory
+        # It allows the cp subtuple to be retried with compromising the
+        # local data.
+        if os.path.exists(local_path_in_pvc):
+            distutils.dir_util.copy_tree(local_path_in_pvc, local_path)
+            logger.info(f'Content from {local_path_in_pvc} has been copied to {local_path}')
```
mblottiere, line 825: Are we able to differentiate the step from the path?
This message (in `prepare_volumes`) is identical to the one in `do_task`: it would be nice to be able to distinguish them in the logs

mblottiere (approved):

mblottiere (approved):

backend/substrapp/tasks/tasks.py
```diff
@@ -797,11 +811,21 @@ def prepare_volumes(subtuple_directory, tuple_type, compute_plan_key, compute_pl
 
     # local volume for train like tuples in compute plan
     if compute_plan_key is not None:
-        volume_id = get_local_folder_name(compute_plan_key)
-        get_or_create_local_volume(volume_id)
+        local_path = path.join(subtuple_directory, 'local')
```
AurelienGasser, line 814: What about extracting this into a function `get_subtuple_local_folder`. 

The function `get_local_folder` could be renamed to `get_cp_local_folder`

backend/substrapp/tasks/tasks.py
```diff
@@ -720,6 +722,18 @@ def do_task(channel_name, subtuple, tuple_type):
...
+
+        # Get CP local folder from local pvc
+        local_path_in_pvc = get_local_folder(compute_plan_key)
+
+        # Copy content of local folder from subtuple directory to local folder in pvc
+        # to fetch modification of local folder content
+        distutils.dir_util.copy_tree(local_path, local_path_in_pvc)
+        logger.info(f'Content from sandbox {local_path} has been copied to pvc {local_path_in_pvc}')
+
```
AurelienGasser, line 736: What about extracting the contents of this whole block into a function `_commit_local_folder`?

The preparation block (in `prepare volumes`) could be extracted into a `_prepare_local_folder` function

AurelienGasser (commented):
LGTM, just some minor naming suggestions

backend/substrapp/tasks/tasks.py
Kelvin-M: I don't know if it's necessary to do so.
For instance, in `prepare_volumes`, it's more explicit to see that the `local volume` part is doing modification over the model dictionary.
For the do task, it 4 lines of code. But I can extract the contents to make them function if other find it better !

AurelienGasser:
Nice work! 

A possible next step would be to implement a unit or end to end test. I'm not sure how to do a unit test (I haven't given it much thought TBH), but an end-to-end test could look like this:

- create a CP with 2 tuples: tuple <i>#</i>2 depends on tuple <i>#</i>1
- tuple <i>#</i>1 sets a local value to "1"
- tuple <i>#</i>2:
  - saves the local value set by tuple <i>#</i>1 to a variable `x`
  - modifies the local value to "2"
  - crashes if x == 1 (so it  crashes right away)
 
If the code is correct, tuple <i>#</i>2 will never succeed. Before this PR, this code would succeed (tuple <i>#</i>2 would crash once then succeed on retry)

There might be an easier way to test this,, but I thought I'd share in case it's useful.

Kelvin-M:
Nice unit tests !

AurelienGasser (approved):

samlesu (approved):
LGTM

PR #374, opened 2021-01-21
Put register node in subprocess

Kelvin-M:
## Description
In case of multichannel, if one channel has a issue, the other can not be used. With that PR, it's now possible ! 

## Closes issue(s)
None

## Companion PRs
None

## How to test / repro
Launch substra e2e test :) 

## Screenshots / Trace
None

## Changes include

## Checklist
- [x] I have tested this code

## Other comments
None

### DIFF
--- a/backend/libs/health_check_middleware.py
+++ b/backend/libs/health_check_middleware.py
@@ -20,20 +20,21 @@ class HealthCheckMiddleware(object):
         """
         Returns that the server is alive.
         """
-        validate_solo_channels()
+        validate_channels()
         return HttpResponse("OK")
 
     def readiness(self, request):
         """
         Returns that the server is alive.
         """
-        validate_solo_channels()
+        validate_channels()
         return HttpResponse("OK")
 
 
-def validate_solo_channels():
+def validate_channels():
+    # Check ledger connection for each channel
     for channel_name, channel in settings.LEDGER_CHANNELS.items():
-        if channel_name.startswith('solo-') or channel['restricted']:
-            with get_hfc(channel_name) as (loop, client, user):
-                # get_hfc will throw if the solo channel has more than 1 member
-                pass
+        with get_hfc(channel_name) as (loop, client, user):
+            # if channel_name starts with 'solo-' channel name is include channel['restricted']
+            # get_hfc will throw if the solo channel has more than 1 member
+            pass
--- a/backend/node-register/apps.py
+++ b/backend/node-register/apps.py
@@ -1,5 +1,6 @@
 import time
 import logging
+import multiprocessing
 from django.apps import AppConfig
 from django.conf import settings
 from substrapp.ledger.api import invoke_ledger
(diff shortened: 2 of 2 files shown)


inalgnu (approved):
👍

mblottiere (approved):

AurelienGasser:
An issue I see with this change is that the pod readiness will be true regardless of the whether or not the connection to the ledger is successful. That's because the pod gets marked as ready as soon as the function returns: with this PR the function returns immediately.

As a result, the pod will be marked as ready, even though potentially it's not. 

One of the potential nefarious impacts is in the context of end to end tests: we use the readiness probe to ensure the backend is ready. So if the connection to the ledger fails, the backend will be marked as ready, the tests will start (and fail). 

What about adding a mechanism to ensure all the ledger connections are functional before marking the pod as ready?

Kelvin-M:
> An issue I see with this change is that the pod readiness will be true regardless of the whether or not the connection to the ledger is successful. That's because the pod gets marked as ready as soon as the function returns: with this PR the function returns immediately.

Yes, you're right ! 

> As a result, the pod will be marked as ready, even though potentially it's not.

When the backend-server pod is ready, that means that the backend is ready to serve request. It's true whether the ledger is ready or not, even if the request will surely return an error because of a ledger connection failure.

> One of the potential nefarious impacts is in the context of end to end tests: we use the readiness probe to ensure the backend is ready. So if the connection to the ledger fails, the backend will be marked as ready, the tests will start (and fail).

Yes

> What about adding a mechanism to ensure all the ledger connections are functional before marking the pod as ready?

This what we want to avoid, because, in case of multi-channels, if one ledger connection fail, all ledger connection are not usable.
Instead, we can make the backend ready if, at least, one ledger connection is functional or we can set a mandatory list of ledger connections to be functional to make the backend ready ! What do you think ?

AurelienGasser:
@Kelvin-M 

> we can make the backend ready if, at least, one ledger connection is functional or we can set a mandatory list of ledger connections to be functional to make the backend ready 

I think setting a mandatory list of connections to be functional would be best. But isn't that already the case today? Today, if the backend is non-functional because one of the channels is non-functional, then all you have to do to fix it is change the helm values and restrict the number of configured channels.

For instance if the values say

```yaml
channels: ['mychannel', 'otherchannel']
```

and `otherchannel` isn't functional, all you have to do is change the config to 

```yaml
channels: ['mychannel']
```

In other words, the solution you describe ("set a mandatory list of ledger connections") is already implemented, and I fail to see how adding a new list of channels in the config would solve any problems. What am I missing?

Kelvin-M:
You're missing solo channel :)

AurelienGasser:
@Kelvin-M If I understand correctly, you'd want to configure the backend to connect to all the channels, but only make the solo channel mandatory. Is that correct? If so, I think introducing  a separate "mandatory" list of channels is reasonable. In this case the list would only contain solo channel(s).

Kelvin-M:
To Do : add a health route for hlf connection !

Kelvin-M:
I've updated the health check to do a `get_hfc` on each channel :)
As we use the e2e run script with `skaffold` and the option `--status-check=true` we do not need to do extra modification !

Kelvin-M:
E2E Tests ✔️  (there is also an init container for substra-test pod which curl-wait readiness url of both backends before starting)

PR #375, opened 2021-01-25
Improve watch pod to handle failed or evicted pod

### LINKED ISSUE #305: Evicted training tasks pods are never deleted
When a training task pod is evicted by kubernetes, the backend waits forever for the pod to either complete or fail.

However, since "Evicted" is not a failure condition that is currently supported in the backend:

- the pod stays in `Evicted` state forever
- the backend waits (for the pod to compelte) forever
- the training tasks (and possibly compute plan) stay in the `doing` state forever.

Kelvin-M:
## Description
Improve `watch_pod` to handle failed or evicted pod.
 
## Closes issue(s)
Fix #305 

## Companion PRs
None

## How to test / repro
None

## Screenshots / Trace
None

## Changes include

## Checklist
- [x] I have tested this code

## Other comments
None

### DIFF
--- a/backend/substrapp/tasks/k8s_backend.py
+++ b/backend/substrapp/tasks/k8s_backend.py
@@ -105,9 +105,8 @@ def watch_pod(name, watch_init_container=False):
             if api_response.status.phase == 'Failed' or (api_response.status.reason and
                'Evicted' in api_response.status.reason):
                 error_pod = f'Pod failed reason {api_response.status.reason})'
-                attempt += 1
-                logger.error(f'Status for pod "{name}" {api_response.status.phase.lower()} status '
-                             f'(attempt {attempt}/{max_attempts}): {api_response.status.reason}')
+                logger.error(f'Status for pod "{name}" {api_response.status.phase.lower()} status')
+                finished = True
 
             if watch_init_container:
                 if api_response.status.init_container_statuses:
--- a/backend/substrapp/tasks/k8s_backend.py
+++ b/backend/substrapp/tasks/k8s_backend.py
@@ -88,7 +88,6 @@ def watch_pod(name, watch_init_container=False):
     attempt = 0
     max_attempts = 5 + (5 if watch_init_container else 0)
     error = None
-    error_pod = None
     watch_container = not watch_init_container
 
     logger.info(f'Waiting for pod {name}...')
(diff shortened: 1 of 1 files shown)


backend/substrapp/tasks/k8s_backend.py
```diff
@@ -100,6 +101,14 @@ def watch_pod(name, watch_init_container=False):
                 pretty=True
             )
 
+            # Handle pod error not linked with containers
+            if api_response.status.phase == 'Failed' or (api_response.status.reason and
+               'Evicted' in api_response.status.reason):
+                error_pod = f'Pod failed reason {api_response.status.reason})'
+                attempt += 1
```
AurelienGasser, line 108: IMO we should immediately exit the function here because the POD will not recover even after multiple attempts: the time.sleep is 0.2 secs and if I interpret the [doc](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy) correctly the first restart of the pod will happen after 10 sec.

Kelvin-M: Yes, thank you for digging in this documentation !

backend/substrapp/tasks/k8s_backend.py
```diff
@@ -100,6 +101,13 @@ def watch_pod(name, watch_init_container=False):
                 pretty=True
             )
 
+            # Handle pod error not linked with containers
+            if api_response.status.phase == 'Failed' or (api_response.status.reason and
+               'Evicted' in api_response.status.reason):
+                error_pod = f'Pod failed reason {api_response.status.reason})'
+                logger.error(f'Status for pod "{name}" {api_response.status.phase.lower()} status')
+                finished = True
```
samlesu, line 109: Does it make sense to continue the execution inside the while loop if `finished` is set to true? At first sight it seems we would like to directly exit the while loop.

backend/substrapp/tasks/k8s_backend.py
```diff
@@ -144,6 +152,13 @@ def watch_pod(name, watch_init_container=False):
             attempt += 1
             logger.error(f'Could not get pod "{name}" status (attempt {attempt}/{max_attempts}): {e}')
 
+    # Merge error
+    if error_pod is not None:
```
samlesu, line 156: What's the reason for not setting the `error` variable inside the `if api_response.status.phase == 'Failed' or (api_response.status.reason and 'Evicted' in api_response.status.reason):` condition ?

To do that, maybe we could move the if condition to the end of code inside the while loop?

backend/substrapp/tasks/k8s_backend.py
Kelvin-M: Yes, you're right we can early exit with a break or continue :) (as we set finished to true)

backend/substrapp/tasks/k8s_backend.py
Kelvin-M: I change the code to be simpler, so it's not necessary to have this merge code :)

samlesu (approved):
LGTM

@Kelvin-M It would be great to wait for another approval before merging.

backend/substrapp/tasks/k8s_backend.py
samlesu: Thanks!

backend/substrapp/tasks/k8s_backend.py
samlesu: Thanks!

AurelienGasser (approved):

PR #382, opened 2021-02-09
Fix default service value for docker-registry

Kelvin-M:
Fix issue when deploying with default value


ClementGautier (commented):
I'd also add a note on the README saying that the pullDomain can't point to a local registry and that you either need an external one OR use node port for local connection

ClementGautier (approved):

inalgnu (approved):

PR #386, opened 2021-03-08
[chart] expose FLUSH_REFRESH_EXPIRED_TOKENS_TASK_PERIOD setting

mblottiere:
This patch exposes a new configuration option to adapt the period of expired token purge.

It defaults to 24h to match application's behavior when the env var is not set.

### DIFF
--- a/charts/substra-backend/CHANGELOG.md
+++ b/charts/substra-backend/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2.0.1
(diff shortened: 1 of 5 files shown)


Kelvin-M:
Do you have any more improvements for this PR ?

Kelvin-M (changes requested):
You need to bump the chart version :)

Kelvin-M (approved):

PR #391, opened 2021-03-12
fix: auto-allocate docker-registry node port

AurelienGasser:
This PR fixes the error 

```
Error: Service "backend-org-1-docker-registry" is invalid: spec.ports[0].nodePort: Invalid value: 32001: provided port is already allocated
```

This error can happen in two scenarios:

1. The application gets deployed, then destroyed, then deployed a second time. In that scenario, the kubernetes API [doesn't provide a way](https://community.apigee.com/questions/78206/apigee-hybrid-invalid-value-31381-provided-port-is.html?childToView=78225#answer-78225) to tell when a port has been fully released and is free for use again.
2. In CI tests on GKE, the clusters we create occasionally have the port 32001 unavailable. We suspect it's due to kube-system services using auto-assigned ports, which can randomly conflict with ports 3200x.

This PR fixes the issue in both scenarios by letting kubernetes choose a port number for the docker-registry (auto-allocated port). We then programmatically (from the application) retrieve the allocated port number using the kubernetes API.


AurelienGasser:
end-to-end tests: :heavy_check_mark:

Kelvin-M (approved):
LGTM !

inalgnu (approved):

PR #399, opened 2021-04-12
feat: add local docker-registry cleanup

### LINKED ISSUE #398: Add an auto-cleanup of old docker-registry images
As the storage of the docker-registry is not infinite, we need to add a periodic job/pod that cleanup old images time to time.

Kelvin-M:
Fix #398 

Another idea : Add a periodic task to do the same thing directly in django-celery!

### DIFF
--- a/charts/substra-backend/Chart.yaml
+++ b/charts/substra-backend/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: substra-backend
 home: https://substra.org/
-version: 2.0.2
+version: 2.0.3
 description: Main package for Substra
 icon: https://avatars1.githubusercontent.com/u/38098422?s=200&v=4
 sources:
--- a/charts/substra-backend/README.md
+++ b/charts/substra-backend/README.md
@@ -106,7 +106,8 @@ The following table lists the configurable parameters of the substra-backend cha
 | `docker-registry.persistence.size` | Amount of space to claim for PVC | `10Gi` |
 | `docker-registry.persistence.deleteEnabled` | Enable the deletion of image blobs and manifests by digest | `true` |
 | `docker-registry.service.type` | service type (If you use the local docker registry, you must use a `NodePort` to expose it kubernetes) | `NodePort` |
-| `docker-registry.service.nodePort` | if `docker-registry.service.type` is `NodePort` and this is non-empty, sets the node port of the service | (undefined)` |
+| `docker-registry.service.nodePort` | if `docker-registry.service.type` is `NodePort` and this is non-empty, sets the node port of the service | (undefined) |
+| `docker-registry.images.duration` | Maximum image duration before removal in seconds| `604800` |
 | `celerybeat.replicaCount` | Replica count for celerybeat service | `1` |
 | `celerybeat.expiredTokensFlushPeriod` | Flushing of expired tokens task period | `86400` |
 | `celerybeat.taskPeriod` | Celery beat task period | `10800` |
--- a/charts/substra-backend/templates/docker-registry-cleanup-operator.yaml
+++ b/charts/substra-backend/templates/docker-registry-cleanup-operator.yaml
@@ -1,0 +1,68 @@
+ {{- if index .Values "docker-registry" "enabled" }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "substra.fullname" . }}-registry-cleanup
+  labels:
+    {{ include "substra.labels" . | nindent 4 }}
+    app.kubernetes.io/name: {{ template "substra.name" . }}-registry-cleanup
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+        app.kubernetes.io/name: {{ template "substra.name" . }}-registry-cleanup
+        {{ include "substra.selectorLabels" . | nindent 8}}
+  template:
+    metadata:
+      labels:
+        {{ include "substra.labels" . | nindent 8 }}
+        app.kubernetes.io/name: {{ template "substra.name" . }}-registry-cleanup
+    spec:
+      serviceAccountName: {{ template "substra.fullname" . }}-registry-cleanup
+      containers:
+      - name: {{ index .Values "docker-registry" "clean-images" "image" "repository" }}
+        image: "{{ index .Values "docker-registry" "clean-images" "image" "repository" }}:{{ index .Values "docker-registry" "clean-images" "image" "tag" }}"
+        imagePullPolicy: "{{ .Values.backend.image.pullPolicy }}"
+        command: ["/bin/sh", "-c"]
+        args:
+        - |
+          # Install dependencies
+          echo "Install dependencies"
+          apk add jq curl
+          curl -sLO https://storage.googleapis.com/kubernetes-release/release/v1.16.6/bin/linux/amd64/kubectl && chmod +x ./kubectl && mv ./kubectl /bin
+
+          while true; do
+
+            REGISTRY_URL={{ .Release.Name }}-docker-registry:5000
+
+            # List images with a tag defined
+            echo "Fetch images mdetada from ${REGISTRY_URL}"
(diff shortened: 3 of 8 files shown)


charts/substra-backend/templates/docker-registry-cleanup-operator.yaml
```diff
@@ -0,0 +1,68 @@
...
+          while true; do
+
+            REGISTRY_URL={{ .Release.Name }}-docker-registry:5000
+
+            # List images with a tag defined
+            echo "Fetch images mdetada from ${REGISTRY_URL}"
+            IMAGES=$(curl -s ${REGISTRY_URL}/v2/_catalog | jq -r .repositories[] | xargs -I '{}' curl -s -H "Accept: application/vnd.docker.distribution.manifest.v2+json" ${REGISTRY_URL}/v2/'{}'/tags/list | jq -r 'select(.tags | index("substra")) | .name')
+
+            for IMAGE_NAME in ${IMAGES}; do
+                # Test if the image is older than MAXIMUM_IMAGES_DURATION
```
AlexandrePicosson, line 43:
```suggestion
                # Test if the image is older than MAXIMUM_IMAGES_DURATION seconds
```

charts/substra-backend/README.md
```diff
@@ -106,7 +106,8 @@ The following table lists the configurable parameters of the substra-backend cha
 | `docker-registry.persistence.size` | Amount of space to claim for PVC | `10Gi` |
 | `docker-registry.persistence.deleteEnabled` | Enable the deletion of image blobs and manifests by digest | `true` |
 | `docker-registry.service.type` | service type (If you use the local docker registry, you must use a `NodePort` to expose it kubernetes) | `NodePort` |
-| `docker-registry.service.nodePort` | if `docker-registry.service.type` is `NodePort` and this is non-empty, sets the node port of the service | (undefined)` |
+| `docker-registry.service.nodePort` | if `docker-registry.service.type` is `NodePort` and this is non-empty, sets the node port of the service | (undefined) |
+| `docker-registry.images.duration` | Maximum image duration before removal in seconds| `604800` |
```
AlexandrePicosson, line 110: Here could you add a small note to the last column with the equivalent duration in a more human readable format ? for example ``` `604800` (7 days) ```. This make it possible to go through the default values quickly without spending time converting it.

charts/substra-backend/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: substra-backend
 home: https://substra.org/
-version: 2.0.2
+version: 2.0.3
```
AurelienGasser, line 4: IMO it's worth creating a minor version here: `2.1.0`

charts/substra-backend/README.md
AurelienGasser: `ttl` might be more explicit than `duration`?

charts/substra-backend/templates/docker-registry-cleanup-operator.yaml
```diff
@@ -0,0 +1,68 @@
...
+        args:
+        - |
+          # Install dependencies
+          echo "Install dependencies"
+          apk add jq curl
+          curl -sLO https://storage.googleapis.com/kubernetes-release/release/v1.16.6/bin/linux/amd64/kubectl && chmod +x ./kubectl && mv ./kubectl /bin
+
+          while true; do
+
+            REGISTRY_URL={{ .Release.Name }}-docker-registry:5000
```
AurelienGasser, line 36: this might be cleaner as an env var

charts/substra-backend/templates/docker-registry-cleanup-operator.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+            # List images with a tag defined
+            echo "Fetch images mdetada from ${REGISTRY_URL}"
+            IMAGES=$(curl -s ${REGISTRY_URL}/v2/_catalog | jq -r .repositories[] | xargs -I '{}' curl -s -H "Accept: application/vnd.docker.distribution.manifest.v2+json" ${REGISTRY_URL}/v2/'{}'/tags/list | jq -r 'select(.tags | index("substra")) | .name')
+
+            for IMAGE_NAME in ${IMAGES}; do
+                # Test if the image is older than MAXIMUM_IMAGES_DURATION
+                IS_OLDER_THAN_MAXIMUM_IMAGES_DURATION=$(curl -s ${REGISTRY_URL}/v2/${IMAGE_NAME}/manifests/substra | jq -r '.history[].v1Compatibility' | jq -s 'sort_by(.created)[-1].created' | jq -r --arg MAXIMUM_IMAGES_DURATION "$MAXIMUM_IMAGES_DURATION" 'sub("\\.[^.]*"; "Z") | fromdate | . < (now | . - ($MAXIMUM_IMAGES_DURATION| tonumber))')
```
AurelienGasser, line 44: This is quite hard to read IMO. Not sure, but maybe doing the date comparison in bash would be cleaner.

charts/substra-backend/templates/rbac.yaml
```diff
@@ -58,3 +58,45 @@ roleRef:
...
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "substra.fullname" . }}-registry-cleanup
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{ include "substra.labels" . | nindent 4 }}
+    app.kubernetes.io/name: {{ template "substra.name" . }}-registry-cleanup
+---
+kind: ClusterRole
```
AurelienGasser, line 73: did you try with `Role`?

values/backend-org-2.yaml
```diff
@@ -5,6 +5,8 @@ rabbitmq:
 docker-registry:
   service:
     type: NodePort
+  clean-images:
+    duration: 3600
```
AurelienGasser, line 9: IMO 1h is a bit short, I often use the same image for >1h. What about 24h ?

AurelienGasser (approved):
Nice work!

AurelienGasser:
In scenarios where we don't have internet access (like Melloddy for instance), we don't want to *ever* delete the base images (`substrafoundation/substra-tools`). Maybe it would be worth introducing an "allow list" of images who are exempt from deletion.

charts/substra-backend/templates/docker-registry-cleanup-operator.yaml
```diff
@@ -0,0 +1,68 @@
...
+                    DIGEST=$(curl --head -s -H "Accept: application/vnd.docker.distribution.manifest.v2+json" -X GET ${REGISTRY_URL}/v2/${IMAGE_NAME}/manifests/substra | grep Docker-Content-Digest | awk '{print $2}')
+
+                    echo "Delete image : ${IMAGE_NAME}:substra ($DIGEST)"
+                    curl -H "Accept: application/vnd.docker.distribution.manifest.v2+json" -X DELETE ${REGISTRY_URL}/v2/${IMAGE_NAME}/manifests/${DIGEST}
+                fi
+            done
+
+            # Garbage collector
+            echo "Run docker-registry garbage-collector"
+            kubectl exec -n {{ .Release.Namespace }} $(kubectl get pods -n {{ .Release.Namespace }} -l app=docker-registry -o jsonpath='{.items[0].metadata.name}') -- /bin/registry garbage-collect /etc/docker/registry/config.yml | tail -n 1
```
ClementGautier, line 58: Having to rely on `kubectl exec` looks like an anti-pattern here. Would it be possible to move that piece of code in a sidecar-container to the docker-registry pod directly ? (by contributing to the sub-chart)

ClementGautier (changes requested):

charts/substra-backend/templates/docker-registry-cleanup-operator.yaml
Kelvin-M: Impossible, the garbage collector should be called inside the docker-registry. There are no remote api :'(

PR #400, opened 2021-04-14
feat: add registry cleaning tasks

Kelvin-M:
Similar to #399 but in python
Add celery docker-registry cleaning tasks + fix celery beat and scheduler
Fix https://github.com/SubstraFoundation/substra-backend/issues/398

### DIFF
--- a/backend/substrapp/tasks/k8s_backend.py
+++ b/backend/substrapp/tasks/k8s_backend.py
@@ -459,6 +459,7 @@ def k8s_remove_image(image_name):
 
 
 def get_image_list_from_docker_registry():
+
     response = requests.get(
         f'{REGISTRY_SCHEME}://{REGISTRY}/v2/_catalog',
         headers={'Accept': 'application/vnd.docker.distribution.manifest.v2+json'},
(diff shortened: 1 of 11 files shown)


backend/substrapp/tasks/utils.py
```diff
@@ -149,3 +150,15 @@ def wrapper(*args, **kwargs):
...
+def fetch_old_algo_image_names(max_duration):
+
+    logger.info(f'Fetch image names older than {max_duration}')
+
+    return k8s_fetch_old_algo_image_names(max_duration)
+
+
+def garbage_collector():
+    logger.info('Launch garbage collect on docker-registry')
+    k8s_docker_registry_garbage_collector()
```
AurelienGasser, line 164: Is there an advantage to having these two methods?

backend/substrapp/tasks/k8s_backend.py
```diff
@@ -446,6 +448,102 @@ def k8s_remove_image(image_name):
...
+        lines = ""
+        resp.update(timeout=1)
+        if resp.peek_stdout():
+            lines = resp.read_stdout()
+        if resp.peek_stderr():
+            lines = resp.read_stderr()
+        for line in filter(None, lines.split("\n")):
+            logs.append(line)
+    else:
+        logger.info(logs[-1])
```
AurelienGasser, line 536: Note that if stdout and stderr are written to within the same second (and in this case only), the corresponding lines might not be in the right order (stdout might be before stderr even though stderr was printed before).

The way to fix this is to use tty=True instead, and to redirect stderr to stdout `2>&1`

backend/substrapp/tasks/k8s_backend.py
```diff
@@ -446,6 +448,102 @@ def k8s_remove_image(image_name):
         logger.exception(e)
 
 
+def k8s_get_image_list():
```
AurelienGasser, line 451: Are you sure the k8s prefixes are necessary?

backend/substrapp/tasks/tasks.py
```diff
@@ -1100,3 +1101,15 @@ def on_compute_plan(channel_name, compute_plan):
...
+@app.task(ignore_result=True)
+def clean_old_images_task(max_duration):
+    algo_image_names = fetch_old_algo_image_names(max_duration)
+    for algo_image_name in algo_image_names:
+        remove_image(algo_image_name)
+
+
+@app.task(ignore_result=True)
+def garbage_collector_task():
+    garbage_collector()
```
AurelienGasser, line 1115: Make it more clear what this is?

```suggestion
def docker_registry_garbage_collector_task():
    docker_registry_garbage_collector()
```

backend/substrapp/tasks/utils.py
Kelvin-M: What do you mean by that ?

AurelienGasser: These functions are just wrappers: they do nothing but write a log and call another function. Maybe just write the log from the wrapped function and delete the wrapper?

Kelvin-M: Done

backend/substrapp/tasks/k8s_backend.py
```diff
@@ -446,6 +448,103 @@ def k8s_remove_image(image_name):
...
+
+        response.raise_for_status()
+        images.append(response.json())
+
+    return images
+
+
+def fetch_old_algo_image_names_from_docker_registry(max_duration):
+
+    logger.info(f'Fetch image names older than {max_duration}')
```
mblottiere, line 477:
```suggestion
    logger.info(f'Fetch image names older than {max_duration}s')
```

mblottiere (approved):

charts/substra-backend/values.yaml
```diff
@@ -203,6 +203,7 @@ docker-registry:
 celerybeat:
   replicaCount: 1
   expiredTokensFlushPeriod: 86400
+  maximumImagesDuration: 86400
```
AurelienGasser, line 206: See [this comment](https://github.com/SubstraFoundation/substra-backend/pull/399#discussion_r612190882)

AlexandrePicosson, line 206: There is no entry in the chart README for this new value.

AurelienGasser (approved):
nice work!

AlexandrePicosson (approved):

inalgnu (approved):

charts/substra-backend/values.yaml
```diff
@@ -203,6 +203,7 @@ docker-registry:
 celerybeat:
   replicaCount: 1
   expiredTokensFlushPeriod: 86400
+  maximumImagesTTL: 86400
```
mblottiere, line 206: This is not the same as the default in celery.py (7 * 24 * 3600 = 604800), is it on purpose?

Kelvin-M: Yep, I don't know what to put as default value in the chart  (24h vs 7 days)

mblottiere: 7 days seems more conservative, that would be a sound default.

Kelvin-M: Done

inalgnu (approved):

PR #435, opened 2022-09-06
refactor: use task outputs

mblottiere:
## Rationale

With generic tasks, we remove the hardcoded per-task outputs.
The compute engine was still processing outputs based on task category though.

The goal of this PR is to leverage outputs instead of task category.

I tested locally, a test failed about image building but this seems unrelated.

## Design notes

OutputResources are managed by the context and combine data from
both task outputs (permissions, transient flag, etc) and algo outputs
(asset kind, multiple flag). They encapsulate path computation as well.

<s>Models don't have a category anymore, adding a dependency to https://github.com/Substra/orchestrator/pull/8.</s>
For now, model categories are hardcoded and inferred from the identifier. This is a short term tactic to avoid dependency on model category removal.


Owlfred:

backend/substrapp/compute_tasks/outputs.py
```diff
@@ -2,56 +2,108 @@
...
+            directories.commit_dir(
+                self._ctx.directories, directories.TaskDirName.Chainkeys, directories.CPDirName.Chainkeys
+            )
+
+    def _save_performance(self, output: context.OutputResource):
+        logger.info("saving performances")
+        with get_orchestrator_client(self._ctx.channel_name) as client:
+            perf_path = os.path.join(self._ctx.directories.task_dir, output.rel_path)
+            perf = _get_perf(perf_path)
+            _register_perf(client, self._ctx.task.key, self._ctx.algo.key, perf, output.identifier)
```
AurelienGasser, line 47: Nit: inline the `_register_perf` call

backend/substrapp/compute_tasks/outputs.py
```diff
@@ -85,3 +137,14 @@ def _get_perf(perf_path: str) -> float:
...
+
+def _get_model_category(identifier: str) -> str:
+    """Temporarily hardcode a match between model identifier and category.
+
+    This is based on the following assumptions and should be removed once we drop model categories:
+    - all models are SIMPLE
+    - except "local" or "head" outputs
+    """
+
+    return model_pb2.MODEL_HEAD if identifier.lower() in ["local", "head"] else model_pb2.MODEL_SIMPLE
```
AurelienGasser, line 150: Can you avoid added hardcoded literals in the function body? e.g. by using top-level constants

```py
OUTPUT_IDENTIFIER_LOCAL = "local"
```

or by refactoring with 

https://github.com/Substra/substra-backend/blob/a780df98e907376378607e4bcb5f8d286c9f5507/backend/localrep/serializers/computetask.py#L36-L41

AurelienGasser (commented):
Great work! It might be worth running the substrafl e2e tests too.

backend/substrapp/compute_tasks/context.py
```diff
@@ -142,16 +143,49 @@ def data_sample_keys(self) -> list[str]:
...
+        print(self._outputs)
+        print(path)
+        # TODO: specific exception
+        raise InvalidContextError("output not found")
+
+    def _get_output_path(self, kind: orchestrator.AssetKind, identifier: str) -> str:
+        if kind == orchestrator.AssetKind.ASSET_MODEL:
+            return os.path.join(TaskDirName.OutModels, f"{identifier}.model")
+        elif kind == orchestrator.AssetKind.ASSET_PERFORMANCE:
+            filename = "-".join([self.algo.key, "perf.json"])
```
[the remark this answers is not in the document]
mblottiere: I just moved the [existing logic](https://github.com/Substra/substra-backend/blob/a780df98e907376378607e4bcb5f8d286c9f5507/backend/substrapp/compute_tasks/command.py#L32-L41).
But we can indeed drop the algo name and use the identifier instead, that will allow a task to output several performances.

backend/substrapp/compute_tasks/outputs.py
mblottiere: The refactor you propose is not behaving the same way, I extracted the identifiers in a constant.

backend/substrapp/compute_tasks/context.py
```diff
@@ -142,16 +143,49 @@ def data_sample_keys(self) -> list[str]:
...
-            path = os.path.relpath(output["value"], SANDBOX_DIR)
-            self._outputs[path] = output["id"]
+        for output in self._outputs:
+            if output.rel_path == path:
+                return output
+
+        print(self._outputs)
+        print(path)
+        # TODO: specific exception
+        raise InvalidContextError("output not found")
```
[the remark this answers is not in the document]
mblottiere: I wish we could have constant message and additional context in exceptions, but that would mean looking seriously at the logger and I prefer to keep that out of this PR.

backend/substrapp/compute_tasks/context.py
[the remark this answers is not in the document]
mblottiere: After a second look, let's keep that unchanged and revisit once the metric task is not a special case anymore.

Owlfred:

mblottiere:
Substrafl tests failed because of the missing substratool image (should be fixed after next release).

backend/substrapp/compute_tasks/context.py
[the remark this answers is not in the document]
AurelienGasser: Is there a specific reason to wait?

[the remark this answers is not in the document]
mblottiere: Done

Owlfred:

backend/substrapp/tests/compute_tasks/test_command.py
```diff
@@ -335,6 +373,14 @@ def test_get_args_test_after_predict():
         compute_plan=None,
         directories=Directories(task.compute_plan_key),
     )
+    ctx._outputs = [
+        context.OutputResource(
+            identifier=InputIdentifiers.PERFORMANCE,
+            kind=orchestrator.AssetKind.ASSET_PERFORMANCE,
+            multiple=False,
+            rel_path=f"perf/{algo.key}-perf.json",
```
AurelienGasser, line 381:
```suggestion
            rel_path=f"perf/{InputIdentifiers.PERFORMANCE}-perf.json",
```

AurelienGasser (dismissed):
Nice work!

backend/orchestrator/resources.py
```diff
@@ -125,11 +125,21 @@ def from_grpc(cls, i: algo_pb2.AlgoInput) -> "AlgoInput":
         return cls(kind=AssetKind.from_grpc(i.kind), multiple=i.multiple, optional=i.optional)
 
 
+class AlgoOutput(pydantic.BaseModel):
+    kind: AssetKind
+    multiple: bool
+
+    @classmethod
+    def from_grpc(cls, o: algo_pb2.AlgoOutput) -> "AlgoOutput":
```
Kelvin-M, line 133: Since python 3.7+, `from __future__ import annotations` can be used to return `AlgoOutput` class as type annotation instead of the string `"AlgoOutput"`. As we don't want have compatibility with python 3.6, it can be a nice improvement.
A good explanation can be found [here](https://stackoverflow.com/a/33533514)
It can be done in another pull request.

backend/substrapp/compute_tasks/outputs.py
```diff
@@ -2,74 +2,116 @@
...
+        except Exception as exc:
+            for model in local_models:
+                self._delete_model(model["key"])
+            raise exc
+
+        for localrep_data in localrep_data_models:
+            localrep_data["channel"] = self._ctx.channel_name
+            localrep_serializer = ModelRepSerializer(data=localrep_data)
+            try:
+                localrep_serializer.save_if_not_exists()
```
Kelvin-M, line 77: I don't know if it should be done in this pull request but it should be better to use a REST to register the model in substra-backend API.
You will surely have to create the post api in the model view to do that as it does not exist. 
It will remove the dependency between the Django database and the compute engine.

For instance, it has been done to register task profiling : 
 - post request :  https://github.com/Substra/substra-backend/blob/main/backend/substrapp/tasks/tasks_compute_task.py#L194-L207
 - post api :  https://github.com/Substra/substra-backend/blob/main/backend/api/views/task_profiling.py#L45-L50

mblottiere: Yes, this is definitely where we want to go, but not in this PR.

backend/orchestrator/resources.py
mblottiere: Good point, I'll do it here. This will all be replaced by `Self` once we switch to 3.11 🎉

Owlfred:

Kelvin-M (approved):
LGTM !

PR #438, opened 2022-09-07
chore: remove root user from dev profile

sergebouchut2:
There is an unexpected side effect when using root user with the `dev` skaffold profile.
The volumes are mounted with root ownership but then the user used in compute pod hasn't the permission to write inside shared volumes.


AlexandrePicosson (dismissed):
LGTM

AlexandrePicosson (approved):

PR #439, opened 2022-09-08
feat: add a new filter key on the model view

AlexandrePicosson:
## Description

Introduce a new filter param `compute_task_key` that makes it possible to filter models by the compute tasks that produced them.

## Motivation and Context

This change is required because I need to filter models on the key of the compute tasks that produced them. This was introduced in the SDK in https://github.com/Substra/substra/pull/275 and I thought it would work with the filter `compute_task` but it was not the case since this filter only accept a single key and the SDK only sends lists for filters.

## How Has This Been Tested?

Using the Django API tool and using the SDK.
Let me know if I need to introduce a unit test for this, I did not do it since the `compute_task` filter was already untested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - BREAKING: allow registration of users with role from the API and reset password mechanism
 - Add address and permissions for `inputs` of kind datamanager and model in compute_tasks api response
 - Synchronize compute task output assets into localrep
+- `compute_task_key` filter on the model view.
 
 ## [0.28.0] 2022-08-29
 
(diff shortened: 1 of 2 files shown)


sergebouchut2:
> I thought it would work with the filter compute_task but it was not the case since this filter only accept a single key and the SDK only sends lists for filters.

@AlexandrePicosson I am not sure to correctly understand the issue:
- Do you need to filter models on multiple compute task keys? If so could you provide more information on the context in which you need that?
- Or there is an SDK limitation you try to bypass by a backend hack?

For your information, we plan to create endpoints to retrieve inputs and outputs for a specific task. Could this answer you need?
cc: @acellard

AlexandrePicosson:
Well it's mainly 2 as the SDK considers that all filters are lists. But I don't see this as a hack but a limitation of the current model API since all other APIs can be filtered using lists of properties.

It could answer my needs but as I am specifically looking for models I don't really need an outputs API. If we decide to wait for this then the transient outputs feature is blocked until this API is available and I can rewrite my end to end test.

sergebouchut2 (approved):

PR #441, opened 2022-09-08
chore: move substrapp views into localrep

sergebouchut2:
Move api views into localrep (which will be renamed api later).
The goal is to remove api purpose code from substrapp and let only code shared by api and compute engine.
Substrapp should not depend on localrep anymore: localrep will depend on substrapp instead.
The focus of this PR is made on views modules (+ views utils and views tests), but some piece of code could remain in substrapp and not used by neither compute engine nor substrapp model / serializer. If so we should move them later progressively. 

---
  - https://app.asana.com/0/0/1201507164505658
  - https://app.asana.com/0/0/1202951134496591


Owlfred:

cboillet-dev:
Currently there is still a `localrep` dependency in the `save_models` method of the compute task. I wonder if we could instead call and api endpoint to perform the local update to complete seperation between the Localrep-API and compute Engine. That could of course be handled as a seperate PR

sergebouchut2:
> Currently there is still a localrep dependency in the save_models method of the compute task. I wonder if we could instead call and api endpoint to perform the local update to complete seperation between the Localrep-API and compute Engine. That could of course be handled as a seperate PR

Yes, I have noticed that too. This will need to be handled in a dedicated PR one the compute engine side. I create an Asana [card](https://app.asana.com/0/1201599852028640/1202951505377087).

AlexandrePicosson (approved):

cboillet-dev (commented):
Thank for that good work @sergebouchut2

PR #454, opened 2022-09-13
refactor: orchestrator client return info object

AlexandrePicosson:
## Description

Create a new object `OrchestratorVersion`.

## Motivation and Context

This is the same as the change done in previous PRs to switch the outputs from the orchestrator from a dict to an object.

## How Has This Been Tested?

I tested the change locally by accessing `/info` and everything seems to work.


AlexandrePicosson:
On a side note the test is located in the api app `backend/api/tests/views/test_views_info.py` but the view is in `backend/backend`.

mblottiere (approved):

PR #458, opened 2022-09-13
feat: compute task input assets

sergebouchut2:
Like for task output, add a task input asset table with resolved inputs.|
- If the task input has an asset key, the input asset is immediately resolved during the task registration.
- Otherwise when an output asset is created, we check if the task output is referenced by a task input. If so we create the input asset.
With this, the API will be able to query the input assets in a straightforward way without worrying of the algo input kind and / or the task output parent. In the same way we do for output asset, see https://github.com/Substra/substra-backend/pull/451.


backend/api/serializers/computetask.py
```diff
@@ -61,31 +88,34 @@ class Meta:
...
+    def to_representation(self, data):
+        data = super().to_representation(data)
+        assets = data.pop("assets")
+        if len(assets) == 0:
+            data["value"] = None
+        elif len(assets) == 1:
+            data["value"] = assets[0]
+        else:
+            data["value"] = assets
+        return data
```
sergebouchut2, line 106: Side change: keep the same behavior but isolate the custom logic

backend/api/management/commands/generate_fixtures.py
```diff
@@ -12,10 +12,30 @@ class Command(BaseCommand):
...
-        simple_algo = factory.create_algo(name="Simple algo", category=Algo.Category.ALGO_SIMPLE)
-        aggregate_algo = factory.create_algo(name="Aggregate algo", category=Algo.Category.ALGO_AGGREGATE)
-        composite_algo = factory.create_algo(name="Composite algo", category=Algo.Category.ALGO_COMPOSITE)
-        metric_algo = factory.create_algo(name="Metric algo", category=Algo.Category.ALGO_METRIC)
+        simple_algo = factory.create_algo(
+            inputs=factory.get_algo_inputs(["datasamples", "opener", "model"]),
+            outputs=factory.get_algo_outputs(["model"]),
+            category=Algo.Category.ALGO_SIMPLE,
+            name="simple algo",
+        )
```
sergebouchut2, line 20: Side improvement:
Algo inputs/outputs are provided to the factory instead of being inferred. It removes adherence between test factory  and category.

backend/api/tests/asset_factory.py
```diff
@@ -324,19 +383,24 @@ def create_computetask(
             TaskDataSamples.objects.create(compute_task_id=key, data_sample_id=data_sample, order=order)
         compute_task.refresh_from_db()
 
-    for position, input in enumerate(compute_task.algo.inputs.all().order_by("identifier")):
-        ComputeTaskInput.objects.create(
-            task=compute_task,
-            identifier=input.identifier,
-            asset_key=INPUT_ASSET_KEY,
```
sergebouchut2, line 331: This constant made the test data invalid. This issue was raised with the new `ComputeTaskInputAsset` which requires valid asset keys.

backend/api/serializers/computetask.py
```diff
@@ -258,17 +282,27 @@ def create(self, validated_data):
...
         for position, input in enumerate(inputs):
-            ComputeTaskInput.objects.create(
+            task_input = ComputeTaskInput.objects.create(
                 channel=compute_task.channel,
                 task=compute_task,
                 position=position,
                 **input,
             )
+            # task input asset could be known during task registration
+            # or could be resolved later if it does not exist yet
```
mblottiere, line 298: There is a catch here: if you add a task with already DONE parents, we already have all the inputs.
It's probably better to only resolve links at query time in this context (we can afford the additional cost since inputs should eventually be a task subresource).

sergebouchut2: Indeed this issue is catched by the unit tests.

sergebouchut2: I am not sure this invalidate having `ComputeTaskInputAsset` though.

backend/api/serializers/computetask.py
```diff
@@ -258,17 +285,44 @@ def create(self, validated_data):
...
+                )
+            elif task_input.parent_task_key:
+                task_output = ComputeTaskOutput.objects.get(
+                    task_id=task_input.parent_task_key,
+                    identifier=task_input.identifier,
+                )
+                try:
+                    task_output_asset = task_output.assets.filter(
+                        asset_kind=task_input.parent_task_output_identifier,
+                    )[position]
```
sergebouchut2, line 317: It seems that we used to get the first asset (see [here](https://github.com/Substra/substra-backend/pull/458/files#diff-9edf1f25f3fa8d5bae29bf35657c7d05f03fb33b6985d7560153d319fc7fc0b2L354))

Owlfred:

Owlfred:

Owlfred:

Owlfred:

sergebouchut2:
Last rebase is only to integrate the fix for the CI https://github.com/Substra/substra-backend/pull/493/ and should not impact e2e tests.

acellard (approved):
LGTM. Great improvement! Thanks

sergebouchut2:
Added data migration in the last push.

backend/api/serializers/computetask.py
```diff
@@ -251,17 +277,41 @@ def create(self, validated_data):
...
+                    task_input=task_input,
+                    asset_key=task_input.asset_key,
+                    asset_kind=input_kinds[task_input.identifier],
+                )
+            elif task_input.parent_task_key:
+                task_output = ComputeTaskOutput.objects.get(
+                    task=task_input.parent_task_key,
+                    identifier=task_input.parent_task_output_identifier,
+                )
+                # FIXME: we should use the position instead of the first asset
```
mblottiere, line 306: We do not support multiple output assets yet, and the final implementation might not rely on positions.
IMO the comment should read:
```suggestion
                # This only supports a single asset per output for now
```

backend/api/serializers/computetask.py
mblottiere: I think I misunderstood the flow: this is in the `create` method, so during processing of the task event right? What about resources registered afterward, how are `ComputeTaskInput` objects created for them? Or is this only happening when building the API response?

sergebouchut2: This should be covered by the compute task output asset [sync](https://github.com/Substra/substra-backend/pull/458/files#diff-50b50fd37d33a5998bcd7a185d86f4a84641f351a3f9c3198adbdd5035e4b141R234).

sergebouchut2:
Rebase to fix conficts in test module.

backend/api/serializers/computetask.py
mblottiere: Thanks for the clarification, that's the part I missed 👍

mblottiere (approved):

PR #466, opened 2022-09-20
feat: add channel for input output

sergebouchut2:
## Description

Following #451
Systematically provide channel name on all assets to avoid error when retrieving or listing outputs. 

## Checklist


backend/api/models/computetask.py
```diff
@@ -30,19 +32,17 @@ class ComputeTaskOutput(models.Model):
...
 
 
 class ComputeTaskOutputAsset(models.Model):
-    class Kind(models.TextChoices):
-        ASSET_MODEL = common_pb2.AssetKind.Name(common_pb2.ASSET_MODEL)
-        ASSET_PERFORMANCE = common_pb2.AssetKind.Name(common_pb2.ASSET_PERFORMANCE)
-
     task_output = models.ForeignKey("ComputeTaskOutput", on_delete=models.CASCADE, related_name="assets")
-    asset_kind = models.CharField(max_length=64, choices=Kind.choices)
+    asset_kind = models.CharField(max_length=64, choices=AlgoOutput.Kind.choices)
```
sergebouchut2, line 43: Side improvement: we should use the algo output kind list here instead of redefining a list with the same values.

sergebouchut2: (No change in migrations as the values are exactly the same by definition)

PR #468, opened 2022-09-20
feat: prevent use of '__' in asset metadata keys

acellard:
## Description

Related [asana task](https://app.asana.com/0/1200346939311555/1202768358328705)

Full details in [orc PR](https://github.com/Substra/orchestrator/pull/33)
As the orchestrator does not allow metadata key with `__`, this PR renames the tag metadata to `_tag_` for all new registered tasks while still supporting `__tag__` for already registered tasks.

### Companion PRs
- https://github.com/Substra/orchestrator/pull/33
- https://github.com/Substra/substra/pull/289
- https://github.com/Substra/substra-tests/pull/212

## How has this been tested?

Full e2e tests ran on Substra standalone instance locally

## Checklist

### DIFF
--- a/backend/api/tests/views/test_views_algo.py
+++ b/backend/api/tests/views/test_views_algo.py
@@ -525,6 +525,10 @@ class AlgoViewTests(APITestCase):
         response = self.client.post(self.url, data={}, format="json")
         self.assertEqual(response.status_code, status.HTTP_500_INTERNAL_SERVER_ERROR)
 
+    # def test_algo_create_fail(self, _):
+    #     response = self.client.post(self.url, data={}, format="json")
+    #     self.assertEqual(response.status_code, status.HTTP_500_INTERNAL_SERVER_ERROR)
+
     def test_algo_retrieve(self):
         url = reverse("api:algo-detail", args=[self.expected_algos[0]["key"]])
         response = self.client.get(url, **self.extra)
--- a/backend/api/tests/views/test_views_algo.py
+++ b/backend/api/tests/views/test_views_algo.py
@@ -525,10 +525,6 @@ class AlgoViewTests(APITestCase):
         response = self.client.post(self.url, data={}, format="json")
         self.assertEqual(response.status_code, status.HTTP_500_INTERNAL_SERVER_ERROR)
 
-    # def test_algo_create_fail(self, _):
-    #     response = self.client.post(self.url, data={}, format="json")
-    #     self.assertEqual(response.status_code, status.HTTP_500_INTERNAL_SERVER_ERROR)
-
     def test_algo_retrieve(self):
         url = reverse("api:algo-detail", args=[self.expected_algos[0]["key"]])
         response = self.client.get(url, **self.extra)
--- a/backend/api/views/computetask.py
+++ b/backend/api/views/computetask.py
@@ -32,6 +32,7 @@ from api.views.utils import TASK_CATEGORY
 from api.views.utils import ApiResponse
 from api.views.utils import get_channel_name
 from api.views.utils import validate_key
+from api.views.utils import validate_metadata
 from libs.pagination import DefaultPageNumberPagination
 from orchestrator import computetask
 from substrapp.orchestrator import get_orchestrator_client
@@ -91,6 +92,7 @@ def _register_in_orchestrator(tasks_data, channel_name):
         if "__tag__" in orc_task["metadata"]:
             raise Exception('"__tag__" cannot be used as a metadata key')
         else:
+            validate_metadata(orc_task["metadata"])
             orc_task["metadata"]["__tag__"] = task_data.get("tag") or ""
 
         extra_data_field = EXTRA_DATA_FIELD[orc_task["category"]]
--- a/backend/api/views/computetask.py
+++ b/backend/api/views/computetask.py
@@ -35,6 +35,7 @@ from api.views.utils import validate_key
 from api.views.utils import validate_metadata
 from libs.pagination import DefaultPageNumberPagination
 from orchestrator import computetask
+from orchestrator.resources import TAG_KEY
 from substrapp.orchestrator import get_orchestrator_client
 
 logger = structlog.get_logger(__name__)
@@ -89,11 +90,8 @@ def _register_in_orchestrator(tasks_data, channel_name):
             "metadata": task_data.get("metadata") or {},
         }
 
-        if "__tag__" in orc_task["metadata"]:
-            raise Exception('"__tag__" cannot be used as a metadata key')
-        else:
-            validate_metadata(orc_task["metadata"])
-            orc_task["metadata"]["__tag__"] = task_data.get("tag") or ""
+        validate_metadata(orc_task["metadata"])
+        orc_task["metadata"][TAG_KEY] = task_data.get("tag") or ""
 
         extra_data_field = EXTRA_DATA_FIELD[orc_task["category"]]
         orc_task[extra_data_field] = _compute_extra_data(orc_task, task_data)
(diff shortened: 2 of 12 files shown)


backend/api/views/computetask.py
```diff
@@ -91,6 +92,7 @@ def _register_in_orchestrator(tasks_data, channel_name):
         if "__tag__" in orc_task["metadata"]:
             raise Exception('"__tag__" cannot be used as a metadata key')
         else:
+            validate_metadata(orc_task["metadata"])
             orc_task["metadata"]["__tag__"] = task_data.get("tag") or ""
```
AlexandrePicosson, line 96: Here you check for `__` in metadata and add a `__tag__` metadata right after, is it expected ?

acellard: I've wondered the same thing but from what I saw, this __tag__ as metadata is only in the orchestrator, tag is a field in localrep. So it should be fine since the issue is only in Django.
I'm not sure 100% of this though, I will check further to be sure this tag is never propagated as metadata outside the orchestrator.

AlexandrePicosson: It makes sense. Something I'm also not comfortable with is re-introducing validation outside of the localrep serializers. If we want this as a rule the two places I would like to see this is in the orchestrator validators for metadata or in the api serializers. (cc @sergebouchut2 @mblottiere)

acellard: I ruled out doing it in orchestrator because of this `__tag__`, but doing it in the backend it makes totally sense to use the serializer. You're right, thanks for raising the point.
So there is 2 options:
- Option 1 : Use orchestrator validators for metadata, we would need to use another name for the tag than `__tag__`
- Option 2 : Move this check in the serializer (as I should have done right away actually, my bad).

I would favor option 2 as it is a very Django-like issue, but not a strong opinion. I will wait for input from @sergebouchut2 and @mblottiere then before moving forward.

Thanks again!

AlexandrePicosson: The main issue I can see with option 2 is that the asset will be created in the orchestrator but then rejected by the backend.

mblottiere: There's no specific check on the orchestrator about `__tag__`, it's treated as any other metadata and IMO should stay that way.

acellard: Would it be possible to rename it in the orc and use `_tag_` instead from now on ? We'd need a data migration. We could then do the check that prevents the use of `__` in metadata key, in the orchestrator validators.

mblottiere: There is nothing to rename in the orchestrator since the orc doesn't know anything about `__tag__`: this is handled by the backend here: https://github.com/Substra/substra-backend/blob/cb66135ddc9f0c2e3ec45542ffee91d0027c0aa8/backend/orchestrator/client.py#L50-L51

This can be changed, just in mind that this can also impact the SDK/substrafl.

Esadruhn: As you say in the SDK `tag` is its own field, so I don't see any impact

mblottiere: > Thanks for your input, that's good news. We'd still need a data migration on orc side though to avoid dealing with both `__tag__` and `_tag_` in the backend when extracting the tag, right ?

Not sure this should be a data migration (as in, part of the orc): this doesn't concern a field the orc is processing but purely user data. Since filtering is done on API models and not orc data, this should not impact orc data.

acellard: Check moved to orchestrator in https://github.com/Substra/orchestrator/pull/33
This PR was updated accordingly.

backend/api/tests/views/test_views_algo.py
```diff
@@ -525,6 +525,10 @@ def test_algo_create_fail(self, _):
         response = self.client.post(self.url, data={}, format="json")
         self.assertEqual(response.status_code, status.HTTP_500_INTERNAL_SERVER_ERROR)
 
+    # def test_algo_create_fail(self, _):
```
mblottiere, line 528: Can this be uncommented?

acellard: Totally removed, this was not relevant anymore. Thanks for spotting it!

acellard:
/e2e --ref orchestrator=fix/metadata_filtering,substra-tests=fix/metadata_filtering,substra=fix/metadata_filtering

Owlfred:
Something happened while processing your message: error: unknown option '--ref'
(Did you mean --refs?)

acellard:
/e2e --refs orchestrator=fix/metadata_filtering,substra-tests=fix/metadata_filtering,substra=fix/metadata_filtering

Owlfred:

acellard:
/e2e --refs orchestrator=fix/metadata_filtering,substra-tests=fix/metadata_filtering,substra=fix/metadata_filtering

Owlfred:

Owlfred:
```
Usage: /e2e [options] [help]

/e2e may appear anywhere as long as it is on its own line

Options:
  --refs <value>                                         Extra refs (branch or tag) with format REPO=GIT_REF,REPO=GIT_REF.
  Supported repositories: substra-hlf-k8s, orchestrator, substra-backend, substra-frontend, substra-tools, substrafl, substra, substra-tests, connect-ci.
  Example: /e2e --refs substra-backend=some_branch,orchestrator=some_tag (default: {})
  --tests-to-run, --tests <tests-to-run>                 Comma-separated list of tests to run. Valid options: sdk,substrafl,frontend or NONE. (default: "sdk")
  --benchmarks-to-run, --benchmarks <benchmarks-to-run>  Comma-separated list of workflows tests to run. Valid options: mnist, camelyon or NONE. (default: "NONE")
  --orchestrator-mode, --mode <orchestrator-mode>        Comma-separated list of orchestrator modes to run tests for. Valid options: standalone,distributed (default: "standalone,distributed")
  -h, --help                                             display help for command
```

acellard:
/e2e --refs orchestrator=fix/metadata_filtering,substra-tests=fix/metadata_filtering,substra=fix/metadata_filtering --mode standalone

Owlfred:

acellard:
/e2e --refs orchestrator=fix/metadata_filtering,substra=fix/metadata_filtering --mode standalone

Owlfred:

backend/api/views/computetask.py
```diff
@@ -91,6 +92,7 @@ def _register_in_orchestrator(tasks_data, channel_name):
         if "__tag__" in orc_task["metadata"]:
             raise Exception('"__tag__" cannot be used as a metadata key')
```
mblottiere, line 93: This becomes redundant with the `__` constraint

acellard: Not if we don't rename it, which seems not necessary since the orchestrator can deal with __ (that's why the metadata check is done before the backend adds this` __tag__`). Should we rename it to _tag_ still ? The backend will have to deal with both `__tag__` and `_tag_` (assets before this change and after).

mblottiere: I'd leave it as is since the backend won't have to deal with before/after value.
Also, maybe move the `__tag__` in a constant to get a better visibility over setters/getters.

Then I still think this check is unnecessary, event without renaming as `orc_task["metadata"] == task_data.get("metadata") or {}`, this comes from the `task_data` (line 89) which is checked line 93.
So this will reject any metadata containing `__`, including `__tag__`. Then the `__tag__` value is added *after* the check on line 96, which is fine.

acellard: Oh yes my bad, I answered too quiclky, this is definitely redundant here.

acellard: unnecessary check remove and `__tag__` moved to a constant

mblottiere (approved):

PR #471, opened 2022-09-21
test: provide algo inputs outputs to factory

sergebouchut2:
## Description

Provides inputs/outputs to the algo and compute task factory, instead of inferring static values from category.
- It removes dependency to the categories we want to deprecate.
- It provides valid task input keys instead of random uuid which is needed to test serializers (required for #458).

Before

```python
algo = create_algo(
    category=Algo.Category.ALGO_SIMPLE,
)
train_task = create_computetask(
    compute_plan,
    algo,
    category=ComputeTask.Category.TASK_TRAIN,
    ...
)

metric = create_algo(
    category=Algo.Category.ALGO_METRIC,
)
test_task = create_computetask(
    compute_plan,
    metric,
    category=ComputeTask.Category.TASK_TEST,
    ...
)
```

After:

```python
algo = create_algo(
    inputs=factory.build_algo_inputs(["datasamples", "opener"]),
    outputs=factory.build_algo_outputs(["model"]),
)
train_task = create_computetask(
    compute_plan,
    algo,
    inputs=factory.build_computetask_inputs(
        algo,
        {
            "opener": [data_manager.key],
            "datasamples": [data_sample.key],
        },
    ),
    outputs=factory.build_computetask_outputs(algo),
    ...
)

metric = create_algo(
    inputs=factory.build_algo_inputs(["datasamples", "opener", "model"]),
    outputs=factory.build_algo_outputs(["performance"]),
)
test_task = create_computetask(
    compute_plan,
    metric,
    inputs=factory.build_computetask_inputs(
        metric,
        {
            "opener": [data_manager.key],
            "datasamples": [data_sample.key],
            "model": [train_task.key],
        },
    ),
    outputs=factory.build_computetask_outputs(metric),
    ...
)
```

## How has this been tested?

E2e not needed as only test modules are changed.

## Checklist

---
  - https://app.asana.com/0/0/1203055469585066

### DIFF
--- a/backend/api/tests/views/test_views_algo.py
+++ b/backend/api/tests/views/test_views_algo.py
@@ -37,18 +37,42 @@ class AlgoViewTests(APITestCase):
     def setUp(self):
         if not os.path.exists(MEDIA_ROOT):
             os.makedirs(MEDIA_ROOT)
-        self.maxDiff = None
         self.extra = {"HTTP_SUBSTRA_CHANNEL_NAME": "mychannel", "HTTP_ACCEPT": "application/json;version=0.0"}
         self.logger = logging.getLogger("django.request")
         self.previous_level = self.logger.getEffectiveLevel()
         self.logger.setLevel(logging.ERROR)
         self.url = reverse("api:algo-list")
 
-        simple_algo = factory.create_algo(category=Algo.Category.ALGO_SIMPLE, name="simple algo")
-        aggregate_algo = factory.create_algo(category=Algo.Category.ALGO_AGGREGATE, name="aggregate")
-        composite_algo = factory.create_algo(category=Algo.Category.ALGO_COMPOSITE, name="composite")
-        predict_algo = factory.create_algo(category=Algo.Category.ALGO_PREDICT, name="predict")
-        metric_algo = factory.create_algo(category=Algo.Category.ALGO_METRIC, name="metric")
+        simple_algo = factory.create_algo(
+            inputs=factory.build_algo_inputs(["datasamples", "opener", "model"]),
+            outputs=factory.build_algo_outputs(["model"]),
+            category=Algo.Category.ALGO_SIMPLE,
+            name="simple algo",
+        )
+        aggregate_algo = factory.create_algo(
+            inputs=factory.build_algo_inputs(["model"]),
+            outputs=factory.build_algo_outputs(["model"]),
+            category=Algo.Category.ALGO_AGGREGATE,
+            name="aggregate",
+        )
+        composite_algo = factory.create_algo(
+            inputs=factory.build_algo_inputs(["datasamples", "opener", "local", "shared"]),
+            outputs=factory.build_algo_outputs(["local", "shared"]),
+            category=Algo.Category.ALGO_COMPOSITE,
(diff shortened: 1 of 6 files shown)


backend/api/tests/views/test_views_computetask.py
```diff
@@ -168,17 +184,43 @@ def mock_register_compute_task(orc_request):
...
+            "permissions": self.data_manager_data["permissions"],
+        }
+        model_input = {
+            "identifier": "model",
+            "asset_key": str(self.done_train_task.key),
+            "parent_task_key": None,
+            "parent_task_output_identifier": None,
+            # FIXME: "addressable": self.train_model_data["address"],
+            # FIXME: "permissions": self.train_model_data["permissions"],
+        }
```
sergebouchut2, line 209: @acellard I think the current response for CREATE request is buggy: `addressable` and `permissions` are not always as expected. It may be related to the current implementation using "if request", see [here](https://github.com/Substra/substra-backend/blob/main/backend/api/serializers/computetask.py#L353) and [here](https://github.com/Substra/substra-backend/blob/main/backend/api/serializers/computetask.py#L366). But it should be fixed in the next [PR](https://github.com/Substra/substra-backend/pull/458).

backend/api/tests/views/test_views_algo.py
```diff
@@ -77,8 +101,14 @@ def setUp(self):
...
-                "inputs": factory.ALGO_INPUTS_PER_CATEGORY["ALGO_SIMPLE"],
-                "outputs": factory.ALGO_OUTPUTS_PER_CATEGORY["ALGO_SIMPLE"],
+                "inputs": {
+                    "datasamples": {"kind": "ASSET_DATA_SAMPLE", "optional": False, "multiple": True},
+                    "opener": {"kind": "ASSET_DATA_MANAGER", "optional": False, "multiple": False},
+                    "model": {"kind": "ASSET_MODEL", "optional": True, "multiple": False},
+                },
+                "outputs": {
+                    "model": {"kind": "ASSET_MODEL", "multiple": False},
+                },
```
sergebouchut2, line 111: Even if it's a bit more verbose, it's better to have an explicit overview of the response we expect IMO. It helps to see the impact of future changes.

acellard (approved):
Thanks a lot for this PR! It will also ease frontend dev process as we usually don't use e2e tests and up-to-date data dumps are not always available. We rely a lot on these simulated data, so the more realistic and correct they are, the easier it is to test.

backend/api/tests/views/test_views_performance.py
```diff
@@ -51,7 +55,14 @@ def setUp(self):
             )
             for i in range(3)
         ]
-        self.performances = [factory.create_performance(self.compute_tasks[i], self.metric) for i in range(3)]
+        self.performances = [
+            factory.create_performance(
+                self.compute_tasks[i],
+                self.metric,
+                identifier="performance",
```
cboillet-dev, line 62: Should we use const for the `performance` string ?

sergebouchut2: I am not sure because identifiers are free text values

PR #480, opened 2022-09-23
chore: add migration test step

cboillet-dev:
## Description

Test forward migrations in the CI between the main branch and any new pull request. The script is inspired by the manual test workflow as described in the [`doc`](https://github.com/Substra/substra-backend/blob/main/docs/test_migrations.md)

This aim to discover migration issues earlier that upon performing substra upgrade with substra-generator as such: https://owkin.slack.com/archives/C02JQ8DN7R8/p1653308666645529

Asana [card](https://app.asana.com/0/1201599852028640/1202320710297939)

## Checklist

- [ x] [changelog](../CHANGELOG.md) was updated with notable changes

When a migration file is added in the PR the github action [pass successfully](https://github.com/Substra/substra-backend/actions/runs/3142053150)

### DIFF
--- a/.github/workflows/migration.yml
+++ b/.github/workflows/migration.yml
@@ -1,0 +1,46 @@
+name: "Test migrations"
+
+on:  
+  pull_request:
+    branches:
+      - "**"
+
+env:
+  DJANGO_SETTINGS_MODULE: backend.settings.localdev
+
+jobs:
+  test-migration:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: main
+      - name: Set up python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - uses: actions/cache@v3.0.8
+        id: cache
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+      - name: Install requirements
+        run: |
+          python -m pip install --upgrade pip
+          pip install --upgrade --upgrade-strategy eager -r backend/dev-requirements.txt
+      - name: Run initial migration and feed data
+        run: |
+          make db
+          cd backend
+          sleep 10
+          python manage.py check --database default 
+          python manage.py migrate
+          python manage.py generate_fixtures
+      - name: Checkout local branch
+        uses: actions/checkout@v3
+      - name: Test migration
+        run: |
+          python backend/manage.py migrate
+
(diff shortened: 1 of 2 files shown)


samlesu:
Thanks @cboillet-dev, any idea on how long it takes to run this action?

.github/workflows/migration.yml
```diff
@@ -0,0 +1,46 @@
+name: "Test migrations"
+
+on:  
+  pull_request:
+    branches:
+      - "**"
```
AlexandrePicosson, line 6: Maybe this could be scoped to run only on PRs that changes the content of the migrations folder, on other PR there is no reason to check this.

.github/workflows/migration.yml
```diff
@@ -0,0 +1,46 @@
...
+            ${{ runner.os }}-pip-
+      - name: Install requirements
+        run: |
+          python -m pip install --upgrade pip
+          pip install --upgrade --upgrade-strategy eager -r backend/dev-requirements.txt
+      - name: Run initial migration and feed data
+        run: |
+          make db
+          cd backend
+          sleep 10
```
AlexandrePicosson, line 37: Is a sleep enough ? Is there another way to check if the service is ready that would "never fail".

cboillet-dev:
@samlesu thanks, from the last run it has been taking [1min44](https://github.com/Substra/substra-backend/actions/runs/3113743150)

.github/workflows/migration.yml
cboillet-dev: I agree that would be a good further optimization. I started this PR to first implement the testing procedure

.github/workflows/migration.yml
cboillet-dev: Best would be to wait for the db to be ready before executing the migration. However I have tested such procedure and the command was never returning and therefore the action idling. Hence the hardcoded sleep value

sergebouchut2: Solved by using a proper service for db (like we already do for unit test).

sergebouchut2 (approved):
So great to have this :heart:

.github/workflows/migration.yml
```diff
@@ -0,0 +1,62 @@
...
+        run: |
+          cd backend
+          python manage.py migrate
+          python manage.py generate_fixtures
+          echo fixture generated
+      - name: Checkout local branch
+        uses: actions/checkout@v3
+      - name: Test migration
+        run: |
+          python backend/manage.py migrate
```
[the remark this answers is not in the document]
sergebouchut2: Reinstall dependencies on the feature branch should solve the issue, isn't it?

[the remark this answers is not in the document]
mblottiere: Yes 👍

PR #484, opened 2022-09-26
fix: Allow to set CSRF_TRUSTED_ORIGINS

mblottiere:
## Description

The default value was not compatible with a locally deployed frontend.
See https://github.com/adamchainz/django-cors-headers#csrf-integration

This change is quite minimal: the setting is created from env var.

## How has this been tested?

Locally, with this change I'm able to login through the frontend.

## Checklist

### DIFF
--- a/backend/backend/settings/common.py
+++ b/backend/backend/settings/common.py
@@ -160,6 +160,7 @@ WSGI_APPLICATION = "backend.wsgi.application"
 SECURE_CONTENT_TYPE_NOSNIFF = True
 SESSION_COOKIE_HTTPONLY = True
 CSRF_COOKIE_HTTPONLY = True
+# @CSRF_TRUSTED_ORIGINS: A list of origins that are allowed to use unsafe HTTP methods
 CSRF_TRUSTED_ORIGINS = json.loads(os.environ.get("CSRF_TRUSTED_ORIGINS", "[]"))
 
 # Database
--- a/backend/backend/settings/common.py
+++ b/backend/backend/settings/common.py
@@ -160,6 +160,7 @@ WSGI_APPLICATION = "backend.wsgi.application"
 SECURE_CONTENT_TYPE_NOSNIFF = True
 SESSION_COOKIE_HTTPONLY = True
 CSRF_COOKIE_HTTPONLY = True
+CSRF_TRUSTED_ORIGINS = json.loads(os.environ.get("CSRF_TRUSTED_ORIGINS", "[]"))
 
 # Database
 # https://docs.djangoproject.com/en/1.9/ref/settings/#databases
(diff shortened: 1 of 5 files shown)


AlexandrePicosson:
Do you know why it was working fine before and not anymore ?

mblottiere:
> Do you know why it was working fine before and not anymore ?

No idea, I did not run the frontend locally in a long time 🤷

AlexandrePicosson (dismissed):

backend/backend/settings/common.py
```diff
@@ -160,6 +160,7 @@ def build_broker_url(user: str, password: str, host: str, port: str) -> str:
 SECURE_CONTENT_TYPE_NOSNIFF = True
 SESSION_COOKIE_HTTPONLY = True
 CSRF_COOKIE_HTTPONLY = True
+CSRF_TRUSTED_ORIGINS = json.loads(os.environ.get("CSRF_TRUSTED_ORIGINS", "[]"))
```
AlexandrePicosson, line 163: Maybe you could add a comment here to detail what is expected as a value, typical things the user should put in this list ?

AlexandrePicosson: Example: https://github.com/Substra/substra-backend/blob/fa637e3a74576eb45218102a3fcba5ea0af0a4f2/backend/backend/settings/deps/cors.py#L13

AlexandrePicosson, line 163: This setting should be set in `substra-backend/backend/backend/settings/deps/cors.py` with the other settings of the `django-cors-header` package.

AlexandrePicosson: But it is related to the dependency, no ?

AlexandrePicosson: Ah no, ok your link of the doc was quite confusing I thought they described config values of their module.

sergebouchut2:
I can login from the frontend with or without this change.

Milouu:
> I can login from the frontend with or without this change.

It's a difficult to reproduce bug because i think it only appears when one of your cookies (signature i think) is expired and maybe some other edge cases, but it does exist and we've encountered it several times in the last weeks.

docs/settings.md
```diff
@@ -30,6 +30,7 @@ Accepted true values for `bool` are: `1`, `ON`, `On`, `on`, `T`, `t`, `TRUE`, `T
 | string | `COMPUTE_POD_RUN_AS_GROUP` | nil |  |
 | string | `COMPUTE_POD_RUN_AS_USER` | nil |  |
 | int | `COMPUTE_POD_STARTUP_TIMEOUT_SECONDS` | `300` |  |
+| json | `CSRF_TRUSTED_ORIGINS` | `[]` |  |
```
AlexandrePicosson, line 33: You need to re-render the doc

sergebouchut2 (approved):
Even if I can't reproduce `CSRF_TRUSTED_ORIGINS` looks like a good addition.

cboillet-dev (approved):

backend/backend/settings/common.py
mblottiere: Done, doc refreshed

PR #486, opened 2022-09-27
feat(workflow_graph): extract input/outputs from db

alexisdeh:
~This PR is draft: Waiting for PR #471 to update the way the test data is built~
[Edit] : PR #471 has been merged, test data generation updated accordingly.

The response of /workflow_graph endpoint was still hardcoded, depending on the task categories. Now they are extracted from db. This means actual input/outputs are now returned. This is also a preparatory work for removing task categories.

Note: This PR follows [this one](https://github.com/Substra/substra-backend/pull/477). It is to be merged afterwards.

⚠️ This PR also comes with a breaking change. Now we use `identifier` field name instead of `id` to match the backend model, and `inputs_specs` and `outputs_specs` instead of `inputs` and `outputs` respectively.
Before:
```
"tasks": [
        {
            [...]
            "inputs": [
                {
                    "id": "datasamples",
                    "kind": "ASSET_DATA_SAMPLE"
                },
            [...]
```
After:
```
"tasks": [
        {
            [...]
            "inputs_specs": [
                {
                    "identifier": "datasamples",
                    "kind": "ASSET_DATA_SAMPLE"
                },
            [...]
```

Also for edges, `source_output_name` and `target_input_name` have been renamed to `source_output_identifier` and `target_input_identifier` respectively.

Before:
```
"edges": [
                {
                    "source_task_key": str(train_task.key),
                    "target_task_key": str(predict_task.key),
                    "source_output_name": "model",
                    "target_input_name": "model",
                },
             ]
```
After:
```
"edges": [
                {
                    "source_task_key": str(train_task.key),
                    "target_task_key": str(predict_task.key),
                    "source_output_identifier": "model",
                    "target_input_identifier": "model",
                },
             ]
```

Companion PR:
- [frontend](https://github.com/Substra/substra-frontend/pull/118) (Note: frontend needs to be updated as well because 1/ now that all inputs/outputs are returned the size of a task in the graph is bigger, 2/ The "secondary branches" are detected from the output of type "performance" - this later name just changed with this PR because before input/outputs were hard-coded)


backend/api/tests/views/test_views_compute_plan_graph.py
```diff
@@ -71,6 +71,9 @@ def test_cp_graph(self):
         aggregate_task = factory.create_computetask(
             compute_plan, algo=algo, category=ComputeTask.Category.TASK_AGGREGATE, parent_tasks=[composite_task.key]
         )
+        factory.connect_input_output(train_task, "model", predict_task, "model")
```
sergebouchut2, line 74:
```suggestion
        train_algo = factory.create_algo(
            outputs=factory.build_algo_outputs(["model"]),
        )
        train_task = factory.create_computetask(
            compute_plan,
            algo=algo,
            outputs=factory.build_computetask_outputs(train_algo),
            category=ComputeTask.Category.TASK_TRAIN,
        )

        predict_algo = factory.create_algo(
            inputs=factory.build_algo_inputs(["model"]),
        )
        predict_task = factory.create_computetask(
            compute_plan,
            algo=algo,
            inputs=factory.build_computetask_inputs(predict_algo, keys={"model": str(train_task.key)}),
            category=ComputeTask.Category.TASK_PREDICT,
            parent_tasks=[train_task.key],
        )
```

sergebouchut2: This is how we ensure test data are consistent.

backend/api/tests/asset_factory.py
```diff
@@ -346,6 +346,15 @@ def create_computetask(
 
     return compute_task
 
+def connect_input_output(source_task: ComputeTask, source_output_identifier: str, target_task: ComputeTask, target_input_identifier: str):
```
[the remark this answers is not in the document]
alexisdeh: ok for me to use the #471 approach once it is merged 👍

jmorel (commented):
Since we're in the middle of breaking changes, I think it'd make more sense to have `target_input_name` and `source_output_name` renamed as `target_input_identifier` and `source_output_identifier`

backend/api/tests/views/test_views_compute_plan_graph.py
```diff
@@ -51,25 +50,78 @@ def test_too_many_tasks(self):
...
+            rank=1,
+            algo=algo_data_to_model,
+            outputs=factory.build_computetask_outputs(algo_data_to_model),
         )
+
         predict_task = factory.create_computetask(
-            compute_plan, algo=algo, category=ComputeTask.Category.TASK_PREDICT, parent_tasks=[train_task.key]
+            compute_plan,
+            rank=2,
+            algo=algo_model_to_model,
```
sergebouchut2, line 84: As you named the task `predict_task`, maybe you can rename the algo accordingly `predict_algo`.

sergebouchut2: Same for the other algos.

sergebouchut2: But I feel you want to output a prediction, not a model. You can find extensive examples in https://github.com/Substra/substra-backend/blob/main/backend/api/management/commands/generate_fixtures.py

backend/api/views/compute_plan_graph.py
```diff
@@ -18,121 +21,61 @@
...
+
+    inputs = (
+        ComputeTask.objects.filter(key=OuterRef("pk"))
+        .annotate(
+            inputs_specs=JSONBAgg(
+                JsonbBuildObj(Value("identifier"), F("algo__inputs__identifier"), Value("kind"), "algo__inputs__kind")
+            ),
+        )
+        .values("inputs_specs")
+    )
```
sergebouchut2, line 54: Could you explain why you choose to serialize data in the query instead of using nested serializers (using `select_related` or `prefetch_related` in the query)?

acellard: Here we would need 3 new serializer (to avoid serialize useless data, that would come from existing serializers), to build a non-standardized read-only view. The code was simpler and shorter doing it directly in the query, and as it is not a view that return proper task, input or output objects per se I feel serializers does not really bring added-value, for this particular case. 
But it's not a strong opinion, and it's still sometimes fuzzy to me when to absolutely use serializers and when it's not necessary in Django.

backend/api/tests/views/test_views_compute_plan_graph.py
acellard: Actually we wanted to test the different kind of inputs/outputs, that's why the algo have those names. Maybe the task names are misleading and stick too much to the pre-generic task era. I agree that it would probably be predictions but if I'm not mistaken predictions is just an identifier and underlying it's a kind model I didn't see the point to make the difference in the context of this test. Having in mind that today we use meaningful names for those identifiers but it's meant to be free text.

sergebouchut2: Having neat test data is better for human reading but obviously it's a minor point, feel free to ignore it.

acellard: You're right! Test data improved in dc68ec8c541ea4f3cecbd94b4a19fc9ca825850a

backend/api/views/compute_plan_graph.py
acellard: After discussion, the fact that this view cannot be paginated yet and will manipulate a high volume of data, we stick with no serialization, for performance optimization. Comment added to explain the choice in the view.
Thanks Serge!

sergebouchut2 (approved):

jmorel (approved):

Owlfred:

Owlfred:

PR #488, opened 2022-09-27
fix: set ingressClassName correctly on the Ingress resource

AlexandrePicosson:
## Description

The `server.ingress.ingressClassName` value was not used properly by the Chart.
Now if you set the ingress class name in your Values your ingress will have the right class.

## How has this been tested?

N/A

## Checklist

### DIFF
--- a/charts/substra-backend/CHANGELOG.md
+++ b/charts/substra-backend/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 22.1.3
+
+### Fixed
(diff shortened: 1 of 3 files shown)


sergebouchut2 (approved):

cboillet-dev (approved):
thanks for the fix, that's funny it went unoticed until now

AlexandrePicosson:
It went unnoticed since most of the users uses the old mechanism through annotations on the ingress instead of ingressClass.

PR #489, opened 2022-09-27
test: rename factory for files

sergebouchut2:
Signed-off-by: Serge Bouchut <serge.bouchut@owkin.com>

## Description

We currently have an heterogeneous naming for the substrapp model in charge of managing the files.
Replace `data` by `files`, as this suffix is more explicit and already used in [views](https://github.com/Substra/substra-backend/blob/main/backend/api/views/algo.py#L30) and test [utils](https://github.com/Substra/substra-backend/blob/main/backend/api/tests/views/test_utils.py#L21).

## How has this been tested?

E2e tests not needed, as only test modules are changed.

## Checklist

---
  - https://app.asana.com/0/0/1203057265000942


cboillet-dev (approved):
thanks, it would ne worse adding and entry in the changelog to trace back this change even if it is only renaming

sergebouchut2:
> thanks, it would ne worse adding and entry in the changelog to trace back this change even if it is only renaming

As the changelog is intended for the "user" of the repo, I think it's better to not flood it with this kind of changes without impact.

PR #494, opened 2022-09-29
chore: enable redis persistence

thibowk:
## Description

Previously, the Redis data of the compute engine was not stored in a Persistent Volume. Hence, in the event of a Redis outage, tasks could disappear without ever being processed by the workers. **This PR updates the Helm chart so that Redis data are written on a Persistent Volume.**

Moreover, by default, Redis uses snapshot-based persistence (RDB). The compute engine use case requires better durability guarantees (and not necessarily a high throughput). **This PR updates the Redis config, so that [append-only file based persistence is used](https://redis.io/docs/manual/persistence/)**.

_Note: In the development environment, the data is not persisted in a persistent volume, similar to what is done with PostgreSQL data._

## How has this been tested?

- in a local dev environment, by checking that Persistent Volume and PVC are created
- [substra-tests](https://github.com/owkin/substra-ci/actions/runs/3149772938/jobs/5121781586)

## Checklist

- [X] [changelog](../CHANGELOG.md) was updated with notable changes

### DIFF
--- a/charts/substra-backend/CHANGELOG.md
+++ b/charts/substra-backend/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 22.2.0
+
+### Change
+- Enable Redis persistence using append-only file (AOF)
+
 ## 22.1.3
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


Owlfred:

sergebouchut2 (approved):

PR #508, opened 2022-10-03
fix: compute task outputs/inputs valid storage address

sergebouchut2:
Signed-off-by: Serge Bouchut <serge.bouchut@owkin.com>

## Description

Fix storage address for compute task inputs/output.
The value is expected to be overwritten by the child serializer, (see [here](https://github.com/Substra/substra-backend/blob/main/backend/api/serializers/model.py#L40) for instance).
As child serializer is not directly set as an attribute to the parent serializer, the context is not automatically provided under the hood by DRF and we need to set it manually.

## How has this been tested?

Manual check via the DRF UI.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


acellard (approved):
Good catch!

Kelvin-M (approved):

PR #511, opened 2022-10-03
feat: add compute plan creator

alexisdeh:
## Description

This draft PR is a result of october 27th "flash hackathon" involving @HamdyD and @alexisdeh. The goal was to figure out how we could store a link to the user who registered each asset (= the `creator`), and then allow to filter assets created by a given user.

A question was raised about the privacy of this information. More precisely: Should other organizations know which user registered which asset ? This would expose information about the users that belong to an organization, and how they work within the substra channel.
Without definitive answer, **we decided to keep it private to the organization the user belongs to**. This implies that:
- The `creator` cannot be stored in the orchestrator, since the meta-data in orchestrator are visible to the entire channel.
- It is stored only in the local backend database, not registered in the orchestrator so not synchronized with other backends.
  - So the value will be null for assets registered from other backend
  - ⚠️ This proposal relies on the assumption that there is a 1-1 relation between the organization and the backend. This needs to be checked.

- Other note: In the migration we leave the default value to null, so previously existing assets will have a null value. This could be interpreted as an asset registered by a user from another organization, but we believe the `owner` field can be used to make the difference.


backend/api/serializers/computeplan.py
```diff
@@ -24,6 +24,7 @@ class ComputePlanSerializer(serializers.ModelSerializer, SafeSerializerMixin):
     failed_task = FailedTaskSerializer(read_only=True, allow_null=True, required=False, source="*")
     duration = serializers.IntegerField(read_only=True)
     status = serializers.ChoiceField(choices=ComputePlan.Status.choices, read_only=True)
+    creator = serializers.SlugRelatedField(slug_field='user__username')
```
alexisdeh, line 27:
```suggestion
    creator = serializers.SlugRelatedField(slug_field='user__username', read_only=True)
```

backend/api/views/computeplan.py
```diff
@@ -55,17 +55,27 @@ def create(request, get_success_headers):
...
     # Step1: register asset in orchestrator
     compute_plan_data = {
         "key": to_string_uuid(request.data.get("key")),
         "tag": request.data.get("tag"),
         "name": request.data.get("name"),
         "metadata": validate_metadata(request.data.get("metadata")),
+        "creator": request.user
     }
-    api_data = _register_in_orchestrator(compute_plan_data, get_channel_name(request))
+    #api_data = _register_in_orchestrator(compute_plan_data, get_channel_name(request))
```
alexisdeh, line 73: TODO: uncomment this

HamdyD:
We have an issue with the instance that are **not migrated**. For instance, the compute plan already created will be set at `null` and translated in the frontend as `external`, where it might not be the case.

This is the solution we discussed with @acellard :

- **no data migration** : all compute plans created before this PR will have a creator set as `null`. We will have to explain this behaviour to the users of the only live instance ;
- for all new compute plans, if we have a creator we store it;
- If we receive `null` as creator from the orchestrator, we set it as `external` in localrep, which avoid confusion with the compute plan created before this PR, also stored as `null`.

HamdyD:
Missing the test and this PR is good for review.

Owlfred:
```
Usage: /e2e [options] [help]

/e2e may appear anywhere as long as it is on its own line

Options:
  --refs <value>                                         Extra refs (branch or tag) with format REPO=GIT_REF,REPO=GIT_REF.
  Supported repositories: substra-hlf-k8s, orchestrator, substra-backend, substra-frontend, substra-tools, substrafl, substra, substra-tests, substra-ci.
  Example: /e2e --refs substra-backend=some_branch,orchestrator=some_tag (default: {})
  --tests-to-run, --tests <tests-to-run>                 Comma-separated list of tests to run. Valid options: sdk,substrafl,frontend or NONE. (default: "sdk")
  --benchmarks-to-run, --benchmarks <benchmarks-to-run>  Comma-separated list of workflows tests to run. Valid options: mnist, camelyon or NONE. (default: "NONE")
  --orchestrator-mode, --mode <orchestrator-mode>        Comma-separated list of orchestrator modes to run tests for. Valid options: standalone,distributed (default: "standalone,distributed")
  -h, --help                                             display help for command
```

Owlfred:

acellard:
>⚠️ This proposal relies on the assumption that there is a 1-1 relation between the organization and the backend. This needs to be checked.

After discussion during a previous AAF : no need to rely on the assumption of a 1-1 relation between the organization and the **backend** (which is true now but can become false in the future) but on 1-1 relation between the organization and an substra deployment, more precisely, postgres db, and this is and will stay true.

Can you confirm @mblottiere ? 

> So the value will be null for assets registered from other backend

PR was updated, the value will be the reference to a specific Django user "external".

Owlfred:

Owlfred:

Owlfred:

Owlfred:

acellard:
/e2e --refs substra-tests=feat/asset_author,substra=feat/asset_author

Owlfred:

acellard:
/e2e --refs substra-tests=feat/asset_author,substra=feat/asset_author

Owlfred:

backend/backend/settings/common.py
```diff
@@ -422,3 +422,9 @@ def build_broker_url(user: str, password: str, host: str, port: str) -> str:
 
 # To encode unique jwt token generated with reset password request
 RESET_JWT_SIGNATURE_ALGORITHM = "HS256"
+
+# Username of additonnal Django user representing user external to organization
+EXTERNAL_USERNAME = "external"
+
+# Username of additonnal Django user representing deleted user to not break fk references
```
oleobal, line 429: nitpick

```suggestion
# Username of additional Django user representing user external to organization
EXTERNAL_USERNAME = "external"

# Username of additional Django user representing deleted user to not break FK references
```

oleobal, line 429: typo

```suggestion
# Username of additional Django user representing user external to organization
EXTERNAL_USERNAME = "external"

# Username of additional Django user representing deleted user to not break FK references
```

acellard: Thanks! Fixed

guilhem-barthes (approved):
Thanks for your work!

Milouu:
/e2e --refs substra-tests=feat/asset_author,substra=feat/asset_author

PR #522, opened 2022-10-06
feat: store user ui preferences

acellard:
### Description

This PR is a result of October 27th "flash hackathon" involving @Milouu and @acellard. The goal was to figure out how to properly manage user ui preferences in the backend, and provide api to get, retrieve and update them.
For now, such preferences involve columns display list and favorites compute plans. With likely more preferences in the future.
To keep it flexible and leave the responsibility to the frontend, the preferences are stored in a json field.

This PR also does some refactoring to use serializer to validate data.


backend/users/serializers/user.py
```diff
@@ -40,16 +45,68 @@ def validate(self, attrs):
...
+
+    def _update_ui_preferences(self, instance, value):
+        ui_preferences = instance.channel.ui_preferences
+        if value:
+            for key in value:
+                ui_preferences[key] = value[key]
+        return ui_preferences
+
+    def update(self, instance, validated_data):
+        instance.set_password(validated_data.get("password"))
```
jmorel, line 103: Maybe this should only be done if there is a new password?

backend/users/models/user_channel.py
```diff
@@ -16,3 +16,4 @@ class Role(models.TextChoices):
         settings.AUTH_USER_MODEL, parent_link=True, on_delete=models.CASCADE, related_name="channel"
     )
     channel_name = models.CharField(max_length=100)
+    ui_preferences = models.JSONField(null=True)
```
jmorel, line 19: would it be possible to not have null values but a default at `{}` instead?

backend/users/serializers/user.py
```diff
@@ -1,10 +1,14 @@
+from django.conf import settings
 from django.contrib.auth.models import User
+from django.contrib.auth.password_validation import validate_password
+from django.core.exceptions import ValidationError as djangoValidationError
```
jmorel, line 4: missing capital "D" for `DjangoValidationError`

backend/users/serializers/user.py
```diff
@@ -40,16 +44,76 @@ def validate(self, attrs):
...
+        fields = ["role", "channel_name", "ui_preferences"]
+
+    def validate_channel_name(self, value):
+        if value not in settings.LEDGER_CHANNELS:
+            raise serializers.ValidationError({"channel": "Channel does not exist"})
+        return value
+
+    def validate_ui_preferences(self, value):
+        if type(value) is not dict:
+            raise serializers.ValidationError({"UI preferences should be a dict"})
```
jmorel, line 59: Aren't you missing the `ui_preferences` key in `{"ui_preferences": "UI  preferences should be a dict"}`?

backend/users/tests/test_user.py
```diff
@@ -102,22 +103,23 @@ def test_retrieve_user(self):
         response = AuthenticatedClient(channel=self.channel).get(url, **self.extra)
         assert response.status_code == status.HTTP_200_OK
 
+    @parameterized.expand(
+        [
+            ({"role": UserChannel.Role.USER},),
+            ({"password": "newpas$w0rdtestofdrea6S43"},),
+            ({"ui_preferences": {"columns": ["col1", "col2"]}},),
```
jmorel, line 110: Could you add a test on the fact that ui_preferences is partially updated and not fully replaced? That is keys that were not in the request payload are left untouched.

backend/users/views/user.py
```diff
@@ -130,57 +102,59 @@ def get_queryset(self):
...
+
+        return ApiResponse(data=serializer.data, status=status.HTTP_200_OK, headers=self.get_success_headers({}))
 
     @action(methods=["put"], detail=True, permission_classes=[IsSelf])
     def password(self, request, *args, **kwargs):
         """Allows an authenticated user to modify his own password"""
-        instance = self.get_object()
+
+        username = kwargs.get("username")
+        instance = self.user_model.objects.get(username=username)
```
jmorel, line 145: Did you test this with a username like "John Doe"? `self.get_object()` is able to decode the urlencodedentities but kwargs only contains raw values that need to be decoded manually.

jmorel (changes requested):

backend/users/views/user.py
acellard: Updated the test and it seems to work yes. But still reverting to `get_object()`. For context, in certain cases there was an uncatch exception in `get_object()`. But this was fixed here https://github.com/Substra/substra-backend/pull/522/files#diff-fa54e25c9fbe9c2560d0fbef5cb51cb8f37b0c7793d322f409194417453082bbR62 so it will be catch before.

backend/users/tests/test_user.py
acellard: Test added.

backend/users/serializers/user.py
acellard: To keep consistency yes, key added.

backend/users/tests/test_user.py
```diff
@@ -51,7 +52,7 @@ def test_user_create_duplicate(self):
         response = AuthenticatedClient(role=UserChannel.Role.ADMIN, channel=self.channel).post(self.url, data=data)
 
         assert response.status_code == status.HTTP_400_BAD_REQUEST
-        assert response.data.get("message") == "Username already exists"
+        assert response.data.get("username")[0] == "A user with that username already exists."
```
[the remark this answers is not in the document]
acellard: So as last commit, I leveraged django validation handling: 
`HTTP 400 Bad Request` response with data format : `{field: [error list on this field]}`.
Is this convenient for clients ? Or should we catch validation exception and reformat response message to have something like `{message: 'error string'}`.
I feel it's convenient to use what django already does but I don't have a strong opinion on this.

guilhem-barthes:
Thank you for your contribution! Unfortunately, due to organisational and roadmap changes, this item has been deprioritised. We currently have no bandwidth to cope with the code drift between the current code state in the last release and this PR, therefore we are closing it.

PR #524, opened 2022-10-07
feat: improved user creation validation

jmorel:
Signed-off-by: Jérémy Morel <jeremy.morel@owkin.com>

## Description

- Adds validation for username so that the actual creation doesn't fail with an IntegrityError
- Adds validation on empty password before using the configured validators. This is necessary because the zxcvbn validator will raise a list index out of range exception on an empty password string. See https://github.com/dwolfhub/zxcvbn-python/issues/64
- Decodes the url encoded `kwargs.get('username')` so that the username is properly handled (`John Doe` is a valid username).

## How has this been tested?

Manually.

## Checklist

### DIFF
--- a/backend/users/views/user.py
+++ b/backend/users/views/user.py
@@ -1,4 +1,5 @@
 import datetime
+from urllib.parse import unquote
 
 import jwt
 from django.conf import settings
@@ -21,6 +22,7 @@ from rest_framework.filters import OrderingFilter
 from rest_framework.mixins import CreateModelMixin
 from rest_framework.viewsets import GenericViewSet
 
+from api.errors import BadRequestError
 from api.views.filters_utils import MatchFilter
 from api.views.utils import ApiResponse
 from api.views.utils import get_channel_name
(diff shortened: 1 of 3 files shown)


backend/users/views/user.py
```diff
@@ -40,12 +41,23 @@ def _validate_channel(name):
...
         raise ValidationError(err.error_list)
 
 
+def _validate_username(username):
+    user_model = get_user_model()
+    try:
+        user_model.objects.get(username=username)
+        raise BadRequestError("Username already exists")
+    except user_model.DoesNotExist:
+        pass
```
sergebouchut2, line 58:
```suggestion
    if user_model.objects.filter(username=username).exists()
        raise BadRequestError("Username already exists")
```

backend/users/views/user.py
```diff
@@ -124,6 +136,7 @@ def create(self, request, *args, **kwargs):
         role = request.data.get("role")
 
         _validate_channel(channel)
+        _validate_username(username)
```
sergebouchut2, line 139: At the end having a DRF serializer would be nice here. But it could be done later.

sergebouchut2 (approved):

backend/users/views/user.py
jmorel: Nice, I had forgotten about the `exists`method!

backend/users/views/user.py
jmorel: Agreed, but I didn't want to rewrite the whole view.

cboillet-dev:
looking good to me but it would be worse adding a few validation tests in `backend/users/tests/test_user.py`

jmorel:
By popular demand (wink wink @cboillet-dev @sergebouchut2), let there be tests!

cboillet-dev (approved):

PR #540, opened 2022-11-17
fix: Add same expiration date that header_payload to signature cookie

Milouu:
## Description

Access token in Substra is secured by separating the cookie in 2 cookies: header_payload and signature. The header_payload cookie has a 24h expiration date but the signature had none. Because of this once the header_payload expired there was a bug as the signature was still existing which blocked the creation of a new pair of cookies (be it from the refresh token or entering credentials again), which led to the user having to delete the signature cookie manually to resolve the issue.
This fix only give the same expiration date to the signature cookie so that it disappears at the same time as header_payload. If the user still has a refresh token (1 week expiration date), a new header_payload-signature pair will be fetched automatically, otherwise credentials have to be entered anew.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Compute task outputs/inputs valid storage address.
+- Signature cookie expiration date (#540)
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


sergebouchut2:
I am not sure to understand why having different dates results in a bug but I think this was defined as is on purpose. See https://github.com/Substra/substra-backend/blob/main/docs/authentication.md#jwt-cookie

PR #559, opened 2023-01-05
fix: update task profiling and step profiling if already exists

ThibaultFy:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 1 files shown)


Kelvin-M (commented):
LGTM ! 
I think "print" should be removed from unit tests

backend/api/tests/views/test_views_task_profiling.py
```diff
@@ -108,6 +108,17 @@ def test_task_profiling_create_success(self):
             {"count": len(expected_result), "next": None, "previous": None, "results": expected_result},
         )
 
+    def test_already_exist_task_profiling(self):
```
Kelvin-M, line 111: Is this test failing without de serializer modification ?

ThibaultFy: Yes, we began by adding the test

Kelvin-M (approved):
LGTM !

PR #570, opened 2023-01-27
fix(skaffold): create secrets before deploying releases

guilhem-barthes:
Signed-off-by: Guilhem Barthes <guilhem.barthes@owkin.com>

## Description

This move the secrets creation in another module so it can be run before deploying on Helm.

## Checklist

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -1,7 +1,6 @@
 apiVersion: skaffold/v2beta24
 kind: Config
 
-# Remove if you desactivate mTLS
 requires:
   - path: examples/secrets/skaffold.yaml
 
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -1,5 +1,10 @@
 apiVersion: skaffold/v2beta24
 kind: Config
+
+# Remove if you desactivate mTLS
+requires:
+  - path: examples/secrets/skaffold.yaml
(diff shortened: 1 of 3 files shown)


skaffold.yaml
```diff
@@ -54,8 +59,6 @@ deploy:
     manifests:
       - examples/values/serviceAccounts/serviceAccount-org-1.yaml
       - examples/values/serviceAccounts/serviceAccount-org-2.yaml
-      - examples/secrets/secret-orchestrator-tls-org-1-client-pair.yaml
-      - examples/secrets/secret-orchestrator-tls-org-2-client-pair.yaml
```
ThibaultFy, line 58: If we delete these lines, do we still need the concerned yaml files ?

skaffold.yaml
```diff
@@ -1,5 +1,10 @@
 apiVersion: skaffold/v2beta24
 kind: Config
+
+# Remove if you desactivate mTLS
+requires:
+  - path: examples/secrets/skaffold.yaml
+
```
ThibaultFy, line 7: What does that mean ? this in not required if we do not use mtls ? Shouldn't we parametrize it regarding the mtls activation then ?

ThibaultFy (approved):

PR #586, opened 2023-02-06
chore: release 0.35.0

guilhem-barthes:
## Checklist

### DIFF
--- a/charts/substra-backend/CHANGELOG.md
+++ b/charts/substra-backend/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 22.3.1
+
+### Changed
+- Update substra-backend image tag to `0.35.0`
+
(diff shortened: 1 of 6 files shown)


charts/substra-backend/CHANGELOG.md
```diff
@@ -13,7 +23,7 @@
 ## 22.2.3
 
 ### Changed
-- Update substra-backend image tag to `0.34.0`
+- Update substra-backend image tag to `0.35.0`
```
oleobal, line 26: Duplicate of above?

oleobal (changes requested):
I think it looks like there might be an error in the changelog

guilhem-barthes:
> I think it looks like there might be an error in the changelog

Thanks for catching this one! It was a mistake.

PR #596, opened 2023-02-17
chore(helm): default image tag to AppVersion

oleobal:
## Description

This adds a new helm "function" that defaults the image tag to the AppVersion if it is not provided.

The removes one spot where the app version needed to be copied at releases.

This is already in place on the frontend, I'd been meaning to port this to the other repos -- also see https://github.com/Substra/orchestrator/pull/159 which is this but for the orchestrator.

## How has this been tested?

Check it works the same way:
```
git switch main
helm template charts/substra-backend > old-render.yaml
git switch chore-helm/default-appversion
helm template charts/substra-backend > new-render.yaml
diff old-render.yaml new-render.yaml
```

I also tried `helm template charts/substra-backend --set server.image.tag=woop` to check it (and only it) was set the given value.

## Checklist

### DIFF
--- a/charts/substra-backend/CHANGELOG.md
+++ b/charts/substra-backend/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 22.3.4
+
+### Changed
+- Make backend image tags default to the chart `appVersion`, and set it to `null` in the default values
+
 ## 22.3.3
 
 ### Changed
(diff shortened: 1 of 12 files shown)


oleobal:
Checks aren't passing:
 - pylint is failing but that's from main, this PR doesn't touch python code
 - There already is a backend chart 22.3.3 due to patch shenanigans, I'll make sure this is OK with Guilhem before resolving and merging this.

guilhem-barthes (approved):

guilhem-barthes:
Can you try to rebase ?  It should fix the issues you got with the CI

guilhem-barthes:
LGTM!

PR #611, opened 2023-03-08
fix: improve perf loading

guilhem-barthes:
## Companion PR

- https://github.com/Substra/substra-frontend/pull/180

## Description

Try to improve loading in performance page. Two strategies have been put in place:

1. Using prefetch for FunctionOutput and using it in the serializer (instead of making a request for each function on FunctionOutput)
2. Removing `data_samples` from the request, as it doesn't seem to be used in the frontend.

The point 2 is more controversial, as it breaks the current architecture of returning linked data on eac hrequest. However, it implies reaching a n+ 3 level (as the data_samples are ordered on `api_taskdatasample.order`, which is maybe not needed .

## Checklist

### DIFF
--- a/backend/api/tests/views/test_views_performance.py
+++ b/backend/api/tests/views/test_views_performance.py
@@ -351,43 +351,6 @@ class PerformanceViewTests(APITestCase):
         self.assertTrue(status in str(content_list[1]))
 
 
-@pytest.fixture
-def create_compute_task():
-    def _create_compute_task(compute_plan, n_data_sample=4):
-        data_manager = factory.create_datamanager()
-        data_samples = [factory.create_datasample([data_manager]) for _ in range(n_data_sample)]
-        input_keys = {
-            "opener": [data_manager.key],
-            "datasamples": [data_sample.key for data_sample in data_samples],
-        }
-        function = factory.create_function(
-            inputs=factory.build_function_inputs(["datasamples", "opener", "model"]),
-            outputs=factory.build_function_outputs(["model", "performance"]),
-            name="simple function",
-        )
-        return factory.create_computetask(
-            compute_plan,
-            function,
-            inputs=factory.build_computetask_inputs(function, input_keys),
-            outputs=factory.build_computetask_outputs(function),
-            data_manager=data_manager,
-            data_samples=[data_sample.key for data_sample in data_samples],
-            status=ComputeTask.Status.STATUS_DONE,
-        )
-
-    return _create_compute_task
-
-
-@pytest.fixture
-def create_compute_plan(create_compute_task):
-    def _create_compute_plan(n_task=20, n_data_sample=4):
-        compute_plan = factory.create_computeplan()
-        [create_compute_task(compute_plan, n_data_sample=n_data_sample) for _ in range(n_task)]
-        return compute_plan
-
-    return _create_compute_plan
-
-
 @pytest.mark.django_db
 def test_n_plus_one_queries_performance_list(authenticated_client, create_compute_plan):
     # Dummy request, the fist request seems to trigger a caching system caching 4 requests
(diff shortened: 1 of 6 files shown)


Owlfred:

guilhem-barthes:
/e2e --tests sdk,frontend,substrafl --benchmarks mnist,camelyon

Owlfred:

backend/api/tests/views/test_views_performance.py
```diff
@@ -347,3 +350,64 @@ def test_performance_export_filter_and(self):
         content_list = list(response.streaming_content)
         self.assertEqual(len(content_list), 4)
         self.assertTrue(status in str(content_list[1]))
+
+
+@pytest.fixture
```
SdgJlbl, line 355: If we use this fixture in multiple places, maybe it's worth putting it in `conftest.py`?

backend/api/tests/views/test_views_performance.py
```diff
@@ -347,3 +350,64 @@ def test_performance_export_filter_and(self):
...
+
+@pytest.mark.django_db
+def test_n_plus_one_queries_performance_list_empty(authenticated_client, create_compute_plan):
+    # This expects 12 requests even without any performance and 14 if performances are made
+    compute_plan = create_compute_plan(n_task=4)
+    url = reverse("api:compute_plan_perf-list", args=[compute_plan.key])
+    with utils.CaptureQueriesContext(connection) as query:
+        print(authenticated_client.get(url))
+    query_tasks = len(query.captured_queries)
+    assert query_tasks == 12
```
SdgJlbl, line 400: I've been told by my local DSF fellow that I should not check for an exact number of queries in my tests, since it's typically the kind of things that can change slightly due to unrelated changes somewhere else in the API. Rather, check for a range to ensure that it's not exploding / evolving linearly with the number of tasks / samples.

SdgJlbl: Up this comment

backend/api/tests/views/test_views_performance.py
```diff
@@ -350,3 +350,64 @@ def test_performance_export_filter_and(self):
...
+    def _create_compute_task(compute_plan, n_data_sample=4):
+        data_manager = factory.create_datamanager()
+        data_samples = [factory.create_datasample([data_manager]) for _ in range(n_data_sample)]
+        input_keys = {
+            "opener": [data_manager.key],
+            "datasamples": [data_sample.key for data_sample in data_samples],
+        }
+        function = factory.create_function(
+            inputs=factory.build_function_inputs(["datasamples", "opener", "model"]),
+            outputs=factory.build_function_outputs(["performance"]),
```
ThibaultFy, line 366: test with a dummy output in addition to performance

ThibaultFy (approved):
lgtm !

backend/api/tests/views/test_views_performance.py
guilhem-barthes: I like this approach, it also really tests what we want to do - that performances do not impact a lot these requests. I also expect the other PR to decrease this number, going from 12 for an empty compute plan to less

guilhem-barthes: @SdgJlbl , I went to the smallest comparison (e.g. won't fail if the number of calls decrease), but then the difference between with and without performance is only < 3, does it look ok?

SdgJlbl: well, if we did not forget any prefetch_related and with the right model, querying all performance should be in O(1), so possibly only 1 more call compared to not querying it. So yeah, not so easy to test, given that Django is caching at least 4 queries between the first time and the subsequent times.

guilhem-barthes:
/e2e --tests sdk,frontend,substrafl --benchmarks mnist,camelyon

Owlfred:

PR #612, opened 2023-03-09
fix: test due to deprecation

guilhem-barthes:
## Description

Add new deprecation to remove error in CI due to new deprecation warning set in `setuptools` version [67.5.0](https://setuptools.pypa.io/en/stable/history.html#v67-5-0)

## How has this been tested?

Local CI

## Checklist


SdgJlbl (approved):
Yes :) 
Do we have a way of setting a reminder to check if this filter is still needed later down the line?

oleobal:
> Do we have a way of setting a reminder to check if this filter is still needed later down the line?

Good idea

oleobal (approved):

PR #634, opened 2023-04-14
feat!: multiple performance per task

ThibaultFy:
# Description

Performance asset were unique regarding their compute task key, and the function key they were computed from. Or, we want to have the possibility to have a function that outputs several performance and store them in the same task.

This implies two main things:
   - Performance are now unique regarding the compute task key, the function key AND the output identifier
   - The frontend must serialize regarding the identifier, and not the function key

The PR batch modify both the orchestrator and backend db in order to include the identifier as an additional unique condition.
In the backend, we use the ComputeTaskOutput as primary key directly.

## Companion PR

- https://github.com/Substra/substra-backend/pull/634
- https://github.com/Substra/substra-tests/pull/251 (main)
- https://github.com/Substra/orchestrator/pull/197
- https://github.com/Substra/substra-frontend/pull/194
- https://github.com/Substra/substra/pull/357

## E2E tests
- https://github.com/owkin/substra-ci/actions/runs/4872054802

-


backend/api/models/performance.py
```diff
@@ -4,9 +4,10 @@
 class Performance(models.Model):
     compute_task = models.ForeignKey("ComputeTask", on_delete=models.deletion.DO_NOTHING, related_name="performances")
     metric = models.ForeignKey("Function", on_delete=models.deletion.DO_NOTHING, related_name="performances")
+    identifier = models.CharField(max_length=100)
```
sergebouchut2, line 7: I think it's better to explicit this is not the id of the perf asset.
```suggestion
    task_output_identifier = models.CharField(max_length=100)
```

sergebouchut2: Or maybe it's even better to use a FK to the [task output asset](https://github.com/Substra/substra-backend/blob/main/backend/api/models/computetask.py#L27) but you will have to ensure the "parent" asset is created prior the perf

sergebouchut2: In this case I wonder if the compute task FK remains relevant or should be replaced by the task output FK

sergebouchut2: You may have to also update https://github.com/Substra/substra-backend/blob/main/backend/api/events/sync.py

backend/api/management/commands/generate_fixtures.py
```diff
@@ -191,11 +191,11 @@ def create_basic_cp(self, cp_status):
             status=task_status,
         )
         if task_status == ComputeTask.Status.STATUS_DONE:
-            factory.create_performance(
-                test_task,
-                self.metric_function,
-                identifier="performance",
-            )
+            for task_output in test_task.outputs:
```
guilhem-barthes, line 194: Won't it create performance for `local`, `model` or `shared` ?

ThibaultFy: test tasks normally only have performances as output, at least in this factory :)

backend/api/serializers/computetask.py
```diff
@@ -85,8 +85,12 @@ def get_value(self, task_output):
...
                 data.append(ModelSerializer(context=self.context, instance=model).data)
             elif output_asset.asset_kind == FunctionOutput.Kind.ASSET_PERFORMANCE:
-                task_key, metric_key = output_asset.asset_key.split("|")
-                perf = Performance.objects.get(compute_task__key=task_key, metric__key=metric_key)
+                compute_task_key, metric_key, identifier = output_asset.asset_key.split("|")
+                task_output = ComputeTaskOutput.objects.get(
+                    task=compute_task_key,
+                    identifier=identifier,
+                )
+                perf = Performance.objects.get(compute_task_output=task_output, metric__key=metric_key)
```
SdgJlbl, line 93: This section is a nightmare for performance (execution time), both curent and new implementation. 
This is not a very constructive comment, because I don't have a good alternative to offer, and anyway, it would be out of scope of this PR. But maybe it's worth to keep it in mind and do some (manual) tests on a project with a lot of data and 10  metrics (you know the one I'm thinking about) 🙂

backend/api/serializers/computetask.py
```diff
@@ -148,8 +152,12 @@ def get_asset(self, task_output_asset):
             model = Model.objects.get(key=task_output_asset.asset_key)
             return ModelSerializer(context=self.context, instance=model).data
         elif task_output_asset.asset_kind == FunctionOutput.Kind.ASSET_PERFORMANCE:
-            task_key, metric_key = task_output_asset.asset_key.split("|")
-            performance = Performance.objects.get(compute_task__key=task_key, metric__key=metric_key)
+            compute_task_key, metric_key, identifier = task_output_asset.asset_key.split("|")
```
SdgJlbl, line 155: Does it still make sense to have both a `get_value` and a `get_asset` or is one of the two soon to be deprecated? 
Again, not the point of this PR, but for my understanding 🙂

backend/api/serializers/computetask.py
ThibaultFy: Yes I agree... But I had in mind that @guilhem-barthes was planning on deleting these get_asset and get_value, so it might be the solution ahah :)

backend/api/serializers/computetask.py
ThibaultFy: Summoning @guilhem-barthes as he already spend some time on the question

backend/api/serializers/performance.py
```diff
@@ -11,14 +12,11 @@ class PerformanceSerializer(serializers.ModelSerializer, SafeSerializerMixin):
     # the unique constraint for compute_task_key + metric_key in the model
     # ensures no duplicates are created
     primary_key_name = "id"
+    compute_task_key = serializers.UUIDField(format="hex_verbose", source="compute_task_output.task_id")
```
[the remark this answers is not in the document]
ThibaultFy: You're right. And I put back the `PrimaryKeyRelatedField`. Not sure how it works but it seems that it is the way to do it regarding all the other serializers

backend/api/tests/views/test_views_performance.py
```diff
@@ -364,9 +371,10 @@ def test_n_plus_one_queries_performance_list(authenticated_client, create_comput
...
     for t in compute_plan.compute_tasks.all():
-        factory.create_performance(t, t.function)
+        perf_output = t.outputs.all()[1]
+        factory.create_performance(perf_output, t.function)
     with utils.CaptureQueriesContext(connection) as query:
         print(authenticated_client.get(url))
     query_task_with_perf = len(query.captured_queries)
-    assert query_task_with_perf < 11
-    assert query_task_with_perf - query_tasks_empty < 3
+    assert query_task_with_perf < 12
```
SdgJlbl, line 379: 🕵️

SdgJlbl: does the number of queries increase linearly with the number of metrics? 🙃

backend/api/serializers/performance.py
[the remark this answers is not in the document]
SdgJlbl: from [DRF documentation](https://www.django-rest-framework.org/api-guide/relations/#primarykeyrelatedfield), it relates to how we want to nest the objects

backend/api/tests/views/test_views_performance.py
ThibaultFy: No, but I have an additional query for the identifier now :/

backend/api/serializers/computetask.py
guilhem-barthes: This is the leftover GT task, but still considering why we don't generate a key instead of concatenating keys

guilhem-barthes: Also, do we still need the `metric_key`? wouldn't be `compute_task_key|identifier` unique?

guilhem-barthes (commented):
Thanks for your work!

backend/api/serializers/performance.py
```diff
@@ -28,27 +30,32 @@ class Meta:
             "compute_task_key",
             "creation_date",
             "metric_key",
+            "compute_task_output_identifier",
             "performance_value",
         ]
 
+    def create(self, validated_data):
```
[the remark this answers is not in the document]
ThibaultFy: As we need to get the compute task output from compute task key + identifier (receive from the orc event), there is an extra step that is not dealt with by default in DRF

backend/api/serializers/computetask.py
ThibaultFy: Mmmmh this is a good question. before, the metric key was used by the frontend to serialize the graph. But with the work we did on the front with Milan, the metric key completely disapear from the frontend now. So we might want to also delete it from the performance...

backend/api/serializers/performance.py
```diff
@@ -28,27 +30,32 @@ class Meta:
...
+        task_output = ComputeTaskOutput.objects.get(
```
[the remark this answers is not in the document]
ThibaultFy: No opinion on this... What do you think would be best ?

backend/api/serializers/performance.py
```diff
@@ -70,15 +77,17 @@ def get_round_idx(self, instance):
 
 
 class CPPerformanceSerializer(serializers.ModelSerializer):
-    compute_task = _PerformanceComputeTaskSerializer(read_only=True)
+    compute_task = _PerformanceComputeTaskSerializer(read_only=True, source="compute_task_output.task")
     metric = _PerformanceMetricSerializer(read_only=True)
```
guilhem-barthes, line 81: Do we still need `metric`? it looks like that we don't use any `metric` fields anymore.

guilhem-barthes (commented):
Do we still need the link between

PR #635, opened 2023-04-17
feat: Add functionName to workflow view & allow ordering in tasks

Milouu:
## Description

-  Adds function name to workflow view so that the frontend can display the function name for each task.
-  Allows ordering by function name in compute tasks list now that the frontend allow it. 

Companion PR : https://github.com/Substra/substra-frontend/pull/188

## Checklist

### DIFF
--- a/backend/api/views/compute_plan_graph.py
+++ b/backend/api/views/compute_plan_graph.py
@@ -61,8 +61,8 @@ def get_cp_graph(request, compute_plan_pk):
 
     tasks = (
         ComputeTask.objects.filter(compute_plan__key=compute_plan_pk, channel=get_channel_name(request))
-        .annotate(inputs_specs=Subquery(inputs), outputs_specs=Subquery(outputs))
-        .values("key", "rank", "worker", "status", "inputs_specs", "outputs_specs")
+        .annotate(inputs_specs=Subquery(inputs), outputs_specs=Subquery(outputs), function_name=F("function__name"))
+        .values("key", "rank", "function_name", "worker", "status", "inputs_specs", "outputs_specs")
     )
 
     # Set a task limitation for performances issues
--- a/backend/api/views/compute_plan_graph.py
+++ b/backend/api/views/compute_plan_graph.py
@@ -61,8 +61,10 @@ def get_cp_graph(request, compute_plan_pk):
 
     tasks = (
         ComputeTask.objects.filter(compute_plan__key=compute_plan_pk, channel=get_channel_name(request))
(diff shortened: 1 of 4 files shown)


backend/api/views/compute_plan_graph.py
```diff
@@ -61,8 +61,8 @@ def get_cp_graph(request, compute_plan_pk):
 
     tasks = (
         ComputeTask.objects.filter(compute_plan__key=compute_plan_pk, channel=get_channel_name(request))
-        .annotate(inputs_specs=Subquery(inputs), outputs_specs=Subquery(outputs))
-        .values("key", "rank", "worker", "status", "inputs_specs", "outputs_specs")
+        .annotate(inputs_specs=Subquery(inputs), outputs_specs=Subquery(outputs), function_name=F("function__name"))
+        .values("key", "rank", "function_name", "worker", "status", "inputs_specs", "outputs_specs")
```
SdgJlbl, line 65: I'm not sure if this is not missing a `prefetch_related` or a `select_related`, could you add a [n + 1 queries test](https://github.com/Substra/substra-backend/blob/main/backend/api/tests/views/test_views_computetask.py#L831) to check that it's fine? 
We can pair on it if you want.

SdgJlbl (commented):
Thanks for the PR, I think this change needs a test to check that the number of queries is not growing linearly, but we can pair on that if you want :)

backend/api/views/compute_plan_graph.py
SdgJlbl: Okay, so the prefetch/select_related is not needed because the `annotate` uses a JOIN underneath https://docs.djangoproject.com/en/4.2/topics/db/aggregation/#joins-and-aggregates

I think it would be worth it adding a comment to mention that fact, because it wasn't obvious (at least for me).

backend/api/views/compute_plan_graph.py
```diff
@@ -61,8 +61,10 @@ def get_cp_graph(request, compute_plan_pk):
 
     tasks = (
         ComputeTask.objects.filter(compute_plan__key=compute_plan_pk, channel=get_channel_name(request))
-        .annotate(inputs_specs=Subquery(inputs), outputs_specs=Subquery(outputs))
-        .values("key", "rank", "worker", "status", "inputs_specs", "outputs_specs")
+        # fetch_related not needed because annotate already performs a JOIN
```
SdgJlbl, line 64:
```suggestion
        # prefetch_related not needed because annotate already performs a JOIN
```

SdgJlbl (approved):
Thanks for the PR 🚀

PR #639, opened 2023-04-24
feat!: add support for multiple API tokens with expanded functionality

IC-1101asterisk:
## Description

https://app.asana.com/0/1204424791229302/1204269294564790

New `BearerToken` model:
- has note attribute
- has token_id attribute
- has expiry attribute
- allows for multiple tokens per user

`active-api-tokens` received a new delete method
`api-token-auth` changed, now you can specify an expiry date and a note, no expiry date means an imperishable token  

## Checklist

## TODO
- [x] add tests 
- [x] add expiration in authentication backend
- [x] remove the reuse `TOKEN_STRATEGY`


backend/backend/views.py
```diff
@@ -72,6 +64,17 @@ def get(self, request, *args, **kwargs):
...
         return ApiResponse({"tokens": [_bearer_token_dict(token, include_payload=False) for token in tokens]})
 
+    def delete(self, request, *args, **kwargs):
+        if BearerToken.objects.filter(token_id=request.GET.get("id")):
+            token = BearerToken.objects.get(token_id=request.GET.get("id"))
+            if request.user == token.user:
+                token.delete()
+                return HttpResponse("Authorized, token removed", status=200)
+            else:
+                return HttpResponse("Unauthorized, token was NOT removed", status=401)
```
oleobal, line 74: This is good, though three small notes here:
 - no need to include "authorized" in the response message since it is implied by the success
 - this is probably paranoia but let's not leak internal status to unauthorized users, just return 404 if authentication fails
 - it might be better to return a simple JSON dictionary; the reason is it's simpler to manipulate the API if it always return the same data format.
   I'm not 100% sure though, we should take a look through the rest of our code base to see whether there are other cases where we return plain text.

backend/users/models/token.py
```diff
@@ -0,0 +1,30 @@
...
+    note = models.TextField(default="Empty Note")
+    # for now it is a date time field but a lenght could also be smart
+    expiry = models.DateTimeField(default=None, null=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL, related_name="bearer_tokens", on_delete=models.CASCADE, verbose_name=("Users")
+    )
+    token_id = models.UUIDField(default=uuid.uuid4, editable=False)
+
+    def expires_at(self) -> models.DateTimeField:
+        return self.expiry
```
oleobal, line 19: This can probably be removed

backend/users/models/token.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def is_token_expired(self) -> bool:
+        if self.expiry is None:
+            return False
+        return self.time_left() < timedelta(seconds=0)
```
guilhem-barthes, line 24: Is there any known drawback to compare the `expiry` to the current date instead of calculating the time left on the token and then comparing this duration to 0?

backend/backend/views.py
```diff
@@ -72,6 +64,17 @@ def get(self, request, *args, **kwargs):
...
+        else:
+            return HttpResponse("Token not found", status=404)
```
guilhem-barthes, line 76: In this block of code, you calling the DB two times to get the same information (`filter` then `get`).

You could either work with a `tokens = BearerToken.objects.filter()` (and then use the first element) or use directly `BearerToken.objects.get()` and leverage the exception management system. In this case, you would have [2 exception types](https://docs.djangoproject.com/en/4.2/topics/db/queries/#retrieving-a-single-object-with-get) to take in account.

In this case, the second option (leveraging Exception) seems cleaner.

backend/backend/views.py
guilhem-barthes, line 74: I would recommend using `from rest_framework import status` (from rest_framework import status)  instead of plain number for the HTTP code, it increase the readability (we directly now if you return a UNAUTHORIZED/FORBIDDEN/OK/CONFLICT etc) and could be automatically auto-completed by your IDE.

backend/backend/views.py
```diff
@@ -34,15 +34,7 @@ def post(self, request, *args, **kwargs):
...
-        if settings.TOKEN_STRATEGY == "reuse":  # nosec
-            token, created = BearerToken.objects.get_or_create(user=user)
-            # handle_token_expiration will check whether the token is expired
-            # and will generate a new one if necessary
-            is_expired, token = bearer_token_utils.handle_token_expiration(token)
-        else:
-            # token should be new each time, remove the old one
-            BearerToken.objects.filter(user=user).delete()
-            token = BearerToken.objects.create(user=user)
+        token = BearerToken.objects.create(user=user, expiry=request.GET.get("expiry"), note=request.GET.get("note"))
```
guilhem-barthes, line 37: Why do we use request.GET in this POST request?

backend/backend/views.py
guilhem-barthes: I do agree that I would prefer returning a JSON file, whether or not we return plain text in other part of the code. I would more consider the other plain text to JSON than the other way around.

I think the `return HttpResponse("Unauthorized, token was NOT removed", status=401)` is more a Forbidden (403) than Unauthorized (which confusingly [relates to authentication, not authorization](https://www.rfc-editor.org/rfc/rfc9110#name-403-forbidden)) , but then I agree with @oleobal - returning 403 would expose the different tokens ID, without being sure what are the implication of it.

backend/backend/views.py
```diff
@@ -12,11 +12,11 @@
...
-from users.utils import bearer_token as bearer_token_utils
+from users.models.token import BearerToken
 
 
 def _bearer_token_dict(token: BearerToken, include_payload: bool = True) -> dict:
-    d = {"created_at": token.created, "expires_at": bearer_token_utils.expires_at(token)}
+    d = {"created_at": token.created, "expires_at": token.expires_at(), "note": token.note, "token_id": token.token_id}
     if include_payload:
         d["token"] = token.key
     return d
```
guilhem-barthes, line 22: This kind of behavior looks like the use-case of DRF serializer, (and how [DRF works internally for their tokens](https://github.com/encode/django-rest-framework/blob/master/rest_framework/authtoken/serializers.py)). Are we trying to make this part of the code less dependant of DRF (summoning @oleobal).

backend/users/utils/bearer_token.py
```diff
@@ -3,9 +3,11 @@
...
 from django.conf import settings
 from django.utils import timezone
-from rest_framework.authtoken.models import Token as BearerToken
 
+from users.models.token import BearerToken
 
+
+# this is not needed anymore ask Oliver to delete TODO
 def expires_at(token: BearerToken) -> datetime:
     return timezone.now() + time_left(token)
```
guilhem-barthes, line 12: You may be able to actually remove this file completely

guilhem-barthes (commented):
Thanks for your work, I think some part of this code can be optimized!

backend/backend/views.py
```diff
@@ -54,9 +46,9 @@ class AuthenticatedBearerToken(DRFObtainAuthToken):
     permission_classes = [IsAuthenticated]
 
     def get(self, request, *args, **kwargs):
-        # create a new token each time (you don't want a token that is about to expire)
-        BearerToken.objects.filter(user=request.user).delete()
-        token = BearerToken.objects.create(user=request.user)
+        token = BearerToken.objects.create(
+            user=request.user, expiry=request.GET.get("expiry"), note=request.GET.get("note")
```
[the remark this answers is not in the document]
IC-1101asterisk: I wrote some checks to change the behavior depending on what is missing in the request, this avoids the 400 and uses default values

backend/backend/views.py
IC-1101asterisk: I made the change except for the JSON dictionary, have you got any example in the code base ? I can't seem to find one, if not could one of you write it for let's say the 200 response ? I will figure it out for the 404

backend/backend/views.py
IC-1101asterisk: I can't explain why but if I use POST my parameters passed in the URL are then ignored

backend/backend/views.py
```diff
@@ -34,15 +35,7 @@ def post(self, request, *args, **kwargs):
...
-        if settings.TOKEN_STRATEGY == "reuse":  # nosec
-            token, created = BearerToken.objects.get_or_create(user=user)
-            # handle_token_expiration will check whether the token is expired
-            # and will generate a new one if necessary
-            is_expired, token = bearer_token_utils.handle_token_expiration(token)
-        else:
-            # token should be new each time, remove the old one
-            BearerToken.objects.filter(user=user).delete()
-            token = BearerToken.objects.create(user=user)
+        token = BearerToken.objects.create(user=user, expiry=request.GET.get("expiry"), note=request.GET.get("note"))
```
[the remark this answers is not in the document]
oleobal: Actually scratch that last remark, let Substra SDK set its own token titles

backend/users/models/token.py
oleobal: +1

backend/backend/views.py
oleobal: Right, Django calls them `request.GET` and `request.POST` but they maybe should be called `request.url_parameters` and `request.body_parameters`.

Basically we copied the code from `AuthenticatedBearerToken.get` and this is why this classes mixes URL and body parameters. It doesn't necessarily _not_ make sense because the URL parameters are token settings while the body params are authentication -- plus that way the token params are the same for both `/api-token-auth` and `/api-token`.

backend/users/models/token.py
IC-1101asterisk: No idea about this, just did it the same way Olivier did before

backend/backend/views.py
```diff
@@ -72,6 +75,16 @@ def get(self, request, *args, **kwargs):
 
         return ApiResponse({"tokens": [_bearer_token_dict(token, include_payload=False) for token in tokens]})
 
+    def delete(self, request, *args, **kwargs):
+        try:
+            token = BearerToken.objects.get(token_id=request.GET.get("id"))
+            if request.user == token.user:
+                token.delete()
+                return HttpResponse("Token removed", status=status.HTTP_200_OK)
+        except ObjectDoesNotExist:
```
guilhem-barthes, line 84:
```suggestion
        except BearerToken.ObjectDoesNotExist:
```
and remove import on top.

You may also want to cover `BearerToken.MultipleObjectsReturned` (not expected though)

backend/backend/views.py
```diff
@@ -54,9 +59,7 @@ class AuthenticatedBearerToken(DRFObtainAuthToken):
     permission_classes = [IsAuthenticated]
 
     def get(self, request, *args, **kwargs):
```
guilhem-barthes, line 61: Not directly linked with this PR, but I am not sure about the use of a `GET` request when the aim is to create a new resource

backend/backend/views.py
```diff
@@ -12,16 +15,26 @@
...
         d["token"] = token.key
     return d
 
 
+def _create_token_for_user(user, params) -> BearerToken:
+    if "expiry" not in params or "note" not in params:
+        raise ValidationError("expiry or note field are missing!")
+    return BearerToken.objects.create(
+        user=user,
+        expiry=None if params.get("expiry") == "never" else params.get("expiry"),
```
guilhem-barthes, line 33: This validation seems insufficient, what if the user send "toto" ? This is nether `None` or a `DateTime`

guilhem-barthes (commented):
Thanks for the changes!

Although not entirely linked with this PR, I would mention that the file organization seems a little messy (test in one Django application, views in another Django app and lastly tests in a third one).

Following discussions on another medium to separate the different tokens (SDK and Frontend), if you decide having different tables for the different kinds of tokens and they are both using  the same (or a similar) model I would recommend looking in [abstract classes](https://docs.djangoproject.com/en/dev/topics/db/models/#abstract-base-classes)

backend/backend/views.py
IC-1101asterisk: done

backend/backend/views.py
```diff
@@ -53,11 +48,11 @@ class AuthenticatedBearerToken(DRFObtainAuthToken):
 
     permission_classes = [IsAuthenticated]
 
-    def get(self, request, *args, **kwargs):
-        # create a new token each time (you don't want a token that is about to expire)
-        BearerToken.objects.filter(user=request.user).delete()
-        token = BearerToken.objects.create(user=request.user)
-        return ApiResponse(_bearer_token_dict(token))
+    def post(self, request, *args, **kwargs):
+        s = BearerTokenSerializer(data=JSONParser().parse(request))
```
guilhem-barthes, line 52: `BearerTokenSerializer(data=request.data)` was not working?

backend/users/models/token.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class ImplicitBearerToken(Token):
+    """
+    Legacy token to make the endpoint api-token-auth/ work like it used to
+    """
+
+    @property
+    def expires_at(self) -> datetime:
+        return timezone.now() + self.time_left
```
guilhem-barthes, line 33: I feel like using `time_left` there makes useless calculation. This equals to:
```
timezone.now() + self.time_left
-> timezone.now() + (settings.EXPIRY_TOKEN_LIFETIME - time_elapsed)
  -> timezone.now() + (settings.EXPIRY_TOKEN_LIFETIME - (timezone.now() - self.created))
    -> timezone.now() + (settings.EXPIRY_TOKEN_LIFETIME - timezone.now() + self.created)
      -> settings.EXPIRY_TOKEN_LIFETIME + self.created
```
We could simplify this line to `settings.EXPIRY_TOKEN_LIFETIME + self.created`. I'm also not sure that the two `timezone.now()` are going to be equal.

Then, as it looks like that `time_left` is only used there, then I would then rather just replace `is_expired` with the line `return self.expires_at > timezone.now()`  and remove `time_left`

backend/users/models/token.py
```diff
@@ -0,0 +1,55 @@
...
+        left_time = settings.EXPIRY_TOKEN_LIFETIME - time_elapsed
+        return left_time
+
+    @property
+    def is_expired(self) -> bool:
+        return self.time_left < timedelta(seconds=0)
+
+    def handle_expiration(self) -> "ImplicitBearerToken":
+        """
+        If token is expired new token will be established.
```
guilhem-barthes, line 47:
```suggestion
        If token is expired a new token will be created.
```

backend/users/models/token.py
```diff
@@ -0,0 +1,55 @@
...
+        If token is expired then it will be removed
+        and new one with different key will be created
```
guilhem-barthes, line 49: Sounds similar to the previous line

guilhem-barthes (commented):
Thanks for all the changes, it looks good, although I think some parts still needs small refinements.

docs/authentication.md
```diff
@@ -1,23 +1,23 @@
...
-  * The session key is a random string encoded in ASCII format. [Source](https://docs.djangoproject.com/fr/4.1/topics/http/sessions)
-  * The server returns a cookie with the session key to the client.
+  - The session key is stored on the server side (in `django_session` table), alongside an expiration date and an encoded dict of data containing the user ID.
+  - The session key is a random string encoded in ASCII format. [Source](https://docs.djangoproject.com/fr/4.1/topics/http/sessions)
+  - The server returns a cookie with the session key to the client.
 
 1. Retrieve the login form to download the CSRF cookie and the form token.
 
-```bash
+````bash
```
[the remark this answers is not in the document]
IC-1101asterisk: nice catch that was an error in the documentation unless we actually want a markdown looking like this
image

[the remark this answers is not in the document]
IC-1101asterisk: oops bad copy paste with the docs, should have previewed the md file

docs/authentication.md
```diff
@@ -75,17 +76,40 @@ curl \
...
   --header 'Accept: application/json' \
+  --header 'Content-Type: application/json' \
   --header 'Authorization: Token <auth_token_value>' \
-  http://substra-backend.org-1.com/algo/
-{"count":0,"next":null,"previous":null,"results":[]}
+  'http://substra-backend.org-1.com:8000/active-api-tokens/'
+{"tokens": {"id": "<uuid4 id>", "note": "<string>", "expires_at":"<ISO timestamp>", "created_at":"<ISO timestamp>"}, "implicit_token":{"expires_at":"<ISO timestamp>", "created_at":"<ISO timestamp>"}}
+```
+
+3. Create a new token using token identication ("expires_at" must be null or DDDD-MM-YY)
```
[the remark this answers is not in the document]
oleobal: Shouldn't it be `authentication`?

backend/api/tests/views/test_views_token.py
```diff
@@ -0,0 +1,114 @@
...
+    token_1 = BearerToken.objects.create(user=user)
+    token_2 = BearerToken.objects.create(user=user)
+
+    tokens_count = BearerToken.objects.count()
+    assert tokens_count == 2
+
+    valid_auth_token_header = f"Token {token_2}"
+    api_client.credentials(HTTP_AUTHORIZATION=valid_auth_token_header)
+
+    response = api_client.get("/active-api-tokens/")  # none type error here ??
```
guilhem-barthes, line 41: This comment looks temporary

guilhem-barthes (approved):
Thanks for your work! I just noticed a comment that seems to be not meant to be left in the code

PR #642, opened 2023-04-28
chore(ci): switch to using a reusable workflow

oleobal:
## Description

This replaces existing workflows for docker build with a new distant one you can find here: https://github.com/Substra/substra-gha-workflows/blob/main/.github/workflows/docker-build.yaml

This also changes how the CI works:
 - instead of git tags being used as reference point for publish packages, GitHub releases are used.
   - this is like the python packages for substra and substrafl
 - the `latest` docker tag is set to the latest release instead of the latest main
 - the trigger conditions are slightly different but shouldn't really make a difference

There will be PRs to do the same thing for the orchestrator, the frontend and hlf-k8s. The main immediate benefit is having a standard release process for publish packages for all our repos. In the future we can move more workflows to a central model so as to reduce code replication and ensure there are no variations from one repo to another.

There are small changes to the image (`VERSION` instead of `BACKEND_VERSION`) because the reusable workflow only sets `VERSION`, but that's a feature, not a bug: this makes it work like the other repos.

### Companion PRs

https://github.com/Substra/orchestrator/pull/205
https://github.com/Substra/substra-frontend/pull/195
https://github.com/Substra/hlf-k8s/pull/153

## How has this been tested?

I tried it a bunch of times on `substra-gha-workflows` but basically we'll fix problems in production.

Also in the checks to this PR ⬇️ 

## Checklist

- ~~[ ] [changelog](../CHANGELOG.md) was updated with notable changes~~ no notable changes
- ~~[ ] documentation was updated~~ no documentation changes


SdgJlbl (approved):
Let's start the debugging process on main then ❤️

PR #658, opened 2023-05-19
feat(helm)!: allow connecting to an external postgres

oleobal:
## Description

This allows connecting to an external PostgreSQL database. The existing `postgresql` key is therefore renamed to `integrated-postgresql`, with `postgresql` now being for connection details to any database, external or internal.

I've also added a new feature: `postgresql.auth.credentialsSecretName` allows using a secret instead of plain text keys, which should be more secure.

Companion: https://github.com/Substra/orchestrator/pull/210

## How has this been tested?

TBD

## Checklist

- [ ] update readme to explain how to connect to an external database
    - [ ] also a warning about credentials


charts/substra-backend/README.md
```diff
@@ -286,6 +286,25 @@ Else, you must strike a balance: longer durations are more convenient, but risk
...
+| ------------------------------------- | ----------------------------------------------------------------------------------------------------------- | ---------- |
+| `database.auth.database`              | what DB to connect to                                                                                       | `substra`  |
+| `database.auth.username`              | what user to connect as                                                                                     | `postgres` |
+| `database.auth.password`              | what password to use for connecting                                                                         | `postgres` |
+| `database.auth.credentialsSecretName` | An alternative to giving username and password; must have `DATABASE_USERNAME` and `DATABASE_PASSWORD` keys. | `nil`      |
+| `database.host`                       | Hostname of the database to connect to (defaults to local)                                                  | `nil`      |
+| `database.port`                       | Port of an external database to connect to                                                                  | `5432`     |
+
+### Integrated PostgreSQL settings
```
SdgJlbl, line 300: Do we want to keep the "Integrated" denomination here? 
Pros: it's more explicit
Cons: it's not a denomination used else where 🤷

charts/substra-backend/README.md
```diff
@@ -286,6 +286,25 @@ Else, you must strike a balance: longer durations are more convenient, but risk
...
+
+See Bitnami documentation
+
+| Name                 | Description                                                     | Value  |
+| -------------------- | --------------------------------------------------------------- | ------ |
+| `postgresql.enabled` | Deploy a PostgreSQL instance along the orchestrator for its use | `true` |
```
SdgJlbl, line 306:
```suggestion
| `postgresql.enabled` | Deploy a PostgreSQL instance along the backend for its use | `true` |
```

SdgJlbl (approved):
One typo, one question about naming, otherwise LGTM 👍 
Thanks

charts/substra-backend/README.md
oleobal: I fully agree, maybe I can just call it "PostgreSQL settings" and add a little blurb in the section description to explain what it is.

SdgJlbl: yes, very good idea

PR #668, opened 2023-06-05
chore: make secret key configurable and reorganize settings

oleobal:
## Description

Rename some files and variables to be a bit clearer. Reorganize the settings to lower the size of common.py. Split off CORS and OIDC settings into their own category of settings (`mods`) because they change settings from common and therefore must be applied afterwards.

Add two new configuration options:`SECRET_KEY_PATH` and `SECRET_KEY_LOAD_AND_STORE`

Inline `gen_secret_key.py`

This is a backwards compatible prerequisite PR to: https://github.com/Substra/substra-backend/pull/657

## Checklist


oleobal:
@thbcmlowk 

> How about moving JWT related statements into a dedicated module?

Why not, I think that would fit under this PR. I didn't think of it because our settings files are usually a bit bigger than this but really splitting `settings/common.py` might be a good idea.

backend/backend/settings/common.py
```diff
@@ -12,16 +12,15 @@
...
 import sys
 from datetime import timedelta
 
 import structlog
 from django.core.files.storage import FileSystemStorage
 
-from libs.gen_secret_key import write_secret_key
 from substrapp.compute_tasks.errors import CeleryRetryError
 
+from .deps.jwt import *
```
SdgJlbl, line 23: why do we need a `import *`? 
It's usually considered bad practice :/ 
(same for the line below though)

Owlfred:

backend/backend/settings/deps/orchestrator.py
```diff
@@ -1,11 +1,11 @@
 import os
 
-from .. import common
+from .utils import to_bool
```
SdgJlbl, line 3: 👍

Owlfred:

PR #676, opened 2023-06-20
fix: issues from SECRET_KEY_LOAD_AND_STORE and settings reorganization

oleobal:
## Description

#668 added a new `SECRET_KEY_LOAD_AND_STORE` boolean setting in the backend. If `SECRET_KEY_LOAD_AND_STORE` is false, then the `SECRET_KEY` is generated at startup and not stored. If it is true, the program tries reading it from `SECRET_KEY_PATH`, and if that fails generate a new one and write it there.

The intent was to allow configuring components that use the same settings as the backend server not to write the `SECRET_KEY` on the file system, because this eases configuration in production.

In #668 this was set to `False` by default. However to match the pre-668 behavior you'd need to set it to `True` by default.

Possibly we could get rid of this setting instead and just write the key every time.

I don't know _why_ frontend authentication on k8s breaks when this setting is set to `false`. The `SECRET_KEY` is used for signing JWTs, so presumably at some point Django is restarted and so previously generated JWTs aren't valid anymore.. But JWTs are requested (and therefore generated) right before they're needed, so that's weird.

---

#668 also moved the file containing this line:

```python
BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```

without updating it, which broke paths

## How has this been tested?

⬇️ 

## Checklist


Owlfred:

Owlfred:

oleobal:
/e2e --tests sdk,frontend

Just to make sure it's not passing by accident

Owlfred:

backend/backend/settings/deps/path.py
```diff
@@ -2,7 +2,7 @@
 import sys
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
-BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+BASE_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "../.."))
 PROJECT_ROOT = os.path.dirname(BASE_DIR)
```
guilhem-barthes, line 6: Would this be clearer?

```suggestion
from pathlib import Path

# [0] == current dir, [2] == 2 levels up
BASE_DIR = Path(__file__).resolve().parents[2]
PROJECT_ROOT = BASE_DIR.parent
```

L9 could also be `sys.path.append(PROJECT_ROOT / "libs")`

/ ! \ Seems that Django can use pathlike, otherwise indeed this solution would require `str()`

oleobal: It looks to be working and that seems to be [how Django recommends to do it](https://github.com/django/django/blob/107865780aa44914e21d27fdf4ca269bc61c7f01/docs/howto/overriding-templates.txt#L32) (the actual `PROJECT_ROOT` and `BASE_DIR` settings are undocumented, oh well).

pyproject.toml
```diff
@@ -22,7 +22,7 @@ addopts = "-n auto --cov=. --cov-config=../pyproject.toml -p no:unraisableexcept
 filterwarnings = [
     "error",
     "ignore:(?s).*Implementing implicit namespace package:DeprecationWarning:pkg_resources",
-    "ignore:pkg_resources is deprecated as an API:DeprecationWarning:pkg_resources",
+    "ignore:pkg_resources is deprecated as an API.:DeprecationWarning:pkg_resources",
```
guilhem-barthes, line 25: Following Thibault comment, I think we would have to change (with another unkown vlaue for me at this point) or remove the `pkg_resources` to make it work

oleobal: I can't get this warning locally.. Perhaps because I'm on 3.10? 🤔

oleobal: Nope, that wasn't it.

Owlfred:

oleobal:
._.

/e2e --tests sdk,frontend

🤞

Owlfred:

oleobal:
don't do this to me now

/e2e --tests sdk,frontend

Owlfred:

oleobal:
> The error you are facing is the same that failed last night - potentially not linked with your PR. Also, as it is a `FutureTimeoutError`, may be we should consider increasing the timeout for this specific test?

Maybe I should start reading the errors 😅 

Why would the timeout suddenly not be exceeded anymore? hmm

Let's discuss it outside this PR perhaps

guilhem-barthes:
Testing a fix here: https://github.com/Substra/substra-tests/pull/260

Owlfred:

backend/backend/settings/dev.py
```diff
@@ -25,8 +25,8 @@
 MODEL_BUCKET_NAME = "substra-model"
 MODEL_STORAGE = MinioStorage(MODEL_BUCKET_NAME)
 
-MEDIA_ROOT = os.environ.get("MEDIA_ROOT", os.path.join(PROJECT_ROOT, f"medias/{ORG_NAME}"))
-SERVERMEDIAS_ROOT = os.environ.get("SERVERMEDIAS_ROOT", os.path.join(PROJECT_ROOT, f"servermedias/{ORG_NAME}"))
+MEDIA_ROOT = os.environ.get("MEDIA_ROOT", str(PROJECT_ROOT / f"medias/{ORG_NAME}"))
```
guilhem-barthes, line 28:
```suggestion
MEDIA_ROOT = os.environ.get("MEDIA_ROOT", str(PROJECT_ROOT / "medias" / ORG_NAME))
```

backend/backend/settings/dev.py
```diff
@@ -25,8 +25,8 @@
...
+SERVERMEDIAS_ROOT = os.environ.get("SERVERMEDIAS_ROOT", str(PROJECT_ROOT / f"servermedias/{ORG_NAME}"))
```
guilhem-barthes, line 29:
```suggestion
SERVERMEDIAS_ROOT = os.environ.get("SERVERMEDIAS_ROOT", str(PROJECT_ROOT / "servermedias" / ORG_NAME))
```

guilhem-barthes (approved):
Thanks for your changes ! I just offered to change from f-string to pathlib path-like style.

oleobal:
/e2e --tests sdk,frontend

(paranoia)

Owlfred:

PR #680, opened 2023-06-29
feat: User awaiting approval

IC-1101asterisk:
## Description
This features adds a way to deploy Substra with users being created by the SSO as "unactivated users". They need the approval from an admin.

## How has this been tested?
Some tests needed to be changed as this is exploiting the fact that a user with no channel is an "unactivated user"

## Checklist


backend/api/tests/common.py
```diff
@@ -41,14 +42,13 @@ def __init__(
...
         user, created = User.objects.get_or_create(username=self.username)
+
         if created:
             user.set_password(self.password)
             user.save()
-            # for testing purpose most authentication are done without channel allowing to mock passing channel in
-            # header, this check is necessary to not break previous tests but irl a user cannot be created
-            # without a channel
             if self.channel:
                 UserChannel.objects.create(user=user, channel_name=self.channel, role=self.role)
```
oleobal, line 50: This doesn't need saving? 🤔  (as in `user.save()` just before)

backend/api/tests/views/conftest.py
```diff
@@ -12,22 +10,15 @@
 from api.tests.common import AuthenticatedClient
 
 _CHANNEL_NAME: Final[str] = "mychannel"
-_EXTRA_HTTP_HEADERS: Final[dict[str, str]] = {"HTTP_SUBSTRA_CHANNEL_NAME": _CHANNEL_NAME}
```
oleobal, line 15: What is the rationale for getting rid of this header? Does the client fill it on on its own when `client.channel` is set?

backend/users/middlewares/user_registration.py
oleobal, line 1: Is this file still useful?

backend/backend/settings/mods/oidc.py
```diff
@@ -34,10 +34,13 @@
...
-        raise Exception("No default channel provided for OIDC users")
-    if OIDC["USERS"]["DEFAULT_CHANNEL"] not in ledger.LEDGER_CHANNELS:
-        raise Exception(f"Channel {OIDC['USERS']['DEFAULT_CHANNEL']} does not exist")
+    OIDC["USERS"]["MUST_BE_APPROVED"] = to_bool(os.environ.get("OIDC_USERS_MUST_BE_APPROVED", "false"))
+    if OIDC["USERS"]["DEFAULT_CHANNEL"] and OIDC["USERS"]["MUST_BE_APPROVED"]:
+        raise Exception("Both 'default channel' and 'user must be approved' options are activated")
+    if not (OIDC["USERS"]["DEFAULT_CHANNEL"] or OIDC["USERS"]["MUST_BE_APPROVED"]):
+        raise Exception(
+            "At least one option between 'default channel' and 'user must be approved' needs to be activated"
+        )
```
oleobal, line 43: Could you add options in the chart values corresponding to this?

Right now OIDC users are given a default channel in the values here: https://github.com/Substra/substra-backend/blob/main/charts/substra-backend/values.yaml#L698

You should add a new option (for instance `oidc.users.requireApproval`), which you can link to this file here: https://github.com/Substra/substra-backend/blob/main/charts/substra-backend/templates/configmap-oidc.yaml

You'll also need to update the chart changelog and version.

backend/users/views/user.py
```diff
@@ -232,3 +243,57 @@ def generate_reset_password_token(self, request, *args, **kwargs):
...
+        channel_name = get_channel_name(request)
+        channel_data = {"channel_name": channel_name}
+        channel_data["role"] = _validate_role(d.get("role"))
+        channel_data["user"] = user
+        UserChannel.objects.create(**channel_data)
+        user.refresh_from_db()
+        data = UserSerializer(instance=user).data
+        return ApiResponse(data=data, status=status.HTTP_200_OK)  # get success header ?
+
+    # TODO THIS SHOULD NOT BE IN THE PULL REQUEST
```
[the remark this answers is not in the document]
IC-1101asterisk: ahahahah oops, good point I'll keep this open until I'm done with the front end too (I'm going to still need it)

backend/api/tests/common.py
IC-1101asterisk: that is a good question, during coding I put the user.save() before the UserChannel, but then Thibault asked me why, that when I notice I does not change anything.
My theory is that when you create a UserChannel it takes the user as an argument, so it must change the value in the User table and save it then

backend/api/tests/views/conftest.py
IC-1101asterisk: since the AuthenticatedClient now has a channel by default ('mychannel') this header (as well as every other headers) are now redundant, as long as you use the AuthenticatedClient, other kinds of Client in the tests still need similar headers

Owlfred:

backend/users/tests/test_user.py
```diff
@@ -338,21 +341,29 @@ def test_filter_user_role(self):
...
+            data={"role": "ADMIN"},
+        )
         assert response.status_code == status.HTTP_200_OK
         data = response.json()
         assert data["count"] == 1
         assert data["results"][0]["role"] == "ADMIN"
 
         # list only non admin users
-        response = AuthenticatedClient(channel=self.channel).get(self.url, data={"role": "USER"}, **self.extra)
+        response = AuthenticatedClient(channel=self.channel).get(
```
guilhem-barthes, line 363: It looks like L108 is not used anymore

backend/users/views/user.py
```diff
@@ -88,6 +93,12 @@ def has_permission(self, request, view):
         return request.user.channel.role == UserChannel.Role.ADMIN
 
 
+class IsAdmin(permissions.BasePermission):
+    def has_permission(self, request, view):
+        if request.user.is_authenticated:
+            return request.user.channel.role == UserChannel.Role.ADMIN
```
guilhem-barthes, line 99: Do we need to check `if request.user.is_authenticated:` ? It looks like `IsAdminOrReadOnly` doesn't

backend/users/views/user.py
```diff
@@ -232,3 +243,57 @@ def generate_reset_password_token(self, request, *args, **kwargs):
...
+
+    # TODO THIS SHOULD NOT BE IN THE PULL REQUEST
+    def post(self, request, *args, **kwargs):
+        d = json.loads(request.body)
+        user, created = User.objects.get_or_create(username=d.get("username"), email=d.get("email"))
+        if created:
+            user.set_password(d.get("password"))
+            user.save()
+        user_json = serialize("json", [user])
+        return ApiResponse({"users_awaiting_approval": user_json})
```
guilhem-barthes, line 299: Why do we return a list if it's always one element?

backend/api/tests/views/test_views_compute_plan_graph.py
```diff
@@ -236,9 +236,7 @@ def test_n_plus_one_queries_compute_graph(authenticated_client):
     url = reverse("api:workflow_graph", args=[compute_plan.key])
 
     with utils.CaptureQueriesContext(connection) as queries:
-        authenticated_client.get(
-            url, {"HTTP_SUBSTRA_CHANNEL_NAME": "mychannel", "HTTP_ACCEPT": "application/json;version=0.0"}
-        )
+        authenticated_client.get(url, {"HTTP_ACCEPT": "application/json;version=0.0"})
```
guilhem-barthes, line 239:
```suggestion
        authenticated_client.get(url, **self.extra)
```

backend/users/views/user.py
```diff
@@ -232,3 +243,57 @@ def generate_reset_password_token(self, request, *args, **kwargs):
...
+
+    def get_queryset(self):
+        return self.user_model.objects.filter(channel=None).exclude(username="deleted")
+
+    def delete(self, request, *args, **kwargs):
+        try:
+            user = User.objects.get(username=request.GET.get("username"))
+            user.delete()
+            return ApiResponse(data={"message": "User removed"}, status=status.HTTP_200_OK)
+        except User.DoesNotExist or User.MultipleObjectsReturned:
```
guilhem-barthes, line 271: I don't think we should return "User not found" if `User.MultipleObjectsReturned`. If it is currently needed to catch this exception (i.e. we don't require uniqueness on `username`), you should change the way you get users

backend/users/views/user.py
```diff
@@ -232,3 +243,57 @@ def generate_reset_password_token(self, request, *args, **kwargs):
...
+        try:
+            user = User.objects.get(username=request.GET.get("username"))
+        except User.DoesNotExist or User.MultipleObjectsReturned:
+            return ApiResponse(data={"message": "User not found"}, status=status.HTTP_404_NOT_FOUND)
+
+        channel_name = get_channel_name(request)
+        channel_data = {"channel_name": channel_name}
+        channel_data["role"] = _validate_role(d.get("role"))
+        channel_data["user"] = user
+        UserChannel.objects.create(**channel_data)
```
guilhem-barthes, line 286:
```suggestion
        channel_name = get_channel_name(request)
        role = _validate_role(d.get("role"))
        UserChannel.objects.create(channel_name=channel_name, role=role, user=user)
```
nitpicking

backend/backend/urls.py
```diff
@@ -43,6 +44,7 @@
                     path("api-token-auth/", obtain_auth_token),  # for expiry token authent
                     path("api-token/", obtain_auth_token_already_authenticated),
                     path("active-api-tokens/", active_bearer_tokens),
+                    path("unactivated-users/", UserAwaitingApprovalViewSet.as_view({"get": "list"})),
```
guilhem-barthes, line 47: Why is this needed?

guilhem-barthes (changes requested):
Thanks for your work!

I think there are still some small changes to be done, especially to not leave dead code behind. I am not sure what is the best practice for HTTP_ACCEPT, but if we go to remove it, we should remove it completely.

backend/api/tests/views/test_utils.py
```diff
@@ -56,7 +56,7 @@ def test_download_file_local_allowed(self):
         self.assertIn("local-organization", metadata.permissions_process_authorized_ids)
 
         with mock.patch("api.views.utils.get_owner", return_value="local-organization"):
-            response = self.client.get(self.function_url, **self.extra)
+            response = self.client.get(self.function_url)
```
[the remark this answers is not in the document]
IC-1101asterisk: I believe this one is needed since here the client is an OrganizationUser not an AuthenticatedUser 
So to me, two solutions, either I leave the extra for this kind of client (there are other extra elsewhere in the code) or I also change this kind of client to have a default channel (but I'm not sure it makes sense, I need to think about this)

backend/users/views/user.py
IC-1101asterisk: because for some reason serialize wants an iterable
but this doesn't matter this bits of code is for frontend testing, I'll delete it before merge (I kept a comment of Olivier open for that)

backend/api/tests/views/test_views_compute_plan_graph.py
IC-1101asterisk: this isn't in the class ComputePlanGraphViewTests thus self.extra does not exist.
Is this fixture suppose to be in the class ?

backend/backend/urls.py
IC-1101asterisk: I assume you mean .as_view
This is because I made a ViewSet instead of a simple View
I'm not sure why this was at the start of this feature I need to check but I feel like I didn't really have a choice

backend/api/tests/views/test_views_task_profiling.py
```diff
@@ -43,21 +43,20 @@ def setUp(self) -> None:
         ]
```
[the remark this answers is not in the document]
IC-1101asterisk: This is needed by a few tests that test with AuthenticatedBackendClient (those do not have a channel by default thus they need the previous workaround with extras)

backend/users/views/user.py
IC-1101asterisk: I'm not sure what you mean, unique username *are* required I'm catching this exception but really not expecting it (unless something *veryyyyyyy* wrong is going on with the DB)

guilhem-barthes: Then I wouldn't recommend not catching `or User.MultipleObjectsReturned`

backend/backend/urls.py
guilhem-barthes: I mean the `({"get": "list"}))`

guilhem-barthes (commented):
I would recommend to remove the

backend/api/tests/views/test_utils.py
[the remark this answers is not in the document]
guilhem-barthes: Sorry I didn't see before the other use of `self.extra`

charts/substra-backend/values.yaml
```diff
@@ -696,6 +694,8 @@ oidc:
     loginValidityDuration: 3600
     ## @param oidc.users.channel The channel to assign OIDC users to (mandatory)
     channel: null
+    ## @param oidc.users.requireApproval Is not compatible with default channel
```
guilhem-barthes, line 697: Make more explicit what this parameter is doing, in addition to tell with which option it is imcompatible

IC-1101asterisk: no they should incompatible, the default channel is assigned to every OIDC created user. While the requireApproval should allow the OIDC to create users with no channel

Owlfred:

IC-1101asterisk:
/e2e --tests sdk,frontend --refs substra-frontend=feat/unactivated/users

Owlfred:

charts/substra-backend/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- New `requireApproval` field, that triggers the User Awaiting Approval functionnality ([#680](https://github.com/Substra/substra-backend/pull/680))
```
guilhem-barthes, line 7:
```suggestion
- New `requireApproval` field, that triggers the User Awaiting Approval functionality ([#680](https://github.com/Substra/substra-backend/pull/680))
```

charts/substra-backend/README.md
```diff
@@ -284,6 +284,7 @@ Else, you must strike a balance: longer durations are more convenient, but risk
 | `oidc.users.useRefreshToken`            | Attempt to refresh user info in the background.                                                                                                                 | `true`  |
 | `oidc.users.loginValidityDuration`      | How long a user account is valid after an OIDC login, in seconds                                                                                                | `3600`  |
 | `oidc.users.channel`                    | The channel to assign OIDC users to (mandatory)                                                                                                                 | `nil`   |
+| `oidc.users.requireApproval`            | This switches on the user approval. A user using the OIDC login for the first time will need approval from an admin. Is not compatible with default channel     | `false` |
```
guilhem-barthes, line 287: nitpicking:
```suggestion
| `oidc.users.requireApproval`            | Activate the user approval. A user using the OIDC login for the first time will need approval from an admin. It is not compatible with default channel     | `false` |
```

IC-1101asterisk:
/e2e --tests sdk,frontend --refs substra-frontend=feat/unactivated/users

Owlfred:

guilhem-barthes (approved):
Thnaks for your work! 🙏  Just a typo and a proposed change

IC-1101asterisk:
/e2e --tests sdk --refs substra-frontend=feat/unactivated/users

Owlfred:

IC-1101asterisk:
/e2e --tests sdk --refs substra-frontend=feat/unactivated/users

Owlfred:

PR #671, opened 2023-06-13
chore!: stop writing on root file system

oleobal:
## Description

_(this PR replaces #657 because I messed up my branches, sorry)_

Writing on the root filesystem fails when `readOnlyRootFilesystem` is set, which is the case in some environments, such as OpenShift. This PR fixes this.

It also removes one permission we used to ask for, because it looks unused. 🤞 

Reverts most changes from #668 >_>

Fixes FL-1000

## Checklist

- [x] [changelog](../CHANGELOG.md) was updated with notable changes
- [x] documentation was updated

PR #693, opened 2023-07-17
chore(helm): stop requesting superfluous permissions

oleobal:
## Description

Remove permissions that we think are superfluous.

(this was split from #671)

## How has this been tested?

⬇️ (sufficient, I hope)

## Checklist


Owlfred:

SdgJlbl (approved):
Do we have a process to ensure that the helm chart version is the right one? 🤔

oleobal:
@SdgJlbl 

> Do we have a process to ensure that the helm chart version is the right one? 🤔

Sort of, the CI checks the proposed version doesn't already exist on the repo: https://github.com/Substra/substra-backend/blob/main/.github/workflows/helm.yml#L41

but that's it

charts/substra-backend/templates/rbac.yaml
```diff
@@ -20,9 +20,12 @@ rules:
     resources: ["secrets"]
     verbs: ["get", "watch", "list", "create", "update", "patch"]
```
guilhem-barthes, line 21: Do we actually need "create", "update", "patch" on secrets? It looks like this Role just use read access

charts/substra-backend/templates/rbac.yaml
```diff
@@ -20,9 +20,12 @@ rules:
...
   - apiGroups: [""]
-    resources: ["pods", "pods/log", "pods/status", "pods/exec"]
+    resources: ["pods/log", "pods/status"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["pods", "pods/exec"]
     verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
-  - apiGroups: ["batch", "extensions"]
+  - apiGroups: ["batch"]
```
guilhem-barthes, line 28: Do we actually use batch jobs in the worker?

oleobal: Ctrl+F-ing `batch` in the code base would lead my to believe we don't

guilhem-barthes: But do we need to run jobs in the worker? I thought we just ran pods :x

SdgJlbl: To the best of my knowledge, we don't use K8s Jobs currently

PR #695, opened 2023-07-24
fix: exclude "external" from UserAwaitingApproval

IC-1101asterisk:
## Description

one line PR to add an exclude as a fix to "external" users being considered as a UserAwaitingApproval

basically like "deleted" those username are the only kind of users allowed to not have a channel without beeing a UserAwaitingApproval

Fixes FL-989

### DIFF
--- a/backend/users/views/user.py
+++ b/backend/users/views/user.py
@@ -259,7 +259,7 @@ class UserAwaitingApprovalViewSet(
     filterset_class = UserFilter
 
     def get_queryset(self):
-        return self.user_model.objects.filter(channel=None).exclude(username="deleted")
+        return self.user_model.objects.filter(channel=None).exclude(username="deleted").exclude(username="external")
 
     def delete(self, request, *args, **kwargs):
         try:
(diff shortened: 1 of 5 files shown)


backend/users/views/user.py
```diff
@@ -259,7 +259,7 @@ class UserAwaitingApprovalViewSet(
     filterset_class = UserFilter
 
     def get_queryset(self):
-        return self.user_model.objects.filter(channel=None).exclude(username="deleted")
+        return self.user_model.objects.filter(channel=None).exclude(username="deleted").exclude(username="external")
```
oleobal, line 262: Perhaps we could use the [`in` field lookup](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#in).

This would allow us to write the exclusions as a list we can give a name to, which is useful for documenting its purpose and reusing it elsewhere.

```suggestion
        VIRTUAL_USERS=["deleted", "external"]
        return self.user_model.objects.filter(channel=None).exclude(username__in=VIRTUAL_USERS)
```

`VIRTUAL_USERS` is not a very good name and the list probably shouldn't be declared there but you get what I mean

oleobal (approved):

PR #698, opened 2023-07-28
feat: improve ImplicitBearerToken

oleobal:
## Description

- Solve the issue of sometimes issuing tokens that are about to expire, by just issuing an new token every time and relying on https://github.com/Substra/substra/pull/381 to clean them up

- Add a new `server.allowImplicitLogin` option, allowing node admins to disable the option altogether in order to improve security practices.

- Extend `/active-api-tokens -X DELETE` to also be able to delete `ImplicitBearerToken`, adding an `id` field to `ImplicitBearerToken` for this purpose. This is to enable users to terminate their sessions, as per security recommendations (this is leveraged by https://github.com/Substra/substra/pull/381)

Closes FL-1067, FL-1140

Companion to https://github.com/Substra/substra-documentation/pull/336

## How has this been tested?

Tried it on my machine

## Checklist


backend/backend/views.py
```diff
@@ -1,6 +1,4 @@
 from django.conf import settings
-from django.core.exceptions import ObjectDoesNotExist
```
SdgJlbl, line 2: Just curious, any particular reason to repackage the Django exception in our own?

backend/backend/views.py
```diff
@@ -30,15 +31,15 @@ class ObtainBearerToken(DRFObtainAuthToken):
     throttle_classes = [AnonRateThrottle, UserLoginThrottle]
 
     def post(self, request, *args, **kwargs):
+        if not settings.EXPIRY_TOKEN_ENABLED:
+            raise GoneException("Implicit login is disabled on this server")
```
SdgJlbl, line 35: I understand the logic behind naming this exception, but if it might not make sense to the user unaware of how it is implemented underneath. Why not raise something like `UnsupportedAuthenticationMethodError`?

backend/backend/views.py
```diff
@@ -67,15 +68,14 @@ def get(self, request, *args, **kwargs):
...
-            implicit_token = None
+        implicit_tokens = [
+            ImplicitBearerTokenSerializer(token).data
+            for token in ImplicitBearerToken.objects.filter(user=request.user).order_by("-created")
+        ]
         return ApiResponse(
             {
                 "tokens": tokens,
-                "implicit_token": implicit_token,
+                "implicit_tokens": implicit_tokens,
```
SdgJlbl, line 78: This endpoint used to return a token or None, now it's always a list (potentially empty). Is this endpoint used externally, in which case we need to adjust?

backend/backend/views.py
oleobal: I'm not sure I understand the question. Do you mean replacing Django's `ObjectDoesNotExist` with a new `GoneException`?

If so, I hadn't thought about using Django builtins (just looked at DRF's exceptions, and there were none for 410), but I'd expect throwing `ObjectDoesNotExist` to result in a 404 or 500.

backend/backend/views.py
oleobal: Good observation. I'm not aware of anything using this at the moment, the plan was to make these tokens visible in the frontend (following remarks from the audit) but it has yet to be implemented.

backend/backend/views.py
oleobal: Oh, right, that makes sense for easier debugging.

Owlfred:

Owlfred:

backend/backend/views.py
```diff
@@ -67,26 +68,23 @@ def get(self, request, *args, **kwargs):
...
         )
 
     def delete(self, request, *args, **kwargs):
-        try:
-            token = BearerToken.objects.get(id=request.GET.get("id"))
-            if request.user == token.user:
-                token.delete()
+        for model in [BearerToken, ImplicitBearerToken]:
+            tokens = model.objects.filter(id=request.GET.get("id"))
+            if len(tokens) == 1 and request.user == tokens[0].user:
```
SdgJlbl, line 85: Minor but if we expect only one object, or raise, [get](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#get) does the job

SdgJlbl (approved):
Thanks for the PR

backend/backend/views.py
oleobal: My worry is since `len(tokens) == 0` is a common outcome of the above request, I'd be controlling flow with exceptions.

SdgJlbl: better ask for forgiveness than permission ☯️

oleobal:
Waiting for https://github.com/Substra/substra/pull/381 to merge

Owlfred:

Owlfred:

PR #699, opened 2023-07-31
chore!: rename the "detail" field in DRF exceptions to "message"

oleobal:
## Description

On error (HTTP 4XX or 5XX), our API's JSON reponses are inconsistent. This is in large part because our own code returns `{"message": ...}` but DRF returns `{"detail": ...}`.

This adds a custom exception handler so DRF error reponses match our own style.

The alternative (renaming `message` to `detail` in our code) was a lot more work, but it might be a bit preferable: this PR adds code and a new layer rather than correct things, and maybe `detail` really is better. I don't think these pros outweigh having to perform and test the change, but I'm open to being wrong.

## How has this been tested?

Tried it out on my personal machine

## Checklist


Owlfred:

guilhem-barthes:
`detail` is a field defined in [RFC 9457](https://www.rfc-editor.org/rfc/rfc9457). It is also widely adopted by different frameworks. What is the added value to use `message` over `detail`?

oleobal:
@guilhem-barthes 

> `detail` is a field defined in [RFC 9457](https://www.rfc-editor.org/rfc/rfc9457). It is also widely adopted by different web frameworks. What is the added value to use `message` over `detail`?

As mentioned in the PR description, it's just easier in our codebase to standardize on `message` over `detail`. I don't know how that weights compared to industry standardization.

SdgJlbl:
I think there is value in following industry standards. Do you have any estimate of how much work that would represent?

oleobal:
See https://github.com/Substra/substra-backend/pull/705 for replacement to this PR

PR #735, opened 2023-09-20
feat(helm)!: separate worker serviceAccount.create from rbac.create

oleobal:
## Description

It's currently impossible to create a service account but not role and rolebinding for the worker, whereas it is for other backend components. This PR fixes that by adding new `worker.serviceAccount.create` and `worker.serviceAccount.name` options, like the other components.

## How has this been tested?

Hasn't been

## Checklist


Owlfred:

charts/substra-backend/README.md
```diff
@@ -124,6 +124,8 @@ See [UPGRADE.md](https://github.com/Substra/substra-backend/blob/main/charts/sub
 | `worker.tolerations`                           | Toleration labels for pod assignment                                                                                                               | `[]`                      |
 | `worker.affinity`                              | Affinity settings for pod assignment, ignored if `DataSampleStorageInServerMedia` is `true`                                                        | `{}`                      |
 | `worker.rbac.create`                           | Create a role and service account for the worker                                                                                                   | `true`                    |
```
guilhem-barthes, line 126:
```suggestion
| `worker.rbac.create`                           | Create a role for the worker                                                                                                   | `true`                    |
```

charts/substra-backend/README.md
```diff
@@ -124,6 +124,8 @@ See [UPGRADE.md](https://github.com/Substra/substra-backend/blob/main/charts/sub
...
+| `worker.serviceAccount.create`                 | Create a service account for the worker                                                                                                            | `true`                    |
+| `worker.serviceAccount.name`                   | The name of the ServiceAccount to use                                                                                                              | `""`                      |
```
guilhem-barthes, line 128: Perhaps adds info on the generated value if it's an empty string

charts/substra-backend/UPGRADE.md
```diff
@@ -1,5 +1,9 @@
 ## unreleased
 
+## 23.0.0
+
+If you have set `worker.rbac.create=false`, you now need to also set `worker.serviceAccount.create=false` to preserve the same behavior.
```
guilhem-barthes, line 5: Do we envision cases where we only create a SA without binding it?

charts/substra-backend/templates/_helpers.tpl
```diff
@@ -140,6 +140,18 @@ Return the user list
...
 
+
+{{/*
+    Create the name fo the service account to use for the worker
+*/}}
+{{- define "substra.worker.serviceAccountName" -}}
+{{- if .Values.worker.serviceAccount.create -}}
+    {{ default (printf "%s-event" ( include "substra.fullname" .)) .Values.worker.serviceAccount.name }}
+{{- else -}}
+    {{ default "default" .Values.worker.serviceAccount.name }}
```
guilhem-barthes, line 151: the default value is "default"? I don't know what are best practices in that case but it looks like it could be improved

guilhem-barthes (commented):
Thanks for that! I left some comments to improve my understand. We would re-use a similar approach for the builder SA / RBAC

charts/substra-backend/UPGRADE.md
oleobal: No, this behavior is just the simplest to implement.

What do you think would be best? If RBAC is off then SA creation is automatically disabled as well?

charts/substra-backend/templates/_helpers.tpl
oleobal: Very good point, I didn't think to check one could fail out of helm templates with the `fail` template until your comment.

Owlfred:

Owlfred:

Owlfred:

SdgJlbl (approved):

PR #764, opened 2023-10-12
[sub]chore(builder): waitPostgresqlInitContainer

thbcmlowk:
…-container waitPostgresqlInitContainer

## Checklist

### DIFF
--- a/charts/substra-backend/templates/statefulset-builder.yaml
+++ b/charts/substra-backend/templates/statefulset-builder.yaml
@@ -51,6 +51,49 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ template "substra.fullname" . }}-builder
+      initContainers:
+      {{- include "common.waitPostgresqlInitContainer" . | nindent 6 }}
+      {{- if .Values.privateCa.enabled }}
+      - name: add-cert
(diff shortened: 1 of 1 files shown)


charts/substra-backend/templates/statefulset-builder.yaml
```diff
@@ -137,48 +180,6 @@ spec:
...
-        - "--verbosity=debug"
-        volumeMounts:
-          - name: docker-cache
-            mountPath: /cache
-            readOnly: False
-          {{- if .Values.kaniko.dockerConfigSecretName }}
-          - name: docker-config
-            mountPath: /kaniko/.docker
-          {{- end }}
       {{- end}}
```
oleobal, line 183: Possibly orphaned `{{- end}}` ?

thbcmlowk: Indeed I think that `{{ -end }}` was misplaced. It should belong with this [if](https://github.com/Substra/substra-backend/blob/chore/wait-pgsql-builder/charts/substra-backend/templates/statefulset-builder.yaml#L80).
```yaml
      {{- if .Values.kaniko.cache.warmer.cachedImages }}
```

Nicely spotted, thank you.

SdgJlbl (approved):

PR #769, opened 2023-10-18
chore(ci): pass force-publish input to docker-build workflows

kaanyagci:
## Description

The input was defined for manual runs but not passed to the `docker-build` from `substra/substra-gha-workflows`, it had no effect

## Checklist

- [ ] [changelog](../CHANGELOG.md) was updated with notable changes
- [ ] documentation was updated

PR #772, opened 2023-10-19
fix(ci): fix syntax issue in docker-build

kaanyagci:
## Description

Fix[ the syntax error](https://github.com/Substra/substra-backend/actions/runs/6574521039) related to the missing default value for force-push input caused by #769 

## Checklist


oleobal (approved):
🤔 if I understand correctly this serves to catch the case where `force-publish` in null?

PR #823, opened 2024-02-13
feat!: add new compute tasks status linked to function status

guilhem-barthes:
## Companion PR

- https://github.com/Substra/orchestrator/pull/366
- https://github.com/Substra/substra-tests/pull/319
- https://github.com/Substra/substra/pull/397
- https://github.com/Substra/substra-frontend/pull/297
- https://github.com/Substra/substra-documentation/pull/390

## Description

Part of FL-1397

## Checklist


backend/api/migrations/0056_alter_computetask_status.py
```diff
@@ -0,0 +1,51 @@
...
+            field=models.CharField(
+                choices=[
+                    ("STATUS_WAITING_FOR_BUILDER_SLOT", "Status Waiting For Builder Slot"),
+                    ("STATUS_BUILDING", "Status Building"),
+                    ("STATUS_WAITING_FOR_PARENT_TASKS", "Status Waiting For Parent Tasks"),
+                    ("STATUS_WAITING_FOR_EXECUTOR_SLOT", "Status Waiting For Executor Slot"),
+                    ("STATUS_DOING", "Status Doing"),
+                    ("STATUS_DONE", "Status Done"),
+                    ("STATUS_CANCELED", "Status Canceled"),
+                    ("STATUS_FAILED", "Status Failed"),
```
ThibaultFy, line 42: Did you tried to deploy a frontend to see how it reacts ? I'm afraid that the new name are too long and might lead to a weird behaviour :/

guilhem-barthes: We are going to address that with product and design tomorrow or beginning next week and do a follow-up PR

backend/api/models/computeplan.py
```diff
@@ -97,23 +105,21 @@ def update_status(self) -> None:
             compute_plan_status = self.Status.PLAN_STATUS_FAILED
         elif self.cancelation_date or stats["canceled_count"] > 0:
             compute_plan_status = self.Status.PLAN_STATUS_CANCELED
-        elif stats["waiting_count"] == stats["task_count"]:
+        elif stats["waiting_builder_slot_count"] == stats["task_count"]:
             compute_plan_status = self.Status.PLAN_STATUS_WAITING
-        elif stats["waiting_count"] < stats["task_count"] and stats["doing_count"] == 0 and stats["done_count"] == 0:
+        elif (
+            stats["waiting_builder_slot_count"] < stats["task_count"]
```
ThibaultFy, line 111: can't we remove this condition ? we check the `==` case in the previous elif, and it can't be `>`

backend/api/tests/asset_factory.py
```diff
@@ -181,6 +181,9 @@ def get_log_permissions(owner: str, public: bool) -> dict:
 def get_computetask_dates(status: int, creation_date: datetime.datetime) -> tuple[datetime, datetime]:
     start_date = end_date = None
     if status in (
+        ComputeTask.Status.STATUS_BUILDING,
+        ComputeTask.Status.STATUS_WAITING_FOR_PARENT_TASKS,
+        ComputeTask.Status.STATUS_WAITING_FOR_EXECUTOR_SLOT,
```
ThibaultFy, line 186: Do we want to start the execution date if waiting ?

backend/api/migrations/0056_alter_computetask_status.py
```diff
@@ -0,0 +1,51 @@
...
+                ],
+                max_length=64,
+                default="STATUS_DONE",
+            ),
+            preserve_default=False,
+        ),
+        migrations.RunPython(migrate_data),
```
SdgJlbl, line 49: no reverse data migration?

backend/api/tests/asset_factory.py
ThibaultFy: Cause we didn't considered the status TODO previously, which seems to be the equivalent of `WAITING_FOR_EXECUTOR_SLOT`. And waiting for a parent task would not be considered as the execution of the task itself imo.

ThibaultFy (commented):
Thx for the PR :) 
Some questions, especially about the start date. You seems to have put some effort on this so I would be curious to understand why you decided to add a start date for the two WAINTING STATUS (parents and executor).

backend/api/migrations/0056_alter_computetask_status.py
guilhem-barthes: I didn't see any down migration, either for data or schema. Should we start adding it?

backend/api/tests/asset_factory.py
guilhem-barthes: The `TODO`from before includes `STATUS_BUILDING`, `STATUS_WAITING_FOR_PARENT_TASKS`, `STATUS_WAITING_FOR_EXECUTOR_SLOT`. The choice I made was to start the compute task when it starts building, I can understand this could be challenged, what do you think @SdgJlbl

backend/api/tests/views/test_views_computetask.py
```diff
@@ -33,6 +33,8 @@ class ComputeTaskViewTests(APITestCase):
     client_class = AuthenticatedClient
 
     def setUp(self):
+        self.MaxDiff = None
```
SdgJlbl, line 36: What does this do?

backend/api/models/computeplan.py
guilhem-barthes: Good call!

backend/api/tests/views/test_views_computetask.py
guilhem-barthes: Thanks for the heads up, that's a left over of the test debugging

backend/api/migrations/0056_alter_computetask_status.py
SdgJlbl: IMO, it depends how much effort it is 🤷 
If it's more or less free, it's always nice to have, in case we need to revert on a production project. Else, if it's not the case for other migrations anyway, don't spend time on it.

backend/api/tests/asset_factory.py
SdgJlbl: I can see an option where we only keep the execution part in this timer, and keep track of the building time separately, and that way, show both of these in the frontend task profiling. Or we can decide that what matters most is the "real" time spent (a bit like user time can be different from clock time when profiliing code). 
I might be wrong but I feel like it's something to keep in mind, but it does not matter too much right now, given that we are going to revamp the task profiling part very soon (unless the start time is also used for some other feature I missed).

backend/api/migrations/0056_alter_computetask_status.py
guilhem-barthes: I don't even find documentation on reverse data_migration and migrations [12](https://github.com/Substra/substra-backend/blob/ef127d130c2ce3544131c8637ceea9c7df7ed040/backend/api/migrations/0012_alter_algo_category.py#L46) and [14](https://github.com/Substra/substra-backend/blob/ef127d130c2ce3544131c8637ceea9c7df7ed040/backend/api/migrations/0014_alter_model_category.py#L42) don't include it

backend/api/tests/asset_factory.py
guilhem-barthes: I keep it that way and we would act when we refactor building time if everyone agree

SdgJlbl (approved):

PR #827, opened 2024-02-20
feat: update computeplan status

guilhem-barthes:
## Description

Fixes FL-1422

## How has this been tested?
CI https://github.com/Substra/substra-tests/pull/320

## Checklist


backend/api/management/commands/generate_fixtures.py
```diff
@@ -17,9 +17,8 @@ def handle(self, *args, **options):
         self.create_functions()
         self.create_data_manager()
         self.create_data_samples()
-        self.create_empty_cp()
+        self.create_created_cp()
```
ThibaultFy, line 20:
```suggestion
        self.create_cp()
```
This naming might make more sense ? WDYT ?

backend/api/management/commands/generate_fixtures.py
```diff
@@ -106,12 +105,12 @@ def create_data_samples(self):
             factory.create_datasample_files(key=test_data_sample.key)
             self.test_data_sample_keys.append(test_data_sample.key)
 
-    def create_empty_cp(self):
-        logger.debug("  Create empty CP")
+    def create_created_cp(self):
+        logger.debug("  Create created CP")
```
ThibaultFy, line 109:
```suggestion
        logger.debug("  Create CP")
```
Same suggestion

backend/api/models/computeplan.py
```diff
@@ -98,17 +96,15 @@ def update_status(self) -> None:
         """Compute cp status from tasks counts."""
         stats = self.get_task_stats()
         if stats["task_count"] == 0:
-            compute_plan_status = self.Status.PLAN_STATUS_EMPTY
+            compute_plan_status = self.Status.PLAN_STATUS_CREATED
+        elif stats["waiting_builder_slot_count"] == stats["task_count"]:
+            compute_plan_status = self.Status.PLAN_STATUS_CREATED
```
ThibaultFy, line 101: You could group the two in the same `if`.

Also, the `self.Status.PLAN_STATUS_TODO` was completely removed. Is that on purpose ? If yes, why ?

backend/api/models/computeplan.py
```diff
@@ -89,7 +87,7 @@ def get_task_stats(self) -> dict:
             waiting_executor_slot_count=Count(
                 "key", filter=Q(status=ComputeTask.Status.STATUS_WAITING_FOR_EXECUTOR_SLOT)
             ),
-            doing_count=Count("key", filter=Q(status=ComputeTask.Status.STATUS_DOING)),
+            executing_count=Count("key", filter=Q(status=ComputeTask.Status.STATUS_EXECUTING)),
```
ThibaultFy, line 90: This is unused if you don't reactivate the case on `self.Status.PLAN_STATUS_TODO`

ThibaultFy (commented):
Some comment, thanks for the PR :)

backend/api/models/computeplan.py
guilhem-barthes: It is used in frontend https://github.com/Substra/substra-frontend/pull/299/files#diff-a70efa5998c05ae5dfe367770344a8a5598e6e38c1d27a340bac6cd546cc38c6R8-R9

backend/api/models/computeplan.py
guilhem-barthes: Yeah I will merge it if you think it is cleaner!

So for the status changes:
- `EMPTY` and `WAITING` have been merged into `CREATED` which states only the existence of CP (EMPTY being just CREATED with 0 tasks)
- `WAITING` lost it's meaning by itself (waiting what? the build? the execution?)
- `TODO` was a reflection of a compute task being ready to be ran (built + execution). FOr compute task statuses, this status became `WAITING_FOR_EXECUTOR_SLOT`. and is not the first action done on the task. With product, we didn't think that the compute plan status needed the granularity of compute tasks (i.e. needed 3 types of waiting) and a compute plan could also have tasks in different waiting states at the same time (therefore what todo would mean? One task has to be built? executed? etc)

ThibaultFy: ok, so you mean that the case `stats["doing_count"] == 0 and stats["done_count"] == 0:` is in fact that a task is `WAITING_FOR_EXECUTOR_SLOT`, so the CP should still be in `DOING` ?

guilhem-barthes: Yes, because `BUILDING` is part of doing. That's why we renamed `DOING` in `EXECUTING`, because `DOING` encompasses `BUILDING` and `EXECUTING`. I don't think we could reach a state where we have neither 

You are raising an interesting point, I think I should revert renaming `doing_count`  to `executing_count` and adding `BUILDING` as part of it

guilhem-barthes: After second thoughts, it seems like we should keep `executing_count` as these metrics refers to task statuses (explicitely providing a count for each status) rather than a compute task one

ThibaultFy (approved):
Thanks for all the work :p

PR #832, opened 2024-02-23
chore(examples/values): default to unique builder

thbcmlowk:
## Description

Enabling the unique builder feature for `substra-ci`

## Checklist

### DIFF
--- a/examples/values/backend-org-1.yaml
+++ b/examples/values/backend-org-1.yaml
@@ -128,3 +128,6 @@ orchestrator:
     - yourchannel:
         restricted: false
         model_export_enabled: true
+
+builder:
+  enabled: true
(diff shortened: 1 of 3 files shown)


Owlfred:

examples/values/backend-org-1.yaml
```diff
@@ -125,6 +125,13 @@ orchestrator:
     - mychannel:
         restricted: false
         model_export_enabled: true
+        chaincode:
+          name: mycc
```
guilhem-barthes, line 129: We don't have a distributed mode, why to add these values?

thbcmlowk: Side-effect of the rebase (distributed mode was still a thing at the time I created those files). Removing it, thank you.

PR #833, opened 2024-02-23
fix: load remote image once

guilhem-barthes:
## Description

Fixes FL-1440

## Checklist


Owlfred:

SdgJlbl (approved):
I guess this do the job for now, but I'm wondering if the clean way of doing it would not be not to download the image if it already exists locally (whether the pod is going to be reused or not).
Definitely for a follow-up, though

PR #837, opened 2024-02-29
chore: add missing repo

guilhem-barthes:
## Checklist

### DIFF
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -22,4 +22,4 @@ jobs:
       uses: substra/substra-gha-workflows/.github/workflows/helm.yml@main
       secrets: inherit
       with:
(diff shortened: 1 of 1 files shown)


ThibaultFy (approved):
thanks 🙏

thbcmlowk (commented):
What do you think about triggering the workflow at least on PR based on changes made on the `.github` workflow?
```diff
   pull_request:
     branches:
       - "main"
     paths:
       - "charts/**"
+      - ".github/workflows/helm.yml"
```

guilhem-barthes:
> ```diff
> - ".github/workflows/helm.yml"
> ```

Sounds good ! But the problem are having is detected when we package, which only called during "publish". Should we add a package in each PR to try that no dependency is missed ?

thbcmlowk:
> > ```diff
> > - ".github/workflows/helm.yml"
> > ```
> 
> Sounds good ! But the problem are having is detected when we package, which only called during "publish". Should we add a package in each PR to try that no dependency is missed ?

Yes, already did! https://github.com/Substra/substra-gha-workflows/pull/10

thbcmlowk (approved):
Re-triggered the CI after merge on GHA but LGTM! Thanks

PR #867, opened 2024-03-29
chore: add limits and requests to pods

SdgJlbl:
Fixes FL-1245

### DIFF
--- a/backend/builder/image_builder/image_builder.py
+++ b/backend/builder/image_builder/image_builder.py
@@ -23,6 +23,7 @@ from substrapp.compute_tasks.compute_pod import Label
 from substrapp.compute_tasks.volumes import get_worker_subtuple_pvc_name
 from substrapp.docker_registry import USER_IMAGE_REPOSITORY
 from substrapp.kubernetes_utils import delete_pod
+from substrapp.kubernetes_utils import get_resources_requirements
 from substrapp.kubernetes_utils import get_security_context
 from substrapp.lock_local import lock_resource
 from substrapp.utils import timeit
@@ -306,6 +307,7 @@ def _build_container(dockerfile_mount_path: str, image_tag: str) -> kubernetes.c
         args=args,
         volume_mounts=volume_mounts,
         security_context=container_security_context,
+        resources=get_resources_requirements(cpu_request="1000m", memory_request="4Gi", memory_limit="32Gi"),
     )
 
 
--- a/backend/substrapp/compute_tasks/compute_pod.py
+++ b/backend/substrapp/compute_tasks/compute_pod.py
@@ -6,6 +6,7 @@ from django.conf import settings
 
 from substrapp.kubernetes_utils import delete_pod
 from substrapp.kubernetes_utils import get_pod_security_context
+from substrapp.kubernetes_utils import get_resources_requirements
 from substrapp.kubernetes_utils import get_security_context
 
 NAMESPACE = settings.NAMESPACE
@@ -112,6 +113,7 @@ def create_pod(
         args=None,
         volume_mounts=volume_mounts + gpu_volume_mounts,
         security_context=get_security_context(),
+        resources=get_resources_requirements(cpu_request="1000m", memory_request="1Gi", memory_limit="64Gi"),
         env=[kubernetes.client.V1EnvVar(name=env_name, value=env_value) for env_name, env_value in environment.items()],
         **container_optional_kwargs,
     )
--- a/backend/substrapp/kubernetes_utils.py
+++ b/backend/substrapp/kubernetes_utils.py
@@ -47,6 +47,14 @@ def get_security_context(root: bool = False, capabilities: list[str] = None) -> 
     return security_context
 
 
+def get_resources_requirements(
+    *, cpu_request: str = "1000m", memory_request: str = "200M", memory_limit: str = "2G"
+) -> kubernetes.client.V1ResourceRequirements:
+    return kubernetes.client.V1ResourceRequirements(
+        requests={"cpu": cpu_request, "memory": memory_request}, limits={"memory": memory_limit}
+    )
+
+
 def pod_exists_by_label_selector(k8s_client: kubernetes.client.CoreV1Api, label_selector: str) -> bool:
     """Return True if the pod exists, else False.
 
--- a/charts/substra-backend/templates/deployment-registry-prepopulate.yaml
+++ b/charts/substra-backend/templates/deployment-registry-prepopulate.yaml
@@ -23,9 +23,21 @@ spec:
       initContainers:
       - name: wait-registry
         image: jwilder/dockerize:0.6.1
+        resources:
+          requests:
+            memory: "200Mi"
+            cpu: "500m"
+          limits:
+            memory: "400Mi"
         command: ['dockerize', '-wait', 'tcp://{{ $.Release.Name }}-docker-registry:5000']
       - name: kaniko
         image: {{ include "common.images.name" $.Values.kaniko.image }}
+        resources:
+          requests:
+            memory: "2Gi"
+            cpu: "1000m"
+          limits:
+            memory: "8Gi"
         args:
           - "--context=/docker-context"
           {{- if .dstImage }}
@@ -48,6 +60,12 @@ spec:
       containers:
       - image: gcr.io/google-containers/pause:3.2
         name: pause
+        resources:
+          requests:
+            memory: "64Mi"
+            cpu: "50m"
+          limits:
+            memory: "128Mi"
       volumes:
       - name: kaniko-dir
         emptyDir: {}
--- a/charts/substra-backend/values.yaml
+++ b/charts/substra-backend/values.yaml
@@ -172,17 +172,19 @@ server:
     ##
     ingressClassName:
 
-  ## @param server.resources Server container resources requests and limits
-  ## e.g:
-  ## resources:
-  ##   limits:
-  ##     cpu: 100m
-  ##     memory: 128Mi
-  ##   requests:
-  ##     cpu: 100m
-  ##     memory: 128Mi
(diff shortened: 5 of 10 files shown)


ThibaultFy (approved):
No idea on the values but let test it on dev then 🤞

charts/substra-backend/values.yaml
```diff
@@ -171,17 +171,19 @@ server:
...
+  ## @param server.resources.limits.memory Server container memory limit
   ##
-  resources: {}
+  resources:
+    requests:
+      cpu: "1000m"
+      memory: "4Gi"
+    limits:
+      cpu: "1000m"
+      memory: "4Gi"
```
guilhem-barthes, line 185: I would set a higher limit and monitor

charts/substra-backend/values.yaml
```diff
@@ -835,6 +865,13 @@ postgresql:
       capabilities:
         drop:
         - ALL
+    resources:
+      requests:
+        cpu: "1000m"
+        memory: "2Gi"
+      limits:
+        cpu: "1000m"
+        memory: "4Gi"
```
thbcmlowk, line 874: Seems pretty high but not too many CP ran in parallel.

image

charts/substra-backend/values.yaml
```diff
@@ -849,6 +886,13 @@ redis:
     service:
       ports:
         redis: 6379
+    resources:
+      requests:
+        cpu: "500m"
+        memory: "512Mi"
+      limits:
+        cpu: "500m"
+        memory: "1024Mi"
```
thbcmlowk, line 895: Same as pg

image

charts/substra-backend/values.yaml
```diff
@@ -888,6 +939,13 @@ minio:
     capabilities:
       drop:
         - ALL
+  resources:
+    requests:
+      cpu: "500m"
+      memory: "16Gi"
+    limits:
+      cpu: "1000m"
+      memory: "64Gi"
```
thbcmlowk, line 948: Obviously same as registry

image

thbcmlowk (commented):
Observations based on the 6 CP [workflow](https://github.com/owkin/substra-ci/actions/runs/8719148148):

image

Will try to launch CP in extreme mode.

backend/builder/image_builder/image_builder.py
```diff
@@ -291,6 +292,7 @@ def _build_container(dockerfile_mount_path: str, image_tag: str) -> kubernetes.c
         args=args,
         volume_mounts=volume_mounts,
         security_context=container_security_context,
+        resources=get_resources_requirements(cpu_request="1000m", memory_request="4Gi", memory_limit="16Gi"),
```
thbcmlowk, line 295: I'm afraid memory limit is too low:

image

* First three small peaks: `torch` with cpu index
* Three last peaks: `torch` with GPU support

backend/substrapp/kubernetes_utils.py
```diff
@@ -47,6 +47,14 @@ def get_security_context(root: bool = False, capabilities: list[str] = None) ->
     return security_context
 
 
+def get_resources_requirements(
+    *, cpu_request: str = "1000m", memory_request: str = "100M", memory_limit: str = "1G"
```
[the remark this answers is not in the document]
thbcmlowk: And no CPU limit?

backend/builder/image_builder/image_builder.py
thbcmlowk, line 295: I'm afraid memory limit is too low:

image

* First three small peaks: `torch` with cpu index
* Three last peaks: `torch` with GPU support 

---

```suggestion
        resources=get_resources_requirements(cpu_request="1000m", memory_request="4Gi", memory_limit="32Gi"),
```

backend/substrapp/compute_tasks/compute_pod.py
```diff
@@ -112,6 +113,7 @@ def create_pod(
         args=None,
         volume_mounts=volume_mounts + gpu_volume_mounts,
         security_context=get_security_context(),
+        resources=get_resources_requirements(cpu_request="1000m", memory_request="1Gi", memory_limit="256Gi"),
```
thbcmlowk, line 116: Memory limit seem pretty high to me (not even sure we have this memory available in usual production setups). Based on the [documentation](https://docs.substra.org/en/stable/how-to/deploying-substra/index.html#hardware-requirements), hardware requirements recommend at least 30GB so I suggest:
```suggestion
        resources=get_resources_requirements(cpu_request="1000m", memory_request="1Gi", memory_limit="32Gi"),
```
---

FYI - even though I am not sure the benchmark performed in this case is exhaustive enough: 
image

backend/substrapp/kubernetes_utils.py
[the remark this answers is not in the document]
SdgJlbl: We don't want to set any CPU limits on CPU-heavy tasks, to avoid CPU throttling
In addition, GPU can't be shared across tasks, so compute functions will be scheduled on separate VM, and it makes no sense to limit CPU in that setting.

charts/substra-backend/values.yaml
```diff
@@ -523,17 +552,18 @@ builder:
...
+  ## @param builder.resources.limits.memory Builder container memory limit
   ##
-  resources: {}
+  resources:
+    requests:
+      cpu: "2000m"
+      memory: "4Gi"
+    limits:
+      cpu: "2000m"
+      memory: "8Gi"
```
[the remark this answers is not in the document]
SdgJlbl: I unfortunately haven't taken any screenshot, but builder with only 1 CPU was throttling, that's why I doubled the value to 2000 millli.
As for the memory, I also have some notes relating usage above 1 Gi, could this be dependent of the image size? 
What do you think about keeping it as is for now, and monitor it in the coming days?

charts/substra-backend/values.yaml
SdgJlbl: Also, I think from past discussions with DBAs that PostGre is choosing different strategies to execute the queries depending on the amount of RAM available, so it's probably a bad idea for performance to restrict the RAM too much.
In fact, I was wondering if I should put an higher value, especially since the amount of data in the DB will only increase when the project is run in production.

charts/substra-backend/values.yaml
```diff
@@ -864,10 +908,17 @@ docker-registry:
...
     deleteEnabled: true
   service:
     type: NodePort
+  resources:
+    requests:
+      cpu: "500m"
+      memory: "16Gi"
+    limits:
+      cpu: "500m"
+      memory: "64Gi"
```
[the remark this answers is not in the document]
SdgJlbl: do you think we should increase the memory limit, or on the contrary put something lower to force the cache to flush?

backend/substrapp/compute_tasks/compute_pod.py
SdgJlbl: I didn't want any task to be OOM killed if the memory was indeed available, that's why I set up a very high limit.
But I can lower it if you think it's an acceptable risk.

charts/substra-backend/values.yaml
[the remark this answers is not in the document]
thbcmlowk: I assume I've tried with consequent docker images. From what I understand, the memory part is mostly taken by kaniko pods, but ok to keep an eye on it.

charts/substra-backend/values.yaml
thbcmlowk: Ok to keep it as is for now, and monitor it in the coming days

charts/substra-backend/values.yaml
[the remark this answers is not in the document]
thbcmlowk: That's a good one. I just checked, and cached memory hasn't dropped:

image

Maybe we can we create a follow-up ticket for this?

backend/substrapp/compute_tasks/compute_pod.py
thbcmlowk: I mean I think 256 is huge, but we can see how it goes from there. If you don't mind me asking, why having restricted the builder part that much then in comparison?

backend/substrapp/kubernetes_utils.py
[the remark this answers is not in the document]
thbcmlowk: I meant, we are setting pretty low defaults in comparison to args passed where that function are called, so I would assume we would provide a default for CPU limit to prevent any misuse of the function (that can be eventually cancelled for CPU-heavy tasks). But as this scenario is a bit unlikely to happen, no problem to keep it as is 👍

charts/substra-backend/templates/deployment-registry-prepopulate.yaml
```diff
@@ -23,9 +23,21 @@ spec:
...
+            memory: "400Mi"
         command: ['dockerize', '-wait', 'tcp://{{ $.Release.Name }}-docker-registry:5000']
       - name: kaniko
         image: {{ include "common.images.name" $.Values.kaniko.image }}
+        resources:
+          requests:
+            memory: "2Gi"
+            cpu: "1000m"
+          limits:
+            memory: "4Gi"
```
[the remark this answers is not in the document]
thbcmlowk: Sorry, what do you think of sticking to what's set for the builder?

backend/substrapp/compute_tasks/compute_pod.py
SdgJlbl: Kaniko pods are unlikely to be scheduled on their own VM because they don't use GPU, so in case of something going wrong with memory usage, it would directly impact all other pods.

charts/substra-backend/templates/deployment-registry-prepopulate.yaml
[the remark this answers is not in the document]
SdgJlbl: I've set both to a limit of 8Gi

PR #886, opened 2024-04-11
feat: add profiling step functions

guilhem-barthes:
## Description

Allow transferring function profiling steps through the orchestrator, allowing the builder to register event without having to have http(s) requests to the backend api server (involving having access to the db to get credentials). This also allows having duration for the function available on every backend.

Still open on naming for the two new steps for the function. For now we have:
- build_image
- save_function

This does not include the API part for the frontend, which woudl be part of another task (FL-1518)

Fixes FL-1519

# Companion PR

- https://github.com/Substra/orchestrator/pull/407

## Checklist


backend/builder/tests/test_task_build_image.py
```diff
@@ -34,7 +34,7 @@ def test_order_building_success(celery_app, celery_worker, mocker, execution_num
     function_2 = orc_mock.FunctionFactory()
 
     # BuildTask `before_start` uses this client to change the status, which would lead to `OrcError`
-    mocker.patch("builder.tasks.task.get_orchestrator_client")
+    mocker.patch("orchestrator.get_orchestrator_client")
```
SdgJlbl, line 37: unsure about that change (and the similar one below), I think you usually want to patch as close to the tested function as possible, to avoid nasty side-effects.

backend/orchestrator/resources.py
```diff
@@ -151,6 +150,16 @@ def from_grpc(cls, s: function_pb2.FunctionStatus.ValueType) -> FunctionStatus:
         return cls(function_pb2.FunctionStatus.Name(s))
 
 
+class FunctionProfilingStep(str, enum.Enum):
+    BUILD_IMAGE = "build_image"
+    SAVE_FUNCTION = "save_function"
+
+    # Not relying on `django.db.models.TextChoices` to avoid multiplicating dependencies with django
```
SdgJlbl, line 157: 👍

backend/substrapp/tests/compute_tasks/test_outputs.py
```diff
@@ -84,7 +84,7 @@ def test_save_model(settings, mocker: MockerFixture, orc_raise: bool):
         client.register_models.return_value = []
 
     add_model_from_path = mocker.patch("substrapp.compute_tasks.outputs.add_model_from_path")
-    mocker.patch("substrapp.compute_tasks.outputs.get_orchestrator_client", return_value=client)
+    mocker.patch("orchestrator.get_orchestrator_client", return_value=client)
```
SdgJlbl, line 87: Same comment as above

backend/substrapp/tests/tasks/test_compute_task.py
```diff
@@ -22,7 +22,7 @@
 
 
 def test_compute_task_exception(mocker: MockerFixture):
-    mocker.patch("substrapp.tasks.tasks_compute_task.get_orchestrator_client")
+    mocker.patch("orchestrator.get_orchestrator_client")
```
SdgJlbl, line 25: same 🙃 (and below in the file)

SdgJlbl (commented):
Thanks 👍 
Unsure about how the mocks are (re)defined in this PR, I feel like it's contrary to best practices. 
Also need a rebase ^^

SdgJlbl (approved):
Thanks for your work 🙏

PR #888, opened 2024-04-15
chore: parametrize user image docker repository

ThibaultFy:
## Description

closes FL-1541


Owlfred:

backend/substrapp/compute_tasks/image_builder.py
```diff
@@ -16,6 +16,7 @@
 IMAGE_BUILD_CHECK_DELAY = settings.IMAGE_BUILD_CHECK_DELAY
 REGISTRY = settings.REGISTRY
 SUBTUPLE_TMP_DIR = settings.SUBTUPLE_TMP_DIR
+USER_IMAGE_REPOSITORY = settings.USER_IMAGE_REPOSITORY
```
SdgJlbl, line 19: Unrelated comment but we really need to rename that module at some point 😅

examples/values/backend-org-1.yaml
```diff
@@ -42,7 +42,7 @@ containerRegistry:
     - image: substra/substra-tools:latest
       sourceRegistry: ghcr.io
       dstImage: substra/substra-tools:latest
-
+  userImageRepository: substra-user-images
```
SdgJlbl, line 45: I think we want to add the parameter explicitly in all orgs, for clarity.

SdgJlbl (approved):
Minor comment but otherwise LGTM 👍 
Thanks!

examples/values/backend-org-1.yaml
ThibaultFy: I also changed the default to `substra/user-image` to remove the `substrafoundation` mention out of date. But I let the org1 to `substra-user-images` to continue to test that the feature works

PR #917, opened 2024-05-31
feat: retry on timeout connections to external registry

SdgJlbl:
fixes fl-1569


backend/builder/image_builder/image_builder.py
```diff
@@ -160,7 +166,11 @@ def _build_container_image(path: str, tag: str) -> None:
...
         # In case of concurrent builds, it may fail. Check if the image exists.
-        if docker.container_image_exists(tag):
+        try:
+            image_exists = docker.container_image_exists(tag)
+        except Exception:
+            image_exists = False
+        if image_exists:
             logger.warning(
                 f"Build of container image {tag} failed, probably because it was done by a concurrent build",
                 exc_info=True,
```
ThibaultFy, line 176:
```suggestion
        try:
             if docker.container_image_exists(tag):
                 logger.warning(
                     f"Build of container image {tag} failed, probably because it was done by a concurrent build",
                     exc_info=True,
                     )
                     return
        except Exception:
            pass
```

ThibaultFy: I find it more explicit that way. Feel free to ignore

ThibaultFy (commented):
small comment, thanks for the PR :)

backend/builder/image_builder/image_builder.py
SdgJlbl: I like to wrap the `try except` on the minimal number of lines, to mark explicitly what could potentially raise, and to avoid to mask unexpected Exceptions (even more when using a blanket `except Exception`).

I'm curious though, what do you find more explicit in your suggestion?

ThibaultFy: I see your point. What I found more explicit is to not set `image_exists = False` cause it's never used afterwards, and to explicitely set the `except Exception: pass` to clearly indicate that we bypass all exceptions.

Owlfred:

Owlfred:

Owlfred:

ThibaultFy (approved):
thanks :)

PR #924, opened 2024-06-03
fix: bypass asset failure creation when internal error

guilhem-barthes:
## Description

Bypass `AssetFailureReport` creation whenerror type is `ERROR_TYPE_INTERNAL`. It fixes errors when the backend receives an internal error from the orchestrator (without log address/checksum, as they are not provided in this scenario)

Fixes FL-1583

## Checklist


Owlfred:

PR #933, opened 2024-06-20
fix: netpol allow worker internet

guilhem-barthes:
## Description

Fixes FL-1611

## Checklist


Owlfred:

charts/substra-backend/templates/statefulset-worker.yaml
```diff
@@ -35,6 +35,8 @@ spec:
         app.kubernetes.io/component: substra-worker
         role-api-k8s-client: 'true'
         role-database-client: 'true'
+        # Needed for downloading models from other backends + connecting to remote registry
+        role-internet-egress: 'true'
```
thbcmlowk, line 39: Do we have leads on why this was working both with skaffold and on substra-dev but not on staging?

guilhem-barthes: Was the post-mortem not clear enough?

SdgJlbl (approved):
🕸️

charts/substra-backend/templates/statefulset-worker.yaml
thbcmlowk: Very clear, thank you. Missed the thought process of it somehow when I first read it.

PR #969, opened 2024-08-05
chore(backend): force token expire

thbcmlowk:
## Description

Part of FL-1656

Disable never expiring `BearerToken`

## Companion

https://github.com/Substra/substra-frontend/pull/385

## How has this been tested?

On dev by: 
* Creating expiring tokens and one never-expiring
* Applying the migration code
* Check db results

### Creating tokens

image

```psql
 select * from users_bearertoken;
```
```psql
            created            |     note      |         expires_at         |                  id                  | user_id 
-------------------------------+---------------+----------------------------+--------------------------------------+---------
 2024-08-27 12:47:09.093701+00 | expires-60    | 2024-10-26 12:47:08.395+00 | 3a07a145-9ecd-46e8-aa00-35a9a3d2ebf2 |       4
 2024-08-27 12:47:16.34687+00  | never-expires |                            | 456efb44-4f30-49f8-bc12-46a673da9f90 |       4
```

### Applying migration

```bash
I have no name!@substra-backend-substra-backend-server-5674697f78-ds9nk:/usr/src/app$ ./manage.py migrate users
Operations to perform:
  Apply all migrations: users
Running migrations:
  Applying users.0008_alter_bearertoken_expires_at... OK
```

image

```psql
            created            |     note      |          expires_at           |                  id                  | user_id 
-------------------------------+---------------+-------------------------------+--------------------------------------+---------
 2024-08-27 12:47:09.093701+00 | expires-60    | 2024-10-26 12:47:08.395+00    | 3a07a145-9ecd-46e8-aa00-35a9a3d2ebf2 |       4
 2024-08-27 12:47:16.34687+00  | never-expires | 2024-08-27 12:50:24.131665+00 | 456efb44-4f30-49f8-bc12-46a673da9f90 |       4
```

## Checklist


thbcmlowk:
/e2e --tests=substrafl,mnist,camelyon,frontend --refs=substra-frontend=chore/disable-never-expiring-token

backend/users/migrations/0008_alter_bearertoken_expires_at.py
```diff
@@ -0,0 +1,19 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("users", "0007_implicitbearertoken_id_and_more"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="bearertoken",
+            name="expires_at",
+            field=models.DateTimeField(default=django.utils.timezone.now),
```
SdgJlbl, line 17: By default, all existing tokens will expire immediately at the time of the migration. Is that what you want?

SdgJlbl (approved):
One question, otherwise LGTM. 
I don't know enough about the OIDC implementation to know if it will work without issues there.
And we have to keep in mind at the next upgrade that all tokens will be revoked.

backend/users/models/token.py
```diff
@@ -10,7 +10,7 @@
 
 class BearerToken(Token):
     note = models.TextField(null=True)
-    expires_at = models.DateTimeField(null=True)
+    expires_at = models.DateTimeField(null=False, default=timezone.now)
```
ThibaultFy, line 13: I don't understand, does that mean that by default it expires instantly ?

backend/users/migrations/0008_alter_bearertoken_expires_at.py
thbcmlowk: I think the idea was rather to set an immediate expiration date for token that does not have one

backend/users/models/token.py
thbcmlowk: Thanks for asking, I am not sure I remember the rational behind this particular change.

guilhem-barthes: It may have been that all already existing tokens are revoked

guilhem-barthes: Do we really want a default value or do we want to force the user to provide it?

thbcmlowk: What it means is that if no value is provided, the token will be created with instant expire. Note that the fronted will disable the possibility to create token with a null `expires_at` field. One other option is to remove the default ; meaning that creating a token with a null `expires_at` field will raise the following error:
```python
django.db.utils.IntegrityError: null value in column "expires_at" of relation "users_bearertoken" violates not-null constraint
```

Not sure of what would be the best behavior, but happy to rework this if we think it's worth.

backend/users/migrations/0008_alter_bearertoken_expires_at.py
thbcmlowk: Updated, thank you

Owlfred:

backend/users/models/token.py
thbcmlowk: > Do we really want a default value or do we want to force the user to provide it

Did not refresh, sorry. Obviously better if we force the user to provide it indeed. Is there a clean way to do so?

guilhem-barthes: How did you get this `IntegrityError` ? The serializer should have returned an error when receiving an empty field for `expires_at`

backend/users/serializers/token.py
```diff
@@ -7,7 +7,7 @@
 
 
 class BearerTokenSerializer(serializers.ModelSerializer):
-    expires_at = serializers.DateTimeField(default_timezone=timezone.utc, allow_null=True)
+    expires_at = serializers.DateTimeField(default_timezone=timezone.utc, allow_null=False)
```
guilhem-barthes, line 10: If a field as a default, you have to set `required=True` to avoid receiving null values

backend/users/models/token.py
thbcmlowk: I have the error when running the [tests](https://github.com/Substra/substra-backend/blob/main/backend/api/tests/views/test_views_token.py#L15) and setting either:
* `required=True` in the `BearerTokenSerializer`
* `blank=False` in the `BearerToken` model

backend/users/serializers/token.py
thbcmlowk: `required=True` with the `default=timezone.now` at the model level does not prevent from calling with no value

backend/api/tests/views/test_views_token.py
```diff
@@ -111,3 +121,17 @@ def test_token_creation_post(authenticated_client):
...
+
+@pytest.mark.django_db
+def test_cannot_post_token_wo_expires_at(authenticated_client):
+    authenticated_client.create_user()
+    payload = {}
+    url = "/api-token/"
+    response = authenticated_client.post(url, payload)
+
+    assert response.json() == {"expires_at": ["This field is required."]}
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
guilhem-barthes, line 134: 👍

guilhem-barthes (approved):
Thanks for your work!

Owlfred:

PR #970, opened 2024-08-05
fix: store asset failure report supports exception without logs

guilhem-barthes:
## Description

Fixes Fl-1689

## Checklist


Owlfred:

charts/substra-backend/changes/970.fixed
```diff
@@ -0,0 +1 @@
+Celery was failing silently when a task called `FailableTask.on_failure` if the task didn't have a `logs` attribute (now return internal error)
```
SdgJlbl, line 1: I don't think that the right `changes` folder :)

backend/builder/exceptions.py
```diff
@@ -14,7 +13,7 @@ class PodTimeoutError(Exception):
     pass
 
 
-class BuildRetryError(_ComputeTaskError, CeleryRetryError):
+class _BuildError(RuntimeError):
```
SdgJlbl, line 16: So if I understand correctly, `BuildError` will never share its logs? 🤔

SdgJlbl (approved):
🙏

PR #975, opened 2024-08-19
fix: skaffold profile three orgs

guilhem-barthes:
## Description

Fixes FL-1705

## Checklist

### DIFF
--- a/changes/975.fixed
+++ b/changes/975.fixed
@@ -1,0 +1,1 @@
+Profile `three-orgs` activates profile in `example/secrets/skaffold.yaml` and profile paths have been updated


ThibaultFy:
/e2e --tests docs

To test, these are the e2e tests that are using this profile

Owlfred:

guilhem-barthes:
Ok, so we have the `three-orgs` profile that does not support the skaffold last version, while the other profiles are not compatible with the previous version. Which skaffold version should be supported?

ThibaultFy:
> Ok, so we have the `three-orgs` profile that does not support last version from skaffold, while the other profiles are not compatible with the previous version. Which skaffold version should be supported?

I guess the latest possible ? Is it possible to upgrade the three orgs profile to match the latest skaffold ?

PR #999, opened 2024-09-12
fix: add account operator does not synchronise content

guilhem-barthes:
## Description

Fixes FL-1697
Fixes FL-1581

## Checklist


charts/substra-backend/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: substra-backend
 home: https://github.com/Substra
-version: 26.9.1
+version: 26.11.0-alpha.1
```
SdgJlbl, line 4: probably want to update that chart number :)

guilhem-barthes:
> Thanks for the refactoring 🙏
> 
> I trust this has been tested on dev?

Yes it has been (each alpha ^^")

PR #1000, opened 2024-09-12
chore: change updateStrategy of the server and the registry

SdgJlbl:
Fixes FL-1733

Server and registry pods should no longer hang in pending because of race condition on RWO PVC. 

Untested yet.


charts/substra-backend/templates/deployment-server.yaml
```diff
@@ -12,6 +12,8 @@ spec:
     matchLabels:
         app.kubernetes.io/name: {{ template "substra.name" . }}-server
         {{ include "substra.selectorLabels" . | nindent 8}}
+  strategy:
+    type: Recreate
```
thbcmlowk, line 16: Wdyt of adding the field in the values file (just like the registry):
```suggestion
  strategy:
    type: {{ .Values.server.updateStrategy }}
```

thbcmlowk (changes requested):
Thanks!

charts/substra-backend/templates/deployment-server.yaml
SdgJlbl: I can make it configurable through the values if you think it's cleaner, I don't see any case where we wouldn't want to use the `Recreate` strategy since it's blocking the deployment because of the shared RWO PVC.
For the registry, we use an external chart, so we have to configure it through values.
No strong opinion in either direction though.

guilhem-barthes:
The default value seems to be Recreate, what this PR is changing?

thbcmlowk (approved):
Thank you for updating 🙏

PR #1011, opened 2024-10-11
fix: save image timeout 2

guilhem-barthes:
## Description

Part of FL-1725

## Checklist


charts/substra-backend/templates/statefulset-builder.yaml
```diff
@@ -200,6 +200,8 @@ spec:
             value: {{ toYaml .Values.builder.kanikoStartup.maxAttempts | quote }}
           - name: BUILDER_KANIKO_STARTUP_PENDING_STATE_WAIT_SECONDS
             value: {{ toYaml .Values.builder.kanikoStartup.checkDelay | quote }}
+          - name: IMAGE_SAVING_TIMEOUT_SECONDS
+            value: {{ .Values.builder.saveImageTimeoutSeconds | quote }}
```
thbcmlowk, line 204: Is it needed for the `builder` itself?

examples/values/backend-org-1.yaml
```diff
@@ -131,3 +131,6 @@ orchestrator:
     - yourchannel:
         restricted: false
         model_export_enabled: true
+
+builder:
+  saveImageTimeoutSeconds: 180
```
thbcmlowk, line 136: I'm not sure that was a problem in the CI, was it?

thbcmlowk (commented):
Should we also update the main default value?

examples/values/backend-org-1.yaml
guilhem-barthes: Good catch, thanks

charts/substra-backend/templates/statefulset-builder.yaml
guilhem-barthes: In my testing, in some cases the  worker was not sufficient, will retest

thbcmlowk: Also confirmed on my end, LGTM

thbcmlowk (approved):
LGTM, thanks for fixing

